Issue #11, opened 2025-01-30
Support streaming


tomkis:
For conversational-like UIs we need to support streaming of the result.

Please drive the effort to analyse the impact on

1) MCP Protocol
2) Our MCP server proxy (cc @jezekra1)

PR #23, opened 2025-01-30
feat(notifications): add broadcast and token-based progress streaming

jezekra1:
Ref: #11

### DIFF
--- a/apps/api/src/beeai_api/services/mcp_proxy.py
+++ b/apps/api/src/beeai_api/services/mcp_proxy.py
@@ -1,10 +1,19 @@
 import asyncio
 import logging
+import uuid
+import warnings
+from asyncio import CancelledError
+from collections import defaultdict
 from contextlib import AsyncExitStack, suppress, asynccontextmanager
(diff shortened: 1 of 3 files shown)


apps/api/src/beeai_api/services/mcp_proxy.py
```diff
@@ -75,6 +99,104 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+    async def register(self, provider: LoadedProvider):
+        self._notification_pipe.start_soon(self._subscribe_for_messages, provider)
+        logger.info(f"Started listening for notifications from: {provider.provider}")
+
+    async def remove(self, provider: LoadedProvider):
+        self._provider_cleanups[provider.provider]()
+        logger.info(f"Stopped listening for notifications from: {provider.provider}")
+
+    @asynccontextmanager
+    async def forward_notifications(
```
jezekra1, line 132: #TODO research
When we can do this for notifications, maybe we can do this for all messages and requests?

PR #36, opened 2025-02-03
chore: UI scaffolding

tomkis:
Couple things to consider:
 - Had to enumerate all python projects in py workspaces as it was clashing with ts projects
 - Used catalogs for libs, not sure if that makes sense

Follow up stuff:
 - Create proper MCP client (capabilities, version, name etc.)
 - Create package for dev stuff (eslint config, typescript config etc.)
 - Figure out configuration of the UI
 - Figure out production build & embedding to the system
 - Create example of how to fetch agents


apps/beeai-ui/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
```
jezekra1: I think the global `.gitignore` covers this, let's have just one if possible

tomkis: Removed gitingnore from UI completely, most of the common stuff is indeed in the root.

apps/beeai-ui/src/mcp-client.tsx
```diff
@@ -0,0 +1,60 @@
...
+      {
+        name: "example-client",
+        version: "1.0.0",
+      },
+      {
+        capabilities: {
+          prompts: {},
+          resources: {},
+          tools: {},
+        },
```
pilartomas, line 31: None of these are client capabilities.

apps/beeai-ui/src/mcp-client.tsx
```diff
@@ -0,0 +1,60 @@
...
+    );
+
+    client.connect(transport).then(() => {
+      if (closed) {
+        console.log("Transport has been closed");
+        return;
+      }
+
+      setConnectedClient(client);
+    });
```
pilartomas, line 42: It is necessary to check the server has the capabilities we require after connecting.

PR #48, opened 2025-02-05
feat(ui): add basic setup for beeai-ui app

PetrBulanek:


apps/beeai-ui/src/hooks/useCreateMCPClient.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const connectClient = async () => {
+      try {
+        await mcpClient.connect(transport);
+
+        if (closed) {
+          console.log('Transport has been closed.');
+
+          return;
+        }
```
kapetr, line 34: Just wondering if this is needed at all. This can happen either on component unmount (but then we don't mind a client with the closed transport is set) or if the dependencies change (but `serverUrl` should never change). But we can keep it to be safe.

apps/beeai-ui/src/vite-env.d.ts
```diff
@@ -1 +1,6 @@
 /// <reference types="vite/client" />
+/// <reference types="vite-plugin-svgr/client" />
+
+interface ImportMetaEnv {
+  readonly VITE_MCP_CLIENT_SERVER_URL: string;
```
kapetr, line 5: We should add `.env.example` file with the default server url: http://localhost:5173/mcp/sse
And we could also add the `.env `file creation step to the README.

PR #62, opened 2025-02-10
chore: switch to a published acp package

jezekra1:


pnpm-lock.yaml
JanPokorny, line 1: I think that we still need the lockfile for `beeai-ui`

apps/beeai-server/src/beeai_server/services/mcp_proxy/proxy_server.py
```diff
@@ -102,18 +114,20 @@ async def call_tool(name: str, arguments: dict | None = None):
             )
             return resp.content
 
+        @server.create_agent()
+        async def create_agent(req: CreateAgentRequest) -> CreateAgentResult:
+            provider = self._client_container.routing_table[f"agent_template/{req.params.templateName}"]
+            return await self._send_request_with_token(provider.session, server, req, CreateAgentResult)
```
pilartomas, line 120: `agent_template`? would except just `agent`

pnpm-lock.yaml
jezekra1: the lockfile wasn't removed, I generated it using `mise setup`, is it wrong?

apps/beeai-server/src/beeai_server/services/mcp_proxy/proxy_server.py
jezekra1: Not sure I understand what you mean, when creating agent you need to specify its template. This is how we currently route - by names of the objects in request.

pnpm-lock.yaml
JanPokorny: Oh nevermind, it just looked removed in the UI since it didn't show the new content

PR #63, opened 2025-02-10
feat(agent): add GPT Researcher agent

xjacka:
terminal 1
```script
cd agents/community/gpt-researcher
uv run gpt-researcher-agent
```

terminal 2
```script
brew install i-am-bee/beeai/beeai
brew services start i-am-bee/beeai/beeai
beeai provider add mcp http://localhost:8000/sse
beeai agent list
beeai agent run GPT-researcher "why is Nvidia stock going up?"
```

### DIFF
--- a/agents/official/bee-agent-framework/src/server.ts
+++ b/agents/official/bee-agent-framework/src/server.ts
@@ -6,9 +6,9 @@ import { StreamlitAgent } from "bee-agent-framework/agents/experimental/streamli
 import { OllamaChatLLM } from "bee-agent-framework/adapters/ollama/chat";
 import { UnconstrainedMemory } from "bee-agent-framework/memory/unconstrainedMemory";
 import { Version } from "bee-agent-framework";
-import { runAgentProvider } from 'beeai-sdk/src/providers/agent.js';
-import { promptInputSchema, promptOutputSchema, PromptOutput } from 'beeai-sdk/src/schemas/prompt.js';
-import { Metadata } from 'beeai-sdk/src/schemas/metadata.js';
+import { runAgentProvider } from 'beeai-sdk/src/beeai_sdk/providers/agent.js';
+import { promptInputSchema, promptOutputSchema, PromptOutput } from 'beeai-sdk/src/beeai_sdk/schemas/prompt.js';
+import { Metadata } from 'beeai-sdk/src/beeai_sdk/schemas/metadata.js';
 
 async function registerAgents(server: McpServer) {
   const streamlitMeta = new StreamlitAgent({


agents/community/gpt-researcher/gpt_researcher_agent.py
```diff
@@ -0,0 +1,35 @@
...
+from mcp.server.fastmcp import FastMCP
+from pydantic import BaseModel
+from dotenv import load_dotenv
+load_dotenv()
+
+class Input(BaseModel):
+    prompt: str
+
+class Output(BaseModel):
+    text: str
```
pilartomas: You can now leverage beeai sdk for i/o types: https://github.com/i-am-bee/beeai-prototype/pull/65

Hopefully, the `uv` supports cross-workspace imports the same way pnpm does.

agents/community/gpt-researcher/gpt_researcher_agent.py
```diff
@@ -0,0 +1,35 @@
+from gpt_researcher import GPTResearcher
+from mcp.server.fastmcp import FastMCP
+from pydantic import BaseModel
+from dotenv import load_dotenv
+load_dotenv()
```
pilartomas, line 6: Usually we'd prefer pydantic settings over dotenv module. However, this seems to be only about configurating gpt_researcher so this is fine 👍 

https://docs.pydantic.dev/latest/api/pydantic_settings/

agents/community/gpt-researcher/.env-example
```diff
@@ -0,0 +1,9 @@
+RETRIEVER=duckduckgo
+
+OLLAMA_BASE_URL=http://localhost:11434
+
+FAST_LLM=ollama:llama3.1
+SMART_LLM=ollama:llama3.1
+STRATEGIC_LLM=ollama:llama3.1
+
+EMBEDDING=ollama:nomic-embed-text
```
pilartomas: commit the .env, it must be there for uv to be able to run it.

agents/community/gpt-researcher/gpt_researcher_agent.py
```diff
@@ -0,0 +1,35 @@
...
+
+class Input(BaseModel):
+    prompt: str
+
+class Output(BaseModel):
+    text: str
+    
+def main() -> int:
+    print("Starting server")
```
pilartomas: prefer logging module for this

agents/community/gpt-researcher/gpt_researcher_agent.py
jezekra1: It does not and we don't want to add it to the same workspace to avoid dependency management hell.

Probably the best option here is to use good ol' filesystem dependency or just build and publish the sdk package:
https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies

jezekra1: Also it's better if the example is "self-contained" so you can copy-paste it into your project without the need for extra work.

agents/community/gpt-researcher/gpt_researcher_agent.py
```diff
@@ -0,0 +1,45 @@
...
+    server = FastMCP("researcher-agent")
+    
+    @server.agent("GPT-researcher", 
+                  "GPT Researcher is an autonomous agent designed for comprehensive web and local research on any given task.", 
+                  input=PromptInput, 
+                  output=PromptOutput)
+    async def run_agent(input: PromptInput, ctx) -> PromptOutput:
+
+        async def send_progress(text: str):
+            await ctx.report_agent_run_progress(text)
```
pilartomas:
```
PromptOutput(text=text)
```

xjacka: There is a wrong type, the text is not a string.

Issue #30, opened 2025-02-03
List of agents


tomkis:
- Query searching
- Tag searching
- Sorting (Cost / Name... ?)
- Show human readable name + desc
- Estimated costs (number of tokens + time)

PR #70, opened 2025-02-12
feat(ui): add agents filter, agent detail page

kapetr:
Refs #30

### DIFF
--- a/agents/official/bee-agent-framework/src/server.ts
+++ b/agents/official/bee-agent-framework/src/server.ts
@@ -51,7 +51,7 @@ async function registerAgents(server: McpServer) {
     },
     {
       title: "Streamlit Agent",
-      framework: "BeeAI",
+      framework: "CrewAI",
       licence: "Apache 2.0",
       fullDescription: `This is an example AI agent.
 ## Features
@@ -96,7 +96,7 @@ async function registerAgents(server: McpServer) {
     },
     {
       title: "GPT Researcher",
-      framework: "CrewAI",
+      framework: "BeeAI",
       licence: "Apache 2.0",
       fullDescription: `GPT Researcher is an autonomous agent designed for comprehensive web and local research on any given task.
 
--- a/apps/beeai-ui/package.json
+++ b/apps/beeai-ui/package.json
@@ -37,6 +37,7 @@
     "@eslint/js": "^9.17.0",
     "@types/react": "^19.0.8",
     "@types/react-dom": "^19.0.3",
+    "@types/hast": "^3.0.4",
     "@vitejs/plugin-react-swc": "^3.5.0",
     "eslint": "catalog:",
     "eslint-plugin-react-hooks": "^5.0.0",
--- a/apps/beeai-ui/src/components/MarkdownContent/MarkdownContent.module.scss
+++ b/apps/beeai-ui/src/components/MarkdownContent/MarkdownContent.module.scss
@@ -30,6 +30,6 @@
   }
   p {
     font-size: rem(14px);
-    margin-block: $spacing-03;
+    margin-block: $spacing-05;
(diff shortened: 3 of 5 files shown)


apps/beeai-ui/src/App.tsx
```diff
@@ -7,7 +7,9 @@ import { AppLayout } from './components/layouts/AppLayout';
 import { MCPClientProvider } from './contexts/MCPClient/MCPClientProvider';
 import { Home } from './pages/Home';
 import { NotFound } from './pages/NotFound';
-import { routes } from './utils/routes';
+import { routesDefinition } from './utils/router';
```
PetrBulanek, line 10: I'd rather name it `routeDefinitions`. :-)

apps/beeai-ui/src/components/ConfirmDialog/ConfirmDialog.module.scss
PetrBulanek, line 1: I know we did use the `$prefix` variable in previous project, but I kinda think it is not necessary and I just used `cds` know, because it is a bit more clear. I am open to discussion, but one way or another, we should make it the same across all files.

apps/beeai-ui/src/components/ErrorMessage/ErrorMessage.module.scss
```diff
@@ -0,0 +1,22 @@
+.root.root {
```
PetrBulanek, line 1: Do we need this? I hope it will work correctly in Vite. :-)

apps/beeai-ui/src/components/layouts/AppHeader.module.scss
```diff
@@ -1,10 +1,10 @@
-@use 'sass:math';
-
 .root {
-  position: sticky;
+  position: fixed;
```
PetrBulanek, line 2: Any particular reason for this change?

apps/beeai-ui/src/modules/agents/api/queries/useAgent.ts
PetrBulanek, line 1: I think we could return the whole query here and pass select param into useListAgents instead? What do you think?

apps/beeai-ui/src/modules/agents/api/types.ts
```diff
@@ -1,5 +1,15 @@
...
+
+export type Agent = SdkAgent & {
+  title?: string;
+  fullDescription?: string;
+  avgRunTimeSeconds?: number;
+  avgRunTokens?: number;
+  framework?: string;
+  licence?: string;
+  authors?: string[];
+};
```
PetrBulanek, line 13: Can't we infer this from `packages/beeai-sdk/src/beeai_sdk/schemas/metadata.ts`?

apps/beeai-ui/src/components/ConfirmDialog/ConfirmDialog.module.scss
kapetr: I agree, let's use just "cds".

apps/beeai-ui/src/components/ErrorMessage/ErrorMessage.module.scss
kapetr: We don't, we had wrong order of imports in `main.tsx` :)

apps/beeai-ui/src/components/layouts/AppHeader.module.scss
kapetr: I guess I was trying a different layout and sticky position didn't play well with it. The reason was mainly to downshift the scrollbar to the main element with `scrollbar-gutter: stable` (because that didn't play well on the root with modals). But now it should work with sticky position well, I'll revert it.

PR #72, opened 2025-02-13
feat: redirect users to index.html on 404 in the UI

jezekra1:


apps/beeai-server/src/beeai_server/__init__.py
```diff
@@ -1,6 +1,11 @@
 import os
 import sys
 
+# configure logging before importing anything
+from beeai_server.logging_config import configure_logging
+
+configure_logging()
```
jezekra1, line 7: This actually makes sure that logging is initialized before anything, the imports in `application.py` are often getting out of hand.

apps/beeai-server/src/beeai_server/routes/mcp_sse.py
```diff
@@ -67,6 +67,3 @@ async def handle_sse(request: Request):
             Mount("/messages/", app=sse.handle_post_message),
         ],
     )
-
-
-app = create_sse_app()
```
jezekra1, line 72: I removed this global object because it breaks if it's imported before `bootstrap_dependencies` is called.

I think it's better to avoid this kind of heavy "global object pattern" in general (e.g. if it uses DI or performs heavy computation). We can still use global object pattern for constants and other lightweight stuff:

See:
https://python-patterns.guide/python/module-globals/

PR #74, opened 2025-02-13
fix(gpt-agent): add metadata

jezekra1:


agents/official/bee-agent-framework/src/server.ts
```diff
@@ -93,47 +93,7 @@ async function registerAgents(server: McpServer) {
       return {
         text: output.result.raw,
       };
-    },
-    {
-      title: "GPT Researcher",
-      framework: "BeeAI",
```
kapetr, line 99: Could you also update the Streamlit agent `framework` to "BeeAI"? Not sure if it's true just for the sake of displaying the BeeAI green badge.

Issue #32, opened 2025-02-03
Chat Agent UI


tomkis:
- Streaming
  - User input for prompt
  - Reset session

PR #88, opened 2025-02-14
feat(ui): chat initial

kapetr:
Refs #32

### DIFF
--- a/packages/beeai-sdk/tsconfig.json
+++ b/packages/beeai-sdk/tsconfig.json
@@ -11,7 +11,7 @@
     // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
 
     /* Language and Environment */
-    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "target": "es2016",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
     // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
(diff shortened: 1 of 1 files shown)


apps/beeai-ui/src/modules/run/chat/schema.ts
PetrBulanek, line 1: Don't we have these types somewhere already to import?

apps/beeai-ui/src/utils/router.ts
```diff
@@ -2,9 +2,11 @@ export const routesDefinition = {
   home: () => '/' as const,
   notFound: () => '/not-found' as const,
   agentDetail: () => '/agents/:agentName' as const,
+  agentRun: () => '/run/:agentName' as const,
 };
 
 export const routes = {
   ...routesDefinition,
   agentDetail: (agentName: string) => `/agents/${agentName}`,
+  agentRun: (agentName: string) => `/run/${agentName}`,
```
PetrBulanek, line 11: Just an idea - what about using object as param? It has no effect here, but if we have routes with multiple parameters, it could be better. :-)

apps/beeai-ui/src/modules/run/chat/schema.ts
kapetr: We do kind of, but I think we might still want define the minimal required schema for the specific UI ourselves. This ended being unused, I'll remove it for now.

apps/beeai-ui/src/utils/router.ts
kapetr: Agree

PR #96, opened 2025-02-17
feat: add hooks and checks

JanPokorny:
- add pre-commit hook installation to `mise setup` (hook is `mise git-hooks:pre-commit`)
- add `mise` tasks `check:py`, `check:js`, `fix:py`, `fix:js` which run appropriate linters and checkers
- update `package.json` where appropriate to include `check` and `fix` tasks
- Python uses `ruff` globally
- Type checking for Python as well as any test running is not set up yet
- Globalize appropriate `eslint` & friends versions
- Set minimal supported Python version to 3.11 -- practical to unify over one version given how venvs work; also practical for users that install with `pip` and don't have the most up to date Python -- 3.11 is the current version in Debian stable branch so a good pick for what to support

First commit is config changes, second is changes needed to pass the new checks.


jezekra1:
Great work! 👏 

I think we'll have to change some things for 3.11, but it shouldn't be a problem.

Notice that I added an optional python version to the manifest and we are now using `uvx --python <version> ...` because some projects like crew ai did not run on 3.13. We can now actually require python <3.13 to make this faster (avoid downloading python again in many cases), although python 3.13 is faster then 3.12 overall :)

jezekra1 (approved):

PR #100, opened 2025-02-18
feat(server): add official provider registry

jezekra1:
Ref: #71

### DIFF
--- a/agents/official/bee-agent-framework/src/server.ts
+++ b/agents/official/bee-agent-framework/src/server.ts
@@ -1,4 +1,4 @@
-#!/usr/bin/env npx -y tsx
+#!/usr/bin/env npx -y tsx@latest
 
 import { AcpServer } from "@i-am-bee/acp-sdk/server/acp.js";
 
--- a/apps/beeai-server/src/beeai_server/configuration.py
+++ b/apps/beeai-server/src/beeai_server/configuration.py
@@ -5,6 +5,8 @@ from pathlib import Path
 from pydantic import BaseModel, field_validator, ValidationError, Field, model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+from beeai_server.utils.github import GithubUrl
+
 
 class LoggingConfiguration(BaseModel):
     level: int = logging.INFO
@@ -26,7 +28,9 @@ class Configuration(BaseSettings):
     model_config = SettingsConfigDict(env_file=".env", env_file_encoding="utf-8", env_nested_delimiter="__")
     logging: LoggingConfiguration = LoggingConfiguration()
     provider_config_path: Path = Path.home() / ".beeai" / "providers.yaml"
+    cache_dir: Path = Path.home() / ".beeai" / "cache"
     port: int = 8333
+    provider_registry_location: GithubUrl = "https://github.com/i-am-bee/beeai@main#path=provider-registry.yaml"
 
 
 @cache
--- a/apps/beeai-server/src/beeai_server/crons/sync_registry_providers.py
+++ b/apps/beeai-server/src/beeai_server/crons/sync_registry_providers.py
@@ -27,7 +27,7 @@ async def check_official_registry(configuration: Configuration, provider_service
         headers={"Cache-Control": "no-cache, no-store, must-revalidate", "Pragma": "no-cache", "Expires": "0"}
     ) as client:
         resp = await client.get(str(registry.get_raw_url()))
-        resp = set(yaml.safe_load(resp.content))
+        resp = set(yaml.safe_load(resp.content)["providers"])
         for provider in resp:
             try:
                 provider_manifest = CreateProviderRequest(location=provider).location
--- a/apps/beeai-server/src/beeai_server/crons/sync_registry_providers.py
+++ b/apps/beeai-server/src/beeai_server/crons/sync_registry_providers.py
@@ -1,0 +1,58 @@
+import logging
+from datetime import timedelta
+
+import httpx
+import yaml
+from kink import inject
+
+from beeai_server.configuration import Configuration
+from beeai_server.schema import CreateProviderRequest, DeleteProviderRequest
+from beeai_server.services.provider import ProviderService
+from beeai_server.utils.periodic import periodic
+
+logger = logging.getLogger(__name__)
+
+
+@periodic(period=timedelta(minutes=5))
+@inject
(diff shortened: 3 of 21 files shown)


agents/official/bee-agent-framework/src/server.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env npx -y tsx
+#!/usr/bin/env npx -y tsx@latest
```
JanPokorny, line 1: This shebang will only work on macOS.

macOS parses it as: `["/usr/bin/env", "npx", "-y", "tsx@latest"]`

Linux parses it as: `["/usr/bin/env", "npx -y tsx@latest"]`, but that does not work, since it searches for an executable called `npx -y tsx@latest`.

Fix by using `-S` (pseudo-option in `env` that mitigates this):

```suggestion
#!/usr/bin/env -S npx -y tsx@latest
```

apps/beeai-server/src/beeai_server/domain/model.py
```diff
@@ -99,19 +102,26 @@ class NodeJsProvider(ManagedProvider):
...
     @asynccontextmanager
-    async def mcp_client(self) -> McpClient:  # noqa: F821
-        async with super()._get_mcp_client(command=["npx", "-y", self.package, *self.command]) as client:
+    @inject
+    async def mcp_client(self, configuration: Configuration) -> McpClient:  # noqa: F821
+        try:
+            github_url = GithubUrl.model_validate(self.package)
+            repo_path = await download_repo(configuration.cache_dir / "github_npm", github_url)
+            package_path = repo_path / (github_url.path or "")
+            await anyio.run_process(["npm", "install"], cwd=package_path)
```
jezekra1, line 120: This is ugly, `npm/x` with tsx and filesystem/github dependencies drives me crazy. To download the repository and install dependencies locally is only way how I managed to make it work

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -26,7 +28,9 @@ class Configuration(BaseSettings):
     model_config = SettingsConfigDict(env_file=".env", env_file_encoding="utf-8", env_nested_delimiter="__")
     logging: LoggingConfiguration = LoggingConfiguration()
     provider_config_path: Path = Path.home() / ".beeai" / "providers.yaml"
+    cache_dir: Path = Path.home() / ".beeai" / "cache"
```
JanPokorny, line 31: Perhaps we could be nice and switch to XDG standard dirs for config/state/cache (https://specifications.freedesktop.org/basedir-spec/latest/)? Looking at my home dir, it looks like many tools use them even on Mac (i.e. `pnpm`) -- and many don't and just go into `~/.<name>` (i.e. `npm`).

apps/beeai-server/src/beeai_server/crons/sync_registry_providers.py
```diff
@@ -0,0 +1,58 @@
...
+
+    async with httpx.AsyncClient(
+        headers={"Cache-Control": "no-cache, no-store, must-revalidate", "Pragma": "no-cache", "Expires": "0"}
+    ) as client:
+        resp = await client.get(str(registry.get_raw_url()))
+        resp = set(yaml.safe_load(resp.content))
+        for provider in resp:
+            try:
+                provider_manifest = CreateProviderRequest(location=provider).location
+                await provider_manifest.resolve()
```
JanPokorny, line 34: Would parallelizing this step make sense? It's in a for loop but awaited sequentially, so it might take longer for many providers. Same for the two loops below.

apps/beeai-server/src/beeai_server/utils/github.py
```diff
@@ -82,3 +110,57 @@ def __str__(self):
...
+        return True
+    return False
+
+
+def extract_targz_safe(tar_path: PathLike | Path, extract_path: PathLike | Path):
+    with tarfile.open(tar_path, "r:gz") as tar:
+        members = []
+        for member in tar.getmembers():
+            if member.linkpath or member.linkname:
+                logger.warning(f"Cannot extract member: {member.name}, symlinks are forbidden")
```
JanPokorny, line 131: Actually what's the threat model here? We run the executables with no sandboxing anyway 😁

provider-registry.yaml
```diff
@@ -0,0 +1,2 @@
+- git+https://github.com/i-am-bee/beeai@main#path=/agents/community/gpt-researcher-agent
+- git+https://github.com/i-am-bee/beeai@main#path=/agents/official/bee-agent-framework
```
JanPokorny: Maybe we want to go a bit more verbose in case we need to extend later (add more config keys)?

```suggestion
providers:
- url: git+https://github.com/i-am-bee/beeai@main#path=/agents/community/gpt-researcher-agent
- url: git+https://github.com/i-am-bee/beeai@main#path=/agents/official/bee-agent-framework
```

JanPokorny (approved):
Looks great, I left some small comments but overall it looks solid

apps/beeai-server/src/beeai_server/crons/sync_registry_providers.py
jezekra1: Can be done, but the operations should be quite fast - fetch default branch from github if not specified, download provider manifest and write it to a file.

The actual work of fetching dependencies etc, spawning process etc is triggered, but it will happen asynchronously, so I wouldn't complicate this and I'm not sure whether github rawusercontent has some concurrent requests quotas

apps/beeai-server/src/beeai_server/configuration.py
jezekra1: I thought about it, but overall I think it's better for the users to just have everything in one place so that you can do `rm -rf ~./beeai` if things go wrong and you have to reset. Also makes for a nicer troubleshooting guide :D  

I'd keep it as is now but consider moving both eventually if we want to be compliant - config to:
`~/.config/beeai` and cache to `~/.cache/beeai`

apps/beeai-server/src/beeai_server/utils/github.py
jezekra1: I think you can unzip to some unwanted paths, like replace executable in `~/local/.bin` :D tar allows both absolute paths (from fs root) and relative paths (`../../`) and links, see this big red warning :D 

https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall

Zip does not have this issue, but it does not preserve fs permissions and we need some files executable

PR #101, opened 2025-02-18
feat(ui): chat improvements

kapetr:
This PR includes:
- Feedback from #88 & #70
- Ability to abort a run
- Generalized `useRunAgent` hook
- Various improvements


apps/beeai-ui/src/modules/agents/detail/AgentDetail.tsx
```diff
@@ -44,7 +46,16 @@ export function AgentDetail({ name }: Props) {
...
-                <Button kind="primary" renderIcon={ArrowUpRight} size="md" href={routes.agentRun(name)}>
+                <Button
+                  kind="primary"
+                  renderIcon={ArrowUpRight}
+                  size="md"
+                  href={routes.agentRun({ name })}
+                  onClick={(e) => {
+                    e.preventDefault();
+                    navigate(routes.agentRun({ name }));
+                  }}
```
PetrBulanek: Any special reason for this? I think we can use `<Button as={Link} href="..." ... />` where `Link` comes from react-router.

apps/beeai-ui/src/App.tsx
```diff
@@ -2,12 +2,11 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { ErrorBoundary } from 'react-error-boundary';
 import { BrowserRouter, Route, Routes } from 'react-router';
 import { ErrorFallback } from './components/fallbacks/ErrorFallback';
-import { MCPFallback } from './components/fallbacks/MCPFallback';
```
PetrBulanek, line 5: We can probably delete the component as well.

apps/beeai-ui/src/hooks/useImmerWithGetter.ts
PetrBulanek, line 1: The `valueRef` is not updated anywhere, so it might contain stale value. Or am I missing anything? Also when should I use `value` and when the `valueRef` getter?

apps/beeai-ui/src/modules/agents/api/queries/useAgent.ts
```diff
@@ -1,20 +1,27 @@
...
-  return {
-    agent,
-    isPending,
-    refetch,
-    isRefetching,
-    error: error ?? (data && !agent ? new Error('Agent not found.') : undefined),
-  };
+      if (!agent) {
+        throw new Error('Agent not found.');
+      }
```
PetrBulanek: Wouldn't it be better to return null instead of throwing an error?

apps/beeai-ui/src/modules/agents/api/queries/useAgent.ts
```diff
@@ -1,20 +1,27 @@
...
-  const { data, isPending, error, refetch, isRefetching } = useListAgents();
+  const client = useMCPClientContext();
 
-  const agent = useMemo(() => data?.find((item) => name === item.name), [data, name]);
+  return useQuery({
+    queryKey: agentKeys.list(),
+    queryFn: () => client!.listAgents(),
+    enabled: Boolean(client),
+    select: (data) => {
+      const agent = data?.agents.find((item) => name === item.name);
```
PetrBulanek, line 18: I originally thought to add `select` function param into the `useListAgents` and reuse it instead of whole new `useQuery` call:

```
const { data } = useListAgents({
  select: (data) => ...
})
```

But it is just a small code duplication, so I guess it is not necessary.

apps/beeai-ui/src/modules/run/api/mutations/useRunAgent.tsx
```diff
@@ -0,0 +1,47 @@
+import { useMutation } from '@tanstack/react-query';
+import { Agent } from '@i-am-bee/acp-sdk/types.js';
+import z, { ZodLiteral, ZodObject } from 'zod';
+import { useCreateMCPClient } from '@/api/mcp-client/useCreateMCPClient';
+
+interface Props<
+  NotificationsSchema extends ZodObject<{
+    method: ZodLiteral<string>;
+  }>,
```
PetrBulanek, line 9: Where is the NotificationsSchema coming from?

Also we could probably wrap it into separate type as it is reused on lines 20-22.

PetrBulanek: Just noticed it is a generic. Nevermind.

PetrBulanek (approved):

PetrBulanek (approved):

apps/beeai-ui/src/modules/agents/detail/AgentDetail.tsx
kapetr: Of course, I forgot it can extend its props with those from `as`.

apps/beeai-ui/src/App.tsx
kapetr: I did delete it.

apps/beeai-ui/src/hooks/useImmerWithGetter.ts
kapetr: I see the getter is not used anywhere for now, but it's convenient to have it and use it, when possible (mainly in callbacks), because `messages` get updated a lot and using the value can make components re-render unnecessarily. 
The `useEffect` that updates the ref value got lost - good catch, thanks!

apps/beeai-ui/src/modules/agents/api/queries/useAgent.ts
kapetr: I know, but I figured it's a small code duplication, and the typing could be tricky (you're changing the query return data type - but I haven't tried it).

PetrBulanek (approved):

PR #124, opened 2025-02-19
fix(server): properly kill orphaned processes

jezekra1:


apps/beeai-server/src/beeai_server/utils/managed_server_client.py
```diff
@@ -37,6 +39,21 @@ class ManagedServerParameters(BaseModel):
     endpoint: str = "/sse"
 
 
+def _kill_process(process: anyio.abc.Process):
+    import psutil
+
+    parent = psutil.Process(process.pid)
+    for child in parent.children(recursive=True):  # or parent.children() for recursive=False
```
JanPokorny: Not sure what the comment means, I think we always want to go recursive (providers can spawn more processes).

apps/beeai-server/src/beeai_server/utils/managed_server_client.py
```diff
@@ -37,6 +39,21 @@ class ManagedServerParameters(BaseModel):
...
+        with suppress(psutil.NoSuchProcess):
+            child.kill()
+            child.wait()
+    with suppress(psutil.NoSuchProcess):
```
JanPokorny: Are we guaranteed to kill all the children by this point? `parent.children()` returns a list, so the result is up-to-date only at the start of iteration, new processes could be spawned in the tree during that time. Not sure how (or whether) to fix this, since there are always timing issues with this approach.

apps/beeai-server/src/beeai_server/utils/managed_server_client.py
jezekra1: copy-paste from stack overflow :D

apps/beeai-server/src/beeai_server/utils/managed_server_client.py
jezekra1: I used the killpg approach after all

PR #128, opened 2025-02-19
feat(server): pass env variables to providers

jezekra1:


apps/beeai-server/src/beeai_server/domain/model.py
```diff
@@ -213,6 +240,19 @@ class Provider(BaseModel):
...
+
+    @asynccontextmanager
+    async def mcp_client(self) -> McpClient:
+        async with self.manifest.mcp_client(env=self.env) as client:
+            yield client
+
+    @model_validator(mode="after")
+    def level_uvicorn_validator(self):
+        required_vars = set(var.name for var in self.manifest.env if var.required)
+        if missing_vars := required_vars - (self.env or {}).keys():
```
jezekra1, line 253: Should we implicitly pass the env if it's not found in the provider config, but it's found in actual `os.environ`?

I'd keep it explicit like this:

Implicit passing would make validations harder, you can successfully add a provider when the server env is present, but then restart the server in different environment and suddenly you have invalid configuration.

Also, server in brew won't have any special env variables.

PR #137, opened 2025-02-20
feat(ui): add page transition animations

kapetr:


apps/beeai-ui/src/modules/agents/detail/AgentDetail.module.scss
```diff
@@ -1,17 +1,20 @@
 .root {
   padding-block-start: $spacing-05;
+  display: flex;
+  flex-direction: column;
```
PetrBulanek, line 4: Is this needed? :-)

PetrBulanek: Aha, I guess, it's because the animation. :)

PetrBulanek (approved):

apps/beeai-ui/src/modules/agents/detail/AgentDetail.module.scss
kapetr: Yes, exactly

PR #157, opened 2025-02-24
fix(server): docker build & run

pilartomas:


.github/workflows/beeai-server.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
     tags:
       - "beeai-server-v*"
+  pull_request:
```
JanPokorny: What's the purpose of this? Is it just for testing now?

apps/beeai-server/src/beeai_server/__init__.py
```diff
@@ -46,7 +46,7 @@ def serve():
     os.execv(
         sys.executable,
         [
-            "python",
+            sys.executable,
```
JanPokorny, line 49: Weirdly this really makes a difference.

.github/workflows/beeai-server.yml
pilartomas: yes

pilartomas (commented):
looks good

PR #183, opened 2025-02-25
feat(ui): add agents import

PetrBulanek:
- if the user closes the modal during the import, the notification will not pop up (needs to be fixed)


apps/beeai-ui/src/modules/agents/components/ImportAgentsModal.tsx
```diff
@@ -14,59 +14,159 @@
...
 
 export function ImportAgentsModal({ onRequestClose, ...modalProps }: ModalProps) {
   const id = useId();
-  const { addToast } = useToast();
-  const { mutate, isPending } = useImportProvider({
-    onSuccess: () => {
-      addToast({ title: 'Provider was imported successfuly' });
-      onRequestClose();
+  const [createdProviderId, setCreatedProviderId] = useState<string>();
+  const [providerSource, setProviderSource] = useState(ProviderSource.LocalPath);
```
kapetr: Is there a reason for not including `providerSource` in the `useForm` values?

kapetr (approved):
The modal close button is off:
Screenshot 2025-02-25 at 16 19 29

apps/beeai-ui/src/modules/agents/components/ImportAgentsModal.tsx
PetrBulanek: It is just a UI "helper" and actually nothing that is processed outside this component or sent to server, so I thought the form will be cleaner.

PetrBulanek: But there's no other deeper reason for it, so we might as well change it if you have a strong opinion on it. :-)

kapetr: Since you actually use it in `onSubmit` of the form, it would be cleaner to be present in the form values. In very edge (probably almost impossible :) ) scenario you could have an old value in onSubmit, because the state haven't updated yet. And it actually is a form item. But no big deal :)

kapetr: And the value actually is sent to the server as part of the `location` prop

PetrBulanek: You are right. I will fix it. :)

kapetr (approved):

Issue #182, opened 2025-02-25
Agent detail


tomkis:
## Feature Description

Show a detail of agent based on data provided from ACP.

### Detail Page

Provide heading of the page with the human readable name of the agent.

All the metadata should be visualised namely:

- Agent short description
- Avg runtime
- Avg token consumption
- Licence
- Tags for framework and technology

Each agent may contain detailed description which should be visible only on detail page.

Clicking on agents in the top bar would bring the user back to the list of agents.

### Command Component

There's `beeai run` command component that simplifies running the agent in local CLI.

This component will contain the full command to run the agent locally as well as provide easy way to copy the command.

When user hover over the copy, the copy button will fills with black. Upon clicking, the button changes to a checkmark, indicating a successful copy action

tomkis:
@anafucs please review & attach figma files.

anafucs:
Figma here: https://www.figma.com/design/uJkcE3aJRPVk9HBOweNnH9/Bee-Dev-Ready-2.0?node-id=2906-10038

anafucs:
quick QA and updates:
- can we reset the scroll position so it always launches the page y=0?
- Bee badge next to title
- Github icon instead of star on info line, include link to repo (underlines text on hover)
- include tag for type (RAG, researcher, chat, code, canvas)
Figma file is updated!

jezekra1:
I added improved CLI `beeai agent info <name>` and `beeai agent info <name> --schema` pages ^^

Issue #181, opened 2025-02-25
List all available agents


tomkis:
## Feature Description

Implement list of agents mimicking the data provided by the ACP client. 

### Agent List

While data is being fetched, the loading placeholder is shown.

Each agent should display following data:

- Agent name (human readable name)
- Agent description
- Avg runtime
- Avg token consumption
- Licence
- Github stars (if available)
- Tags for framework and technology
- Tag for verified

Each agent should provide a link to github where the implementation is located or the link to the provider

Because the initial list of agents will be limited, we don't need to implement virutalized list of agents.

There's an information about number of shown agents in the list.

MCP does not provide pagination capability so we need to show all agents at once.

Once user starts scrolling there's an arrow to bring user back to the start of the page.

### Filtering

Focusing the search field will visually emphasize the field.

We should allow fulltext search that would filter the results based on agent human readable name as well as description.

MCP does not provide any functionality to expose filtering, sorting and pagination so it all has to be client side.

Aggregate all frameworks/runtimes to create tags which can then be used to filter the results. 

Tag filter is single select, so eg. clicking on different tag would cancel out the previous filter. 

Number of agent will adjust based on the filter in the form of

> Showing X of Y agents

tomkis:
@anafucs can you please review & attach Figma designs?

anafucs:
I directly added on the description. The designs are here: https://www.figma.com/design/uJkcE3aJRPVk9HBOweNnH9/Bee-Dev-Ready-2.0?node-id=2906-10039
Minor changes TBD after design review (tomorrow):
- Bee icon for verified agents
- GitHub link (keep or remove)

PR #185, opened 2025-02-25
feat(web): agents pages

Zycon42:
it needs new env like
```
BEEAI_HOST=http://localhost:8333
```

Refs: #181, #182


apps/beeai-web/src/acp/client.ts
```diff
@@ -0,0 +1,50 @@
...
+  name: "example-client",
+  version: "1.0.0",
+};
+const ACP_EXAMPLE_AGENT_PARAMS = {
+  capabilities: {
+    prompts: {},
+    resources: {},
+    tools: {},
+  },
+};
```
pilartomas: These are not client capabilities (they are used as such in the client ctor). This client has no capabilities so `capabilities` would be an empty object.

apps/beeai-web/src/acp/client.ts
```diff
@@ -0,0 +1,50 @@
...
+        if (init != null && init.cache != null) {
+          delete init.cache;
+        }
+        return fetch(url, init);
+      },
+    },
+  });
+  const client = new Client(ACP_EXAMPLE_AGENT_CONFIG, ACP_EXAMPLE_AGENT_PARAMS);
+  await client.connect(transport);
+  return client;
```
pilartomas: After connecting, there should be some sort of check that the server supports `agents` capability if agent-related methods are to be used on it (like `listAgents`).

apps/beeai-web/src/acp/api.ts
```diff
@@ -0,0 +1,33 @@
...
+import { getAcpClient } from "./client";
+
+export const getAgentsList = unstable_cache(
+  async () => {
+    const client = await getAcpClient();
+
+    const { agents } = await client.listAgents();
+
+    await client.close();
+
```
pilartomas: rather close in `finally` block, it is important to ensure the connection closes to avoid leaks on the server.

apps/beeai-web/src/acp/client.ts
pilartomas: There is something like `client.getServerCapabilities` to access it

pilartomas: This is not that important when connecting to the platform server, that one will likely always support agents, tools, etc.

But it will be useful when connecting directly against some agents/tool/both provider.

apps/beeai-ui/src/modules/agents/components/AgentCard.tsx
```diff
@@ -14,32 +14,27 @@
...
+export function AgentCard({ agent, renderTitle }: Props) {
+  const { description } = agent;
   return (
     <article className={classes.root}>
-      <h2 className={classes.name}>
-        <TransitionLink className={classes.link} to={route}>
-          {getAgentTitle(agent)}
-        </TransitionLink>
-      </h2>
+      <h2 className={classes.name}>{renderTitle({ className: classes.link, agent })}</h2>
```
kapetr, line 37: This works, but couldn't we come up with a solution that wouldn't require to do this in every component that has a link? It makes the components less clean and also introduces some code duplication across the projects. 

A solution that comes to mind is to introduce a `RouteProvider` to the root of both projects, that would accept a function with `to` parameter (e.g. `navigate: (to: string) => void` ), that would do the routing. Then we'd create custom `Link` element that would just use the `navigate` function from `RouteProvider` to route the request. What do you think @PetrBulanek @Zycon42 ?

apps/beeai-web/src/acp/client.ts
Zycon42: I just copied it from the beeai-ui https://github.com/i-am-bee/beeai/blob/main/apps/beeai-ui/src/api/mcp-client/useCreateMCPClient.ts, should I change it there as-well?

apps/beeai-web/src/acp/client.ts
Zycon42: If I understood it correctly beeai-web will be used with our hosted version of the server, so it should be granted it supports listing agents no?

apps/beeai-web/src/acp/api.ts
Zycon42: fixed

apps/beeai-ui/src/modules/agents/components/AgentCard.tsx
Zycon42: Link component in nextjs does more than just navigate on a click (e.g. prefetching routes when in viewport/hover), you cannot replace it with just a function, the whole component should be used. However I think we can put the component class in the react context and have a shared link component use that under the hood. I think it will be probably leaky abstraction anyway.

kapetr: That would be another option, but I would rather go the first way. I think we could support `prefetch` as well just by adding `prefetch` function to the context and then using it in our custom link in useEffect. Imho we could mimic the Next `Link` component with all its features this way.

apps/beeai-web/src/acp/client.ts
pilartomas: Please do

apps/beeai-web/src/acp/client.ts
pilartomas: yes, it is really just a nice to have and good pattern it this setup, not a requirement.

Also eventually, we might need to classify the servers based on capabilities, but that is far future.

JanPokorny:
@Zycon42 Regarding env, we wanted to unify this under `BEEAI_HOST` anyway -- it's already defined in `mise.local.toml`

apps/beeai-ui/src/modules/agents/components/AgentCard.tsx
Zycon42: I'm very opposed to not using nextjs Link component, it's maintenance burden, as it's actually not documented what it does beside reading it's source code. It does have over 600 lines of code, it also e.g. contains intersection observer, do you really want to duplicate that and maintain that?

kapetr: I wouldn't want to duplicate everything, just the basic features. But you're right, I don't have the knowledge of what it does in its complexity, so we might lose something. Let's leave it as si then, we can discuss later.

apps/beeai-web/src/acp/client.ts
Zycon42: done

apps/beeai-web/src/acp/client.ts
Zycon42: Not sure I've done it correctly

apps/beeai-ui/src/modules/agents/components/AgentCard.tsx
Zycon42: ok

kapetr (approved):

apps/beeai-web/src/acp/client.ts
pilartomas: I think it's spot on 👍

PetrBulanek (approved):

PR #203, opened 2025-02-27
feat(agents): add composition agent

jezekra1:


agents/official/composition/src/composition/configuration.py
```diff
@@ -0,0 +1,12 @@
+from pydantic import AnyUrl, computed_field
+from pydantic_settings import BaseSettings
+
+
+class Configuration(BaseSettings):
+    platform_url: AnyUrl = "http://localhost:8333"
```
JanPokorny, line 6: I'd unify this everywhere under `BEEAI_HOST` (including `beeai-server`, `beeai-cli` and other projects)

JanPokorny (approved):

Issue #176, opened 2025-02-25
UI - various improvements


matoushavlena:
Figma: https://www.figma.com/design/aAVa3AWNprAf9uwA1LVSoe/Bee-JAM-3.0?node-id=5124-184726&t=Bjy5eUMYWzdREoJi-4

Tasks: 
1. Typography update (smaller fonts)
2. Agent detail - add github link - only if `metadata.githubUrl` field is populated
3. Agent detail - update "Thy this agent" to new text and icon
4. Agent detail - add CLI example - only if `metadata.exampleInput` is populated (please sync with @PetrBulanek on the code block implementation)

PetrBulanek:
@penge For the code block component you can partially reuse the similar component from `bee-ui` https://github.com/i-am-bee/bee-ui/blob/5b0d805d21bf5c63d452a1a9e78bb85b6755cf9c/src/modules/chat/message/MarkdownContent.tsx#L75

PR #208, opened 2025-02-27
chore(ui): update typography

penge:

Most notable change:
Decreased the size of the agent's **name** and **description**.

Other changes:
- the **line** under the agent's **name** is reduced in height to match the smaller font size,
- spacing between the **cards** slightly reduced,
- spacing between the elements inside the **card** updated (more space above tag list),
- fixed `<AgentCard.Skeleton>` where CSS class should be set to the container,
- spacing for the `.divider` line in `<AgentDetail>` is made much smaller.

### Preview

Showcases the updates and several page refreshes.

https://github.com/user-attachments/assets/b46a6125-e481-4ece-8ac9-4d583bdf2d8f

### DIFF
--- a/apps/beeai-ui/src/modules/agents/components/AgentCard.module.scss
+++ b/apps/beeai-ui/src/modules/agents/components/AgentCard.module.scss
@@ -16,7 +16,7 @@
 
 .root {
   border-block-end: 1px solid $border-subtle-00;
-  padding-block: $spacing-06;
+  padding-block: $spacing-05;
   position: relative;
   display: flex;
   flex-direction: column;
@@ -24,13 +24,13 @@
 }
 
 .name {
-  font-size: rem(32px);
-  line-height: math.div(42, 32);
+  font-size: rem(24px);
+  line-height: 1.3;
 }
 
 .link {
   @include link-mask();
-  display: block;
+  display: inline-block;
   color: inherit;
   text-decoration: none;
   max-inline-size: max-content;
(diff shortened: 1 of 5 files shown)


matoushavlena (approved):

apps/beeai-ui/src/modules/agents/components/AgentCard.module.scss
```diff
@@ -68,8 +67,11 @@
 .description {
   max-inline-size: 90%;
   color: $text-secondary;
-  font-size: rem(18px);
-  line-height: math.div(24, 18);
+  font-size: rem(14px);
+  line-height: math.div(24, 14);
+  letter-spacing: 0.32px;
```
PetrBulanek, line 72: We usually use rems also for letter-spacing => `rem(0.32px)` ;-)

penge: This one is tricky.

**First:** Using either `px` or `rem` here, while font is in `rem`, just works the same (browser magic).

**Second:** All Typography from Carbon uses `letter-spacing` in `px`, and I'd say we should stick to that. Example:

```
@include type-style(body-01);
```

Uses:

```
--cds-body-01-letter-spacing: 0.16px;
```

And so on.

(Otherwise we could end up with inconsistent styles, for the better or worse.)

PetrBulanek: You are right. I didn't notice Carbon using px in this case, we were just used to using rem. As far as I know, we use only to values though - 0.16px or 0.32px - so we could at least make them variables and use those. Not needed to do that right now. Just a suggestion for future improvement. :-)

PR #211, opened 2025-02-27
chore(ui): rename button to Launch this agent

penge:

Updating button to **"Launch this agent"** with icon _arrow right_. (Before: "Try this agent" with icon _arrow up right_.)

### Preview

launch-this-agent


matoushavlena (approved):

apps/beeai-ui/src/modules/agents/components/AgentDetail.tsx
```diff
@@ -93,7 +93,7 @@ AgentDetail.Skeleton = function AgentDetailSkeleton() {
 
       <div className={classes.buttons}>
         {/* .cds--layout--size-md fixes Carbon bug where button size prop is not respected */}
-        <ButtonSkeleton size="md" className={clsx('cds--layout--size-md', classes.tryButton)} />
```
PetrBulanek, line 96: I would rather move this skeleton inside the AgentLaunchButton component (we do the same with some other components - for example AgentCard), so we don't need to duplicate the CSS class here.

`<AgentLaunchButton.Skeleton />`

PetrBulanek: Already done that, if you don't mind. 0:-)

penge: I think that is a good suggestion. Keeping the **Skeleton** and the Final close (like in the same file), so the **Skeleton** has access to the same styles (e.g.: height, width) for a **consistent** look.

I have noticed `.copySnippet`, right on the next line has the same flaw: **duplicated** style of the Final, and that can lead to bugs.

Didn't want to go **overboard** here and refactor it all, but I definitely agree with the suggestion and it should be refactored (here and everywhere where needed) at some point, and future Skeleton-s should be done that way from the beginning.

PR #214, opened 2025-02-27
feat(ui): add preview task

penge:
To run the task:

```
$ mise beeai-ui:preview
```


PetrBulanek:
I think it could be convenient to update the `pnpm preview` to build the project first (`vite build && vite preview`). What do you think?

PetrBulanek (approved):

penge:
Sticking to the original behavior, when the `preview` does not perform `build`. I think that is more intuitive.

PR #217, opened 2025-02-28
fix(web): force dynamic build for agent pages

pilartomas:

### DIFF
--- a/apps/beeai-web/src/acp/client.ts
+++ b/apps/beeai-web/src/acp/client.ts
@@ -50,7 +50,7 @@ function assertServerCapability(
   }
 }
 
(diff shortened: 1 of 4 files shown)


apps/beeai-web/src/acp/client.ts
```diff
@@ -50,7 +50,7 @@ function assertServerCapability(
   }
 }
 
-const ACP_SSE_TRANSPORT_URL = new URL("/mcp/sse", BEEAI_HOST);
+const ACP_SSE_TRANSPORT_URL = new URL(BEEAI_HOST);
```
pilartomas, line 53: The path needs to be configurable as well

PR #225, opened 2025-02-28
feat(agents): update aider docs

matoushavlena:


agents/community/aider-agent/src/aider_agent/agent.py
```diff
@@ -17,23 +18,78 @@ class Output(PromptOutput):
     files: dict[str, str] = Field(default_factory=dict)
     text: str = Field(default_factory=str)
 
+exampleInput = {"text": "Make a program that asks for a number and prints its factorial"}
+exampleInputStr = json.dumps(exampleInput, ensure_ascii=False, indent=2)
+
+fullDescription = f"""
+> [!NOTE]
```
JanPokorny: This is not standard Markdown, it's a GitHub extension so it may not work depending on where we render it.

A "poor man's version" is:

> ℹ️ NOTE
> 
> This agent works...

```suggestion
> ℹ️ NOTE
> 
```

JanPokorny (approved):

PR #232, opened 2025-03-02
feat(ui): add Example requests

penge:

Adding a new **"Example requests"** section to the detail page. (This section shows only when example input is provided.)

Other changes:
- Moving full description under the **"Description"** section. Fixing styles in the **"Description"** section (_decreased_ font sizes, _increased_ paddings for lists) to match the new design and new sections.
- Adding a helper to easily create CLI **command**. (In future, can be extended and reused for any commands.)
- Updating `@carbon/react`.

### Preview

https://github.com/user-attachments/assets/1553a74d-57ef-4a3d-8b6c-9361746b2db5


matoushavlena (approved):

apps/beeai-ui/src/components/CopySnippet/CopySnippet.module.scss
```diff
@@ -22,6 +22,15 @@
   border: 1px solid $text-dark;
   border-radius: $border-radius;
   background-color: $layer;
+
+  &.multi {
+    --cds-layer: #{$cool-gray-10};
```
PetrBulanek, line 27: This is probably not dark mode safe. I'd rather use color token - `$layer-02` could work. :-)

apps/beeai-ui/src/styles/components/_tabs.scss
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@use 'styles/common' as *;
+
+.cds--tab--list {
+  --cds-border-interactive: #{$black};
```
PetrBulanek, line 20: This is also probably not dark mode safe. Maybe `$text-dark` might work here? :-)

apps/beeai-ui/src/utils/commands.ts
```diff
@@ -0,0 +1,40 @@
...
+function convertInputToCliArgument(input: string) {
+  try {
+    const parsed = JSON.parse(input);
+    return `'${JSON.stringify(parsed, null, 2)}'`;
+  } catch {
+    return `"${input}"`;
+  }
+}
+
+export default {
```
PetrBulanek, line 28: We are used to not using default imports where possible, please change this to named import.

PR #234, opened 2025-03-02
feat(server): bootstrap telemetry

pilartomas:


apps/beeai-server/src/beeai_server/domain/collector/beeai.yaml
```diff
@@ -0,0 +1,10 @@
+exporters:
+  otlp/beeai:
+    endpoint: telemetry.beeai.com:4317
```
pilartomas: @matoushavlena TDB, I know we agreed on the domain but can't recall what it was exactly.

PR #252, opened 2025-03-03
fix(ui): agent detail design review 1

Zycon42:
- moved bee badge next to the title (on the list too)
- moved github link to the metadata line (as per matous input I didn't bother with stars)

Refs: #182


apps/beeai-ui/src/modules/agents/components/BeeBadge.module.scss
```diff
@@ -14,6 +14,21 @@
...
+  padding-inline-start: rem(7px) !important;
+  padding-inline-end: rem(7px) !important;
+
+  :global(.#{$prefix}--tag__custom-icon) {
+    block-size: rem(18px);
+    inline-size: rem(18px);
+
+    > svg {
+      block-size: rem(18px) !important;
+      inline-size: rem(18px) !important;
```
PetrBulanek, line 31: I think we should avoid `!important` as much as possible. Is it really needed? :-)

Zycon42: Yes it unfortunately is, because the svgs files have width and height directly in them and you need `!important` to override them.

The `!important` in the paddings are there again as a hack because the tag component is not very nice to override when you need bigger icon.

So to remove importants I feel big refactor of tag component styles is needed.

PetrBulanek: Ok, thanks for explaining. Let's leave it as it is.

PR #255, opened 2025-03-04
fix(ui): GitHub rate limit

penge:
**Details:**
- cache result of the github repo query (for **2 hours**)
- cache only github queries at the moment

<br>

### Preview

> For presentation purposes, the limit is set to 20 seconds. (The actual limit is **2 hours**.)
> Preview showcases how query is called at second 1 and then only after it expires at second 21.

https://github.com/user-attachments/assets/a312cbcd-2b38-4139-8af9-0981981bb793


apps/beeai-ui/src/contexts/QueryProvider/QueryProvider.tsx
```diff
@@ -14,16 +14,33 @@
...
+
+const localStoragePersister = createSyncStoragePersister({
+  storage: window.localStorage,
+});
+
+const persistOptions: PersistQueryClientProviderProps['persistOptions'] = {
+  persister: localStoragePersister,
+  dehydrateOptions: {
+    shouldDehydrateQuery(query) {
+      return query.queryKey[0] === gitHubRepoKeys.all()[0];
```
PetrBulanek, line 33: I think, we should not specify the queries that should be persisted here in the provider. It should be the query's itself responsibility. We could utilize the query's meta such as `meta: { persist: true }`. What do you think? cc. @kapetr

PetrBulanek (changes requested):

apps/beeai-ui/src/contexts/QueryProvider/QueryProvider.tsx
penge: I kindly disagree. Meta is used to store additional data about the query to the cache, not to give it a behavior or functionality.

There are benefits in configuring cache from a central place. Most notably:
- Caching policies are applied consistently.
- Single point of control.

kapetr: Both good points, but I'm leaning towards @PetrBulanek proposal, we already use the `meta` to  control invalidation of queries after mutations and we are pretty happy with it. I believe it should be defined within the query itself, so you know exactly what's happening with the data when you look at the query.

PetrBulanek (approved):

PR #263, opened 2025-03-04
fix(beeai-ui): check & fix types

JanPokorny:


apps/beeai-ui/src/modules/agents/components/AgentDetail.tsx
PetrBulanek, line 1: This should not be needed in world with `exampleInput` typed correctly as `string | undefined`. :-)

apps/beeai-ui/src/modules/agents/components/AgentDetailSection.tsx
PetrBulanek, line 1: Same as above.

apps/beeai-ui/src/utils/commands.ts
PetrBulanek, line 1: Same as above.

PetrBulanek: Under the classes.uiNotAvailable div, there should be also check for `agent.ui?.type` and not only `agent.ui`. And it should also go to the message itself.

PetrBulanek (changes requested):

PR #275, opened 2025-03-05
feat(agents): add progress streaming

xjacka:


agents/official/beeai-framework/src/podcast-creator/podcast-creator.ts
```diff
@@ -52,28 +56,40 @@ ALWAYS START YOUR RESPONSE DIRECTLY WITH SPEAKER 1:
...
+        abortSignal: signal,
+      })
+      .observe((emitter) => {
+        emitter.on("newToken", async (token) => {
+          params._meta?.progressToken &&
+            (await server.server.sendAgentRunProgress({
+              progressToken: params._meta.progressToken,
+              delta: {
+                logs: [
+                  { level: "info", message: JSON.stringify(token, null, 2) },
```
jezekra1, line 76: We don't have concatenation of tokens in log implemented anywhere or is it implemented on UI @kapetr?

Do we want to stream this in all details? Wouldn't it be sufficient to have something like "Generating podcast draft" (only informative messages about intermediate steps)

agents/official/beeai-framework/src/chat/chat.ts
```diff
@@ -66,17 +66,34 @@ const run =
...
+                  delta: {
+                    logs: [
+                      { level: "info", message: `${lastKey}: ${lastValue}` },
+                    ],
+                  } as MessageOutput,
+                }));
+              lastKey = update.key;
+              lastValue = "";
+            }
+            lastValue += update.value;
```
pilartomas: So the value is is sent only on the next iteration? What if it never comes?

agents/official/beeai-framework/src/chat/chat.ts
```diff
@@ -66,17 +66,34 @@ const run =
...
+                  messages: [{ role: "assistant", content: update.value }],
+                } as MessageOutput,
+              });
+            } else if (lastKey !== update.key) {
+              lastKey &&
+                (await server.server.sendAgentRunProgress({
+                  progressToken: _meta.progressToken,
+                  delta: {
+                    logs: [
+                      { level: "info", message: `${lastKey}: ${lastValue}` },
```
pilartomas: If UI is supposed to parse this, JSON would be safer.

agents/official/beeai-framework/src/podcast-creator/podcast-creator.ts
kapetr: @jezekra1 No, it's not, we'd have to add that.

agents/official/beeai-framework/src/chat/chat.ts
xjacka: The last message should be the result.

agents/official/beeai-framework/src/chat/chat.ts
xjacka: I agree, but the message must be a string.

agents/official/beeai-framework/src/podcast-creator/podcast-creator.ts
xjacka: I think that we should send all the details.

agents/official/beeai-framework/src/chat/chat.ts
pilartomas: JSON is a string

pilartomas (approved):

Issue #236, opened 2025-03-03
UI - hands-off task


tomkis:
# Hands Off Task agent UI

The actual flow of hands off task is pretty straightforward and can be defined as

1. Input task description
2. Awaiting execution of the task while seeing potential progress
3. Visualizing result of the task

## Input Task Description

The first state of the view is just a prompt text area where user can provide a prompt. 

Agent should also provide title to ask user for a prompt. For instance; Research Agent would ask for something like:

> "What is your research task?"

This could be exposed from metadata of the agent.

Once prompt is defined, user can click Run to proceed to next step.

## Awaiting Execution

User should see notifications coming from the Agent. This will be in a form of textual steps as they come in.

There should also be visible current runtime of the agent.

At this point, the agent is not cancelable and all the state happens transiently in the UI.

## Visualizing result of the task

Once the task is done, the agent will then automatically proceed to the next view which is two column layout to show input and progress on the left and output on the right

### Left Panel

User should see the inputed prompt. 

User should see task execution time in clear readable format - eg. 22 seconds or 2 min 32s.

User should see (with extra action to expand) list of all events that resulted in generating of the document.

### Right Panel

There's a paper component showing the detail of generated output in form of document.
 
Two actions can be done:

- Copy the content to clipboard
- Download the result as a file

tomkis:
@anafucs please review and attach Figma designs.

anafucs:
Designs here: https://www.figma.com/design/uJkcE3aJRPVk9HBOweNnH9/Bee-Dev-Ready-2.0?node-id=3002-7102
In the future, would be nice to be able to format the traces. We have designs for highlighting queries, sources, tools, as needed, depending on the agent.

PR #294, opened 2025-03-06
feat(ui): add hands-off ui

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai/issues/236


apps/beeai-ui/src/modules/run/chat/ChatInput.tsx
```diff
@@ -0,0 +1,113 @@
...
+    register,
+    watch,
+    handleSubmit,
+    formState: { isSubmitting },
+  } = form;
+
+  const resetForm = useCallback(() => {
+    {
+      formRef.current?.resetForm();
+    }
```
kapetr: redundant braces?

apps/beeai-ui/src/modules/run/components/AgentRunLogItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+    return () => observer.disconnect();
+  }, [children]);
+
+  return (
+    <div className={clsx(classes.root, { [classes.isExpanded]: isExpanded })}>
+      <AnimateHeightChange className={classes.holder}>
+        <p ref={contentRef} className={clsx(classes.content, { [classes.clamped]: !isExpanded })}>
+          {children}
+        </p>
+      </AnimateHeightChange>
```
kapetr, line 55: I might be missing something, but couldn't this be done with css only?
I liked the solution with grid: https://keithjgrant.com/posts/2023/04/transitioning-to-height-auto/

apps/beeai-ui/src/modules/run/components/AgentRunLogs.tsx
```diff
@@ -0,0 +1,59 @@
...
+import classes from './AgentRunLogs.module.scss';
+
+interface Props {
+  logs: TextNotificationLogs;
+  toggleable?: boolean;
+}
+
+export function AgentRunLogs({ logs, toggleable }: Props) {
+  const [isExpanded, setIsExpanded] = useState(false);
+  const filteredLogs = logs.filter(isNotNull);
```
kapetr: `null` logs are already filtered out upon recieval in `HandsOffProvider` - I'd change the type in the context, so this is not necessary.

apps/beeai-ui/src/modules/run/components/AgentRunLogItem.tsx
PetrBulanek: I originally thought so and implemented it that way. However, it didn't work. There's a problem with the animation working when it's animated from 0fr to 1fr, but I need the first line of the text to be visible, and when I tried it for example from 1.25rem to 1fr, it didn't animate and just jumped weirdly.

PR #310, opened 2025-03-07
refactor(agents): unify env

JanPokorny:
Unify environment variables in agents. Now, all* of the agents are unified to use env vars `LLM_API_KEY`, `LLM_API_BASE` and `LLM_MODEL`, assuming an OpenAI-compatible endpoint.

This could eventually help us implement our own OpenAI-compatible endpoint proxy, but for now it at least unifies most of the agents under a single interface.

Yeah, this potentially limits compatibility with extra features of some providers. But it seems like a good tradeoff at this point.

~*`gpt-researcher` uses an embedding endpoint which is not available on Groq, so I left it at the current state (Ollama only) for the time being.~

### DIFF
--- a/agents/community/aider-agent/beeai-provider.yaml
+++ b/agents/community/aider-agent/beeai-provider.yaml
@@ -7,10 +7,15 @@ mcpTransport: sse
 mcpEndpoint: /sse
 command: ["aider-agent"]
 env:
-# Required Aider configuration:
-- name: AIDER_MODEL
+- name: LLM_MODEL
   required: true
-  description: "Model specifier in the LiteLLM format, e.g. 'groq/deepseek-r1-distill-llama-70b'. Do not forget to set the corresponding API key -- see docs: https://docs.litellm.ai/docs/providers."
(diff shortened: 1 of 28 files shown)


jezekra1:
This collides with our current env management, are we sure we want to go this way now?

`beeai env add API_KEY=...` will set the api key globally for all providers after this PR because they'll use the same env name.

There might be a use-case where I want `openai` for one agent but grok for another one for example, the current approach with `OPENAI_API_KEY` and `GROK_API_KEY` would allow it

JanPokorny:
@jezekra1 My take is:
- it was partially like that already (eg. `API_BASE` used by several agents)
- for the demo we need lowest friction possible => we can't really make it zero-config but this is the fewest config manageable
- going forward we may want to provide an in-platform OpenAI-compatible API proxy, which could allow LLM routing on the platform level -- or at least allow for per-agent (/ per-provider?) config
	- or at the very least allow for something like `beeai env add --only=aider-agent API_KEY=...`

tomkis:
I'd personally prefer to prefix the envs with `BEEAI_`

PR #319, opened 2025-03-10
feat(ui): add providers list to the Settings page

PetrBulanek:


apps/beeai-ui/src/modules/agents/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isNotNull } from '#utils/helpers.ts';
+import { uniq } from 'lodash';
```
penge:
```suggestion
import uniq from 'lodash/uniq';
```

apps/beeai-ui/src/modules/providers/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Agent } from '#modules/agents/api/types.ts';
+import { groupBy } from 'lodash';
```
penge:
```suggestion
import groupBy from 'lodash/groupBy';
```

penge (changes requested):

apps/beeai-ui/src/hooks/useTableSearch.ts
```diff
@@ -0,0 +1,45 @@
...
+    setSearch(event.target.value);
+  }, []);
+
+  const items = useMemo(() => {
+    if (!searchQuery) {
+      return entries;
+    }
+
+    return entries.filter((item) => fields.some((field) => String(item[field]).toLowerCase().includes(searchQuery)));
+  }, [searchQuery, entries, fields]);
```
penge: > In React, the `useMemo` hook will not track a variable that is just a regular local variable (without `useMemo`, `useState`, or `useReducer`) because React's hooks only track **state and props** to determine when to re-run.

Let's change dependency to `search` and calculate `searchQuery` within the `useMemo`.

penge (changes requested):

apps/beeai-ui/src/modules/agents/utils.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Agent } from './api/types';
+
+export const getAgentsLanguages = (agents: Agent[] | undefined) =>
```
penge, line 21: Would be nicer to always expect an array. Or at least have a guard clause that returns `[]` on undefined input.

apps/beeai-ui/src/hooks/useTableSearch.ts
PetrBulanek: I was originally returning the searchQuery from the hook, but noticed I don't need to. I'll change it. :-)

apps/beeai-ui/src/modules/agents/utils.ts
PetrBulanek: The `uniq` function from lodash handles null and undefined values and returns empty array in that case. :-)

kapetr (approved):

penge (approved):

PR #320, opened 2025-03-10
chore(telemetry): add semi-stable instance id

pilartomas:

### DIFF
--- a/apps/beeai-server/src/beeai_server/utils/id.py
+++ b/apps/beeai-server/src/beeai_server/utils/id.py
@@ -1,0 +1,35 @@
+# Copyright 2025 IBM Corp.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


apps/beeai-server/src/beeai_server/utils/id.py
```diff
@@ -0,0 +1,35 @@
...
+    except AttributeError:
+        # Fallback for Windows
+        system_info = [
+            platform.system(),  # OS name
+            platform.node(),  # Hostname
+            platform.machine(),  # Machine type
+        ]
+        combined_info = "".join(system_info).encode("utf-8")
+
+    return hashlib.sha256(combined_info).hexdigest()[:16]
```
pilartomas: I'd prefer KDF, is there some in the standard lib?

jezekra1: there is something in `hashlib` probably, but sha256 seems good enough for this

apps/beeai-server/src/beeai_server/utils/id.py
```diff
@@ -0,0 +1,35 @@
...
+import os
+import platform
+import hashlib
+
+
+def generate_stable_id() -> str:
+    try:
+        # Use os.uname() for Unix-like systems
+        uname_info = os.uname()
+        # Only use stable attributes: sysname, nodename, machine to make it more stable
```
jezekra1, line 25: Be aware that this might not be unique if the user uses some generic nodename, for example (`Darwin`, `MacBook`, `arm64`)

jezekra1 (approved):

apps/beeai-server/src/beeai_server/utils/id.py
pilartomas: 👍 that's acceptable for our purposes

PR #330, opened 2025-03-11
chore: update README.md to provide info about LLM config

tomkis:


tomkis:
This is obsolete with #354

Issue #312, opened 2025-03-07
BeeAI landing page updates


anafucs:
Our current landing page targets the framework. We should update to focus on the BeeAI platform.

Updates:

- Header: add link to Docs https://docs.beeai.dev/
- Header: include icons with links for:
  - GitHub: https://github.com/i-am-bee/beeai
  - Discord: Discord: https://discord.gg/NradeA6ZNF
  - Youtube: https://www.youtube.com/@BeeAIAgents
  - Bluesky: https://bsky.app/profile/beeaiagents.bsky.social
- Value prop copy: "The open-source platform to discover, run, and compose AI agents from any framework"
- Replace CTAs
  - `brew install beeai`
  - "Star on GitHub" https://github.com/i-am-bee/beeai
- Add single video exactly as in the old webpage (design attached). Current video attached (to be updated soon!)

![Image](https://github.com/user-attachments/assets/8c7a6a77-4c0d-4dd4-bde2-511c43cb8738)

https://github.com/user-attachments/assets/a2c9b398-a51c-4d36-800a-af0ec2410da5

anafucs:
video stills. @penge let me know if this works?
![Image](https://github.com/user-attachments/assets/a1871f02-04d0-4770-ac6b-58aa26fbbc2a)
<img width="658" alt="Image" src="https://github.com/user-attachments/assets/e20dec8b-1985-4c72-9452-370f5aef74ad" />

penge:
@anafucs

I think the second image will work. Thank you.

kapetr:
@anafucs I'm just wondering, should the updates to social icons from #338 apply also to the landing page to keep things consistent?
cc @penge

PetrBulanek:
Is there anything else that needs to be done? Or can we close this? @penge @anafucs

anafucs:
can we remove the dark mode toggle? :) 
it should be set based on system preferences but we don't need an extra icon

kapetr:
@anafucs It's live on beeai.dev :)

PR #334, opened 2025-03-11
chore(ui): update GitHub button

penge:
Refs: https://github.com/i-am-bee/beeai/issues/312

### Preview

github-button


apps/beeai-ui/src/components/CommunityNav/CommunityNav.tsx
```diff
@@ -25,7 +25,7 @@ export function CommunityNav() {
       <ul className={classes.list}>
         {NAV_ITEMS.map(({ label, href, Icon }) => (
           <li key={label} className={classes.item}>
-            <a href={href} target="_blank" aria-label={label} className={classes.link}>
+            <a href={href} target="_blank" rel="noreferrer" aria-label={label} className={classes.link}>
```
PetrBulanek, line 28: We had a discussion about this in the past and it shouldn't be needed anymore, because browsers automatically handle links with target="_blank" this way.

apps/beeai-ui/src/modules/agents/components/AgentMetadata.tsx
```diff
@@ -59,7 +59,7 @@ AgentMetadata.Skeleton = function AgentMetadataSkeleton({ className }: SkeletonP
 
 function GithubLink({ githubUrl }: { githubUrl: string }) {
   return (
-    <a target="_blank" href={githubUrl} className={classes.githubLink} aria-label="Open on Github">
+    <a target="_blank" rel="noreferrer" href={githubUrl} className={classes.githubLink} aria-label="Open on Github">
```
PetrBulanek, line 62: Same as above.

apps/beeai-ui/src/modules/home/components/GitHubButton.tsx
```diff
@@ -14,16 +14,16 @@
...
-  if (!response.ok) {
-    throw new Error('Failed to fetch repository data.');
-  }
-
-  const data = await response.json();
-
-  return GitHubRepoSchema.parse(data);
+export function GitHubButton() {
+  return (
+    <Button as="a" href={GITHUB_REPO_LINK} target="_blank" rel="noreferrer" size="md" className={classes.root}>
```
PetrBulanek, line 24: Same as above.

apps/beeai-ui/package.json
PetrBulanek, line 1: You can remove the `millify` package as well.

apps/beeai-ui/src/contexts/QueryProvider/QueryProvider.tsx
```diff
@@ -15,26 +15,10 @@
...
-});
-
-const persistOptions: PersistQueryClientProviderProps['persistOptions'] = {
-  persister: localStoragePersister,
-  dehydrateOptions: {
-    shouldDehydrateQuery(query) {
-      return query.queryKey[0] === gitHubRepoKeys.all()[0];
-    },
-  },
-};
```
PetrBulanek, line 36: At least we enjoyed the discussion about query keys. :-D

apps/beeai-ui/src/components/CommunityNav/CommunityNav.tsx
penge: Maybe you mean **"noopener"**.

> Browsers do not automatically add **rel="noreferrer"** when **target="_blank"** is used. However, they do add **rel="noopener"** by default in some cases for security reasons.

> Modern browsers (Chrome ≥ 88, Edge ≥ 88, Firefox ≥ 79 and Safari ≥ 12.2) automatically imply **rel="noopener"**. Therefore this rule is no longer needed, if legacy browsers are not supported.

https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

https://biomejs.dev/linter/rules/no-blank-target/

apps/beeai-ui/package.json
penge: Package deleted.

apps/beeai-ui/src/contexts/QueryProvider/QueryProvider.tsx
penge: Case closed 😄

apps/beeai-ui/src/components/CommunityNav/CommunityNav.tsx
PetrBulanek: Hmm, I guess you are right. The "referrer" is not a security issue, but more of a privacy one. But I think we might use it with "noreferrer" then.

PetrBulanek: We might even enforce the eslint rule to make it the same everywhere.

PetrBulanek (approved):

apps/beeai-ui/src/components/CommunityNav/CommunityNav.tsx
penge: Yes. I mean. It is rule in the `recommended` config so I would keep using it until the rule stops being relevant.

We can update ESLint. Or just install Biome and you will see those errors instantly. Or Both!! 😄

penge: Btw. `"noreferrer"` implies `"noopener"` so it's 2 in 1.

PR #343, opened 2025-03-11
feat(ui): prompted sequential workflow 

### LINKED ISSUE #289: UI - Prompted sequential workflow

kapetr:
Closes #289 

Implements UI for the prompted sequential workflow agent with an interim design based on already implemented components and aiming for the final design in Figma:
https://www.figma.com/design/uJkcE3aJRPVk9HBOweNnH9/Bee-Dev-Ready-2.0?node-id=3831-50140&t=GCwQnDjE22vYoqcO-1

Known issues:
- auto scroll on new logs recieved gets stuck sometimes
- some agents don’t send notifications, so there’s no data indicating that execution has started, making it impossible for the UI to reflect it.


apps/beeai-ui/src/modules/compose/sequential/utils.ts
```diff
@@ -16,26 +16,16 @@
...
-const messageOutputJsonSchema = zodToJsonSchema(messageOutputSchema);
-const textInputJsonSchema = zodToJsonSchema(textInputSchema);
-const textOutputJsonSchema = zodToJsonSchema(textOutputSchema);
+import { SEQUENTIAL_COMPOSE_AGENT_NAME, textInputJsonSchema, textOutputJsonSchema } from './constants';
 
 export function isValidForSequentialWorkflow(agent: Agent) {
   return (
-    (validateJsonSchema(agent.inputSchema as JSONSchema, messageInputJsonSchema as JSONSchema) &&
-      validateJsonSchema(agent.outputSchema as JSONSchema, messageOutputJsonSchema as JSONSchema)) ||
+    agent.ui?.type === 'hands-off' ||
```
PetrBulanek: There is a UiType enum in `apps/beeai-ui/src/modules/run/types.ts` that should be used instead.

apps/beeai-ui/src/modules/compose/ComposeLanding.module.scss
```diff
@@ -14,6 +14,69 @@
...
+}
+
+.actionBar {
+  margin-block-start: $spacing-07;
+  display: flex;
+  justify-content: flex-end;
+
+  button {
+    inline-size: rem(202px);
+  }
```
PetrBulanek: The button is actually anchor link, so this is not applied. I think using class would be more safe. :-)

apps/beeai-ui/src/modules/compose/components/AddAgentButton.tsx
PetrBulanek, line 1: I think we should order the available agents alphabetically and the official agents are missing the BeeAI badge.

apps/beeai-ui/src/modules/compose/components/ComposeResult.module.scss
```diff
@@ -0,0 +1,49 @@
...
+.root {
+  display: flex;
+  flex-direction: column;
+  background-color: $background;
+  padding-block-start: 0;
+  inline-size: 0;
+  @include scrollbar();
+  overflow-y: auto;
+  overflow-x: hidden;
+  background-color: $layer-02;
```
PetrBulanek: The background-color is defined twice.

apps/beeai-ui/src/modules/compose/components/ComposeStepListItem.module.scss
PetrBulanek, line 1: This seems on the edge of file length to me. I might consider breaking it down into smaller components. However, there's no need to address that right now.

apps/beeai-ui/src/modules/compose/contexts/ComposeProvider.tsx
```diff
@@ -42,60 +73,52 @@ export function ComposeProvider({ children }: PropsWithChildren) {
...
-          if (delta.agent_name !== agent.data.name) {
-            console.error(`Agent name and index mismatch: ${delta.agent_name} is supposed to be at index '${idx}'`);
-            return agent;
-          }
+      step.isPending = true;
+      step.stats = {
+        startTime: step.stats?.startTime ?? Date.now(),
+      };
+      step.result = `${step.result ?? ''}${delta.text}`;
+      step.logs = [...(step.logs ?? []), ...delta.logs.filter(isNotNull).map((item) => item.message)];
```
PetrBulanek: Wouldn't it be safer not to mutate the object and create a new one instead?

```
const updatedStep = {
  ...step,
  isPending: true,
  ...
}


updateStep(delta.agent_idx, updatedStep);
```

apps/beeai-ui/src/modules/compose/contexts/ComposeProvider.tsx
```diff
@@ -42,60 +73,52 @@ export function ComposeProvider({ children }: PropsWithChildren) {
...
+            step.stats = { ...step.stats, endTime: Date.now() };
+            updateStep(stepsBeforeIndex, step);
+          }
+        });
+      }
+    },
+    [getValues, updateStep],
+  );
+
+  const { runAgent } = useRunAgent<SequentialWorkflowInput, ComposeNotificationSchema>({
```
PetrBulanek, line 122: Now there's no way for the user to know about the error when the toast message is off, right? Shouldn't we keep it?

apps/beeai-ui/src/modules/compose/contexts/ComposeProvider.tsx
kapetr: I don't know why I was under the impression that getValues returns a copy of objects.  But it's not :) Thanks!

apps/beeai-ui/src/modules/compose/contexts/ComposeProvider.tsx
kapetr: Yes, I know about the missing handling. Some agents don't event throw the error, they send error messages in response text. I'd handle this later if you're okay with that :)

PetrBulanek: Ok. 👍

PR #346, opened 2025-03-11
Fix ACP cancellation

jezekra1:


packages/acp-python-sdk/src/acp/shared/session.py
```diff
@@ -116,10 +116,10 @@ async def cancel(self) -> None:
         """Cancel this request and mark it as completed."""
         if not self._entered:
             raise RuntimeError("RequestResponder must be used as a context manager")
-        if not self._cancel_scope:
+        if not self.task_group:
             raise RuntimeError("No active cancel scope")
 
-        self._cancel_scope.cancel()
```
jezekra1, line 122: This cancelled the parent task which caused cancelling of the entire `ServerSession`

Issue #313, opened 2025-03-07
UI - Design QA


anafucs:
Visual and UX QA for the current app:

- Handoff task agent
  - align input field vertically as screenshot (same distance from field to top / bottom)
  - scroll up logs as it gets to the bottom of the page so the user can follow the latest step without needing to scroll it
  - tooltips with 8px rounded corners

![Image](https://github.com/user-attachments/assets/e9d93bbb-c8d2-4ea8-a293-150571a498fb)

![Image](https://github.com/user-attachments/assets/0caca259-3fdb-46b9-8d49-e90ced2ee993)

tomkis:
Finding:

When agent fails, it's not handled nor communicated gracefuly to the user. 

For instance: Try gpt-researcher without running ollama, it just terminates immediately.

anafucs:
In settings, can we truncate the variable value and keep it 1 line? Remove button should be visible
<img width="774" alt="Image" src="https://github.com/user-attachments/assets/ad8bd195-d1d9-48a8-a566-b724aaa8f7cc" />

rachel-c-h:
Hi Team! Quick update - the settings overflow now has a divider to separate the social links and documentation.   

<img width="303" alt="Image" src="https://github.com/user-attachments/assets/fbe336ab-ddcb-431a-a960-8b91b6883398" />
<img width="197" alt="Image" src="https://github.com/user-attachments/assets/af9af2cb-4012-46b4-9e56-b1ef2d73ad55" />

anafucs:
All handoff task agents are prompting "What is your research task". We should change to:
"What is your task?"
![Image](https://github.com/user-attachments/assets/3c9a009f-e607-4b73-9836-356cad73b768)

anafucs:
Can we add this favicon to the hosted app?
![Image](https://github.com/user-attachments/assets/e1ab3911-8f05-4a61-b56c-332c8044aed2)

PetrBulanek:
> Can we add this favicon to the hosted app? ![Image](https://github.com/user-attachments/assets/e1ab3911-8f05-4a61-b56c-332c8044aed2)

This should be already done, right @penge? :-)

PetrBulanek:
> All handoff task agents are prompting "What is your research task". We should change to: "What is your task?" ![Image](https://github.com/user-attachments/assets/3c9a009f-e607-4b73-9836-356cad73b768)

[...]

PR #347, opened 2025-03-11
feat(ui): hands-off qa

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai/issues/313

### DIFF
--- a/apps/beeai-ui/src/hooks/useAutoScroll.ts
+++ b/apps/beeai-ui/src/hooks/useAutoScroll.ts
@@ -1,0 +1,64 @@
+/**
+ * Copyright 2025 IBM Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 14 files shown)


apps/beeai-ui/src/hooks/useAutoScroll.ts
```diff
@@ -0,0 +1,64 @@
...
+      } else if (scrolledToBottom) {
+        setShouldAutoScroll(true);
+      }
+    },
+    [scrollableContainer],
+  );
+
+  useEffect(() => {
+    if (scrollableContainer) {
+      scrollableContainer.addEventListener('wheel', handleWheel);
```
kapetr, line 45: I'd consider adding `touchmove` event as well.

apps/beeai-ui/src/hooks/useAutoScroll.ts
kapetr, line 1: It works great 👍  Just the usage seems not very intuitive to me - having to add the ref to the anchor in the end of scrollable container. I'd like it more, if it was a React component that just wraps the scrollable content and the anchor element would be part of it. You wouldn't event need the `useScrollableContainer` then, because you'd control the scrollable element.

But no need to change that in this PR - just my thoughts on the implementation approach! :)

kapetr (approved):

PR #350, opened 2025-03-11
feat(ui): add video

penge:
Adds **BeeAI video**. No preview image until https://github.com/i-am-bee/beeai/issues/312#issuecomment-2715632864 resolved.

Additional changes:
- Moved social icons from footer to header.
- Fixed wrapping of _snippet_ and _github_ on small screen. (they didn't wrap and there was no vertical spacing)

### Preview

https://github.com/user-attachments/assets/047aedd8-8f01-4b21-bbec-9d9ce2eef369


penge:
I'd say it looks great without preview image.
_Safari_ (and second _Firefox_) even creates something which very much looks so as a preview image with Play button.
It is also faster to use. (can click anywhere, not just on the Play button; doesn't load additional image; play controls already visible)

<br>

### Safari

safari

<br>

### Firefox

firefox

<br>

### Chrome

chrome

apps/beeai-web/src/app/page.module.scss
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+// Styles should match the center and edges of VideoBeeAIPoster.webp
+.video {
+  --container-background: linear-gradient(to bottom, #f3f3f3, #f0f0f0);
+  --play-button-background: white;
+  --play-button-fill: black;
+
+  @media (prefers-color-scheme: dark) {
```
PetrBulanek: This won't actually work if the user has light mode in system and changes app's mode to dark. We use classes on html for theming. Take a look `apps/beeai-ui/src/styles/style.scss`. Ideally you should add these variables here `apps/beeai-ui/src/styles/_theme.scss` (similarly like for example `modal-background` variable).

apps/beeai-web/src/layouts/AppLayout.tsx
```diff
@@ -14,25 +14,25 @@
...
       </AppHeader>
 
       <main className={classes.main} data-route-transition>
         {children}
       </main>
 
-      <AppFooter className={classes.footer} />
+      <footer className={classes.footer}>
+        <CommunityNav />
+      </footer>
```
PetrBulanek: I am not sure, if the footer should stretch to side infinitely. I think there should be some max-width limitation such as in header. Why using custom footer element and not reuse <AppFooter /> where we already have the container with max-width? Also the ThemeToggle disappeared here. :-)

PetrBulanek (changes requested):

apps/beeai-web/src/layouts/AppLayout.tsx
penge: Footer reverted. I was wondering where this Toggle is hidden :)

Fix: 81dbcf50d60885ac217bce36751f59adf6701bf8

apps/beeai-web/src/app/page.module.scss
penge: Updated to use SASS variables. Defined `hero-background` variable.

See: ef1b1e2b56c940b887ce231aa4e95ffb6ec4db93

PetrBulanek (approved):

Issue #241, opened 2025-03-03
Add Traces Button


tomkis:
Add Traces button to the UI.

- We only want to show the button when Pheonix is up and running
- Clicking on the button would open Pheonix in new tab

GALLLASMILAN:
We don't support remote experience. Only for local using

PR #367, opened 2025-03-13
feat(phoenix): add traces button (#241)

GALLLASMILAN:
I added a new "Traces" tab to the menu, that is enabled only when the Phoenix server is running. Otherwise, the disabled item with the tooltip with the link to the Observability documentation part is visualized. 

It works only locally. (the Phoenix server target is set to `http://localhost:6006`) the same way as for mcp and API proxy. See `vite.config.ts` for more details. 

The state is checked every 60 seconds.

### DIFF
--- a/apps/beeai-ui/src/components/MainNav/MainNav.tsx
+++ b/apps/beeai-ui/src/components/MainNav/MainNav.tsx
@@ -38,8 +38,8 @@ export function MainNav() {
         }).map(({ label, to, section, target, rel, isDisabled }, idx) => (
           <li key={idx} className={clsx({ [classes.active]: section && isSectionActive(section) })}>
             {isDisabled ? (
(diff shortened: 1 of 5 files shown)


apps/beeai-ui/src/components/MainNav/MainNav.tsx
```diff
@@ -19,38 +19,76 @@ import classes from './MainNav.module.scss';
...
+    {
+      label: 'Compose playground',
+      to: routes.compose(),
+      section: sections.compose,
+    },
+    {
+      label: 'Traces',
+      to: `${phoenixServerTarget}/projects`,
+      target: '_blank',
+      rel: 'noopener noreferrer', // Security best practice
```
kapetr, line 90: `noreferrer` is sufficient, it behaves as if `noopener` was also used.

kapetr (approved):

Issue #385, opened 2025-03-14
beeai.dev - 404 for Bee.svg


Tomas2D:

tomkis:
Seems to be fixed on latest beeai.dev.

PR #386, opened 2025-03-14
fix: rename Bee.svg to bee.svg

Tomas2D:
Ref: #385


apps/beeai-ui/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <link rel="icon" type="image/svg+xml" href="/src/svgs/Bee.svg" />
+    <link rel="icon" type="image/svg+xml" href="/bee.svg" />
```
penge, line 6: 1. The previous `Bee.svg` is no longer referenced and can **_be removed_**.
2. To use a _root-relative_ asset like `bee.svg`, it must **_be placed_** in the `public` directory, such as:
```
public/bee.svg
```

kapetr (approved):

apps/beeai-ui/index.html
kapetr: It seems to me there's nothing to fix—please take another look at the changes. `svgs/bee.svg` **is** referenced, it can't be removed and `bee.svg` **is** also placed in public folder.

penge: But change is also made in `beeai-ui` and it has its own public directory. You can try to change favicon to something else to invalidate it, then back to `/bee.svg` and it won't load.

@kapetr

kapetr: You're right, I added `bee.svg` to `beeai-ui/public` as well.

penge (approved):

PR #391, opened 2025-03-14
fix(beeai-server): brew node path

JanPokorny:


apps/beeai-server/src/beeai_server/domain/model.py
```diff
@@ -194,12 +198,30 @@ async def mcp_client(
         if not with_dummy_env:
             self.check_env(env)
 
+        with contextlib.suppress(Exception):
+            env["PATH"] = (
+                str(
+                    sorted(
+                        list(
+                            Path(subprocess.check_output(["brew", "--cellar", "node"], text=True).strip()).glob("*/bin")
```
jezekra1: can we make this async?
both `Path` glob (from anyio) and `subprocess run`. Or at least run using `anyio.to_thread`

jezekra1 (changes requested):

jezekra1 (approved):

PR #409, opened 2025-03-17
feat(web): autoplay video, footer style update

kapetr:
- remove the dark mode button
- autoplay video
- icons under the content if not enough space

### DIFF
--- a/apps/beeai-ui/src/components/layouts/AppFooter.tsx
+++ b/apps/beeai-ui/src/components/layouts/AppFooter.tsx
@@ -21,15 +21,15 @@ import { Container } from './Container';
 
 interface Props {
   className?: string;
-  hideThemeToggle?: boolean;
+  showThemeToggle?: boolean;
 }
 
-export function AppFooter({ className, hideThemeToggle }: Props) {
+export function AppFooter({ className, showThemeToggle = true }: Props) {
   return (
     <footer className={className}>
       <Container size="max">
         <div className={classes.holder}>
-          {!hideThemeToggle && <ThemeToggle />}
+          {showThemeToggle && <ThemeToggle />}
 
           <CommunityNav className={classes.communityNav} />
         </div>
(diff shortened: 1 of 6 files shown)


apps/beeai-ui/src/components/layouts/AppFooter.tsx
```diff
@@ -21,14 +21,15 @@ import { Container } from './Container';
 
 interface Props {
   className?: string;
+  hideThemeToggle?: boolean;
```
PetrBulanek: Personally, I'm more inclined to use positive props, i.e. the “showThemeToggle” with default true value. But it's just a matter of taste.. :-)

apps/beeai-ui/src/modules/home/components/VideoBeeAI.tsx
```diff
@@ -43,6 +43,31 @@ export function VideoBeeAI({ src, type, poster }: VideoBeeAIProps) {
     });
   };
 
+  useEffect(() => {
+    const video = videoRef.current;
+    if (!video || playedOnce) return;
+
+    const observer = new IntersectionObserver(
```
PetrBulanek: Not sure, if it actually fits here, but usehooks package has useIntersectionObserver hook https://usehooks-ts.com/react-hook/use-intersection-observer, just fyi. :-)

PetrBulanek (approved):

apps/beeai-ui/src/components/layouts/AppFooter.tsx
kapetr: Same here! I think I went with the negative phrasing because I don’t really approve of hiding it.  😄

apps/beeai-ui/src/modules/home/components/VideoBeeAI.tsx
kapetr: Of course, much better :)

PR #417, opened 2025-03-18
chore(web): enable ISR

kapetr:


apps/beeai-ui/src/modules/agents/list/AgentsView.tsx
PetrBulanek, line 1: Why is this needed? I noticed you added the ::marker display: none rule in CSS, so just want to now, where was the issue. I just find it a bit discrepant when we have li only inside the skeleton and not in the card itself. :-)

apps/beeai-web/src/acp/client.ts
PetrBulanek, line 1: It'd be good to also update the comment with reasoning.

apps/beeai-web/src/app/agents/AgentsView.tsx
PetrBulanek, line 1: Regarding the changes with li tag, I think there is actually no ul tag wrapping the skeleton items during loading and no li tags wrapping the cards once loaded. It might be worth chaning this so that each component renders the skeletons inside itself.

apps/beeai-web/src/app/agents/[name]/page.tsx
Comment (PetrBulanek) [inline apps/beeai-web/src/app/agents/[name]/page.tsx:1]: We should add a comment as to why this is important so that someone doesn't accidentally remove it in the future.

apps/beeai-web/src/app/agents/page.tsx
PetrBulanek, line 1: I'd rather add a comment here as well.

PetrBulanek (changes requested):

apps/beeai-ui/src/modules/agents/list/AgentsView.tsx
kapetr: I didn't pay enough attention to this, I was focused on the issues with isr. This fix just addresses the visible markers in skeletons on the web. I’ll take a closer look now.

PetrBulanek: I think the markers appeared due to having li tags rendered without ul wrapper as described in this comment https://github.com/i-am-bee/beeai/pull/417#discussion_r2002696417

kapetr: yes, already fixed.

apps/beeai-ui/src/modules/agents/components/AgentCard.tsx
PetrBulanek, line 1: The div with classes.root should actually be here.

PetrBulanek (changes requested):

PetrBulanek (approved):

pilartomas (approved):

Issue #415, opened 2025-03-18
beeai.dev - add LF copyright


geneknit:
BeeAI is now a series of Linux Foundation Projects. 🥳 

We need to update our website with the following.

> Copyright © BeeAI a Series of LF Projects, LLC
> For web site terms of use, trademark policy and other project policies please see https://lfprojects.org/.

To do:
- [x] Design
- [ ] Implement

anafucs:
![Image](https://github.com/user-attachments/assets/95796b04-4231-448a-9996-b6385f9db1c2)
Figma here: https://www.figma.com/design/uJkcE3aJRPVk9HBOweNnH9/Bee-Dev-Ready-2.0?node-id=4654-30669

PetrBulanek:
@anafucs @geneknit Just checking - should I add the copyright to the local app as well?

PR #440, opened 2025-03-21
feat(web): add copyright to footer and simplify some shared components

PetrBulanek:
Refs: https://github.com/i-am-bee/beeai/issues/415

- adding copyright into web's footer
- refactoring some components shared between ui and web to duplicate less code
- creating SplitPanesView component shared for handsoff and compose ui


apps/beeai-ui/src/components/MainContentView/MainContentView.module.scss
```diff
@@ -16,15 +16,30 @@
...
+@include breakpoint-up(md) {
+  .root {
+    &.lg {
+      padding-block: $spacing-09;
+      row-gap: $spacing-09;
+    }
+
+    &:has(.footer) {
+      padding-block-end: 0;
+    }
```
penge, line 43: Looks like a duplicity (although I know it's here to override the style few lines above).

Could we maybe move the media querie(s) under the base classname `.root` to have it at one place? It could also help remove the duplicity.

apps/beeai-ui/src/components/MainContentView/MainContentView.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { useScrollbarWidth } from '#hooks/useScrollbarWidth.ts';
+import { useToTopButton } from '#hooks/useToTopButton.ts';
+import clsx from 'clsx';
+import type { CSSProperties, PropsWithChildren } from 'react';
+import { mergeRefs } from 'react-merge-refs';
+import { ToTopButton } from '../ToTopButton/ToTopButton';
+import classes from './MainContentView.module.scss';
+
+export interface MainContentViewProps extends PropsWithChildren {
+  spacing?: 'md' | 'lg' | false;
```
penge: Why do we have `false` as one of the options? It doesn't seem valid spacing or intuitive. Better alternative would be `'none'` if we want to be explicit about it.

apps/beeai-ui/src/components/MainContentView/MainContentView.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <div
+      ref={mergeRefs([toTopRef, scrollbarRef])}
+      className={clsx(classes.root, spacing && classes[spacing], className)}
+      style={{ '--scrollbar-width': `${scrollbarWidth}px` } as CSSProperties}
+    >
+      <div className={classes.body}>{children}</div>
+
+      {showButton && <ToTopButton onClick={handleToTopClick} />}
+
+      {showFooter && <AppFooter className={classes.footer} />}
```
penge, line 54: This looks to me like something that could be part of a Layout (or one of the layouts).

apps/beeai-ui/src/components/MainNav/MainNav.module.scss
```diff
@@ -40,3 +43,9 @@
     color: $background-inverse;
   }
 }
+
+@include breakpoint-up(md) {
+  .list {
+    column-gap: $spacing-05;
+  }
+}
```
penge, line 51: I think it's best to keep media queries under the base classname. (because it applies to the same classname)

apps/beeai-ui/src/components/MainNav/MainNav.tsx
```diff
@@ -14,82 +14,69 @@
...
-    {
-      label: 'Traces',
-      to: `${phoenixServerTarget}/projects`,
-      target: '_blank',
-      rel: 'noreferrer',
-      isDisabled: Boolean(!showPhoenixTab || !phoenixServerTarget),
-    },
-  ];
+  const LinkComponent = isExternal ? 'a' : linkComponent;
+  const linkProps = isExternal ? { target: '_blank', rel: 'noreferrer' } : null;
```
penge, line 62: This doesn't seem right to me. Because although `linkComponent` is provided as a prop, still, a different component might be used (`a`).

I think `isExternal` prop can be avoided as well (it can lead to bugs).

What could be ideal? If it just works, and the correct link is used based on the href. It very much reminds me:
https://github.ibm.com/granite/docs/blob/f91d218c7a98b27a7a742c7a321a90a013c9a376/packages/core/src/links/GatsbyLink.jsx
(if you don't have access, let me know)

apps/beeai-ui/src/components/MainNav/MainNav.tsx
```diff
@@ -14,82 +14,69 @@
...
+
+  return (
+    <li className={clsx({ [classes.active]: isActive })}>
+      {isDisabled && disabledTooltip ? (
```
penge: How about replacing both props with only one prop? Such as:
```
disabled?: {
  tooltip: string
}
```

It would avoid having bugs like when `isDisabled` is true, but `disabledTooltip` is unset.
(I think if multiple props should be set to get 1 functionality, it should be 1 prop.)

apps/beeai-ui/src/components/SplitPanesView/SplitPanesView.tsx
```diff
@@ -0,0 +1,82 @@
...
+import type { MainContentViewProps } from '#components/MainContentView/MainContentView.tsx';
+import { useScrollbarWidth } from '#hooks/useScrollbarWidth.ts';
+import { moderate02 } from '@carbon/motion';
+import { AnimatePresence, motion } from 'framer-motion';
+import type { CSSProperties, PropsWithChildren, ReactNode } from 'react';
+import classes from './SplitPanesView.module.scss';
+
+interface Props {
+  leftPane: ReactNode;
+  rightPane: ReactNode;
```
penge, line 29: Props **grouping** could help understand the overall structure (left, right), and remove duplicity. How about?

```
panes: {
  left:  ReactNode;
  right: ReactNode;
}
```

apps/beeai-ui/src/components/SplitPanesView/SplitPanesView.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { ref: leftPaneRef, scrollbarWidth } = useScrollbarWidth();
+
+  return (
+    <AnimatePresence mode="wait">
+      {isSplit ? (
+        <Wrapper key="split-view" className={classes.splitView} immediateExit>
+          <div
+            className={classes.leftPane}
+            ref={leftPaneRef}
+            style={{ '--scrollbar-width': `${scrollbarWidth}px` } as CSSProperties}
```
penge: This seems common. How about:
```
...createScrollbarStyles({ width })
```

Which would be a helper and take care of it.

apps/beeai-ui/src/components/layouts/Navigation.tsx
```diff
@@ -0,0 +1,63 @@
...
+        label: 'Compose playground',
+        href: routes.compose(),
+        section: sections.compose,
+      },
+      {
+        label: 'Traces',
+        href: `${PHOENIX_SERVER_TARGET}/projects`,
+        isExternal: true,
+        isDisabled: Boolean(isPending || error || !showPhoenixTab || !PHOENIX_SERVER_TARGET),
+        disabledTooltip: <TracesTooltip />,
```
penge: Example use based on the other comment about merging props:

```tsx
disabled: Boolean(isPending || error || !showPhoenixTab || !PHOENIX_SERVER_TARGET)
  ? { tooltip: <TracesTooltip /> }
  : undefined,
```

If the object is set, it automatically means disabled.

apps/beeai-ui/src/hooks/useToTopButton.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export function useToTopButton({ enabled = true }: Props = {}) {
+  const ref = useRef<HTMLDivElement>(null);
+  const [showButton, setShowButton] = useState(false);
+
+  const handleScroll = useCallback((event: Event) => {
+    const { scrollTop } = event.target as HTMLDivElement;
+
+    setShowButton(scrollTop > SCROLLED_OFFSET);
```
penge, line 33: `SCROLLED_OFFSET` is offset, that's true, but in the grand scheme of things, it means something else. Why is it 48?
How about calling it `APP_HEADER_HEIGHT`?

apps/beeai-ui/src/modules/compose/sequential/SequentialSetup.tsx
```diff
@@ -55,74 +54,71 @@ export function SequentialSetup() {
...
+            </Button>
+          ) : (
+            <Button
+              renderIcon={StopOutlineFilled}
+              kind="tertiary"
+              size="md"
+              iconDescription="Cancel"
+              onClick={(e) => {
+                onCancel();
+                e.preventDefault();
```
penge, line 116: I think preventing the default first, then calling the other action, may be more natural:
```
e.preventDefault();
onCancel();
```

Btw. What is the default here? Why preventing it?

apps/beeai-ui/src/components/SplitPanesView/SplitPanesView.module.scss
```diff
@@ -16,21 +16,21 @@
 
 $max-container: get-value($grid-breakpoints, 'max.width');
 
-.root {
+.splitView {
+  display: grid;
+  grid-template-columns: 1fr calc((100vw / 3 * 2) - ((100vw - #{$max-container}) / 2 / 3));
```
kapetr: The second pane grows beyond 2/3 of the viewport on smaller screens, was that intentional? Shouldn't we enforce a consistent 1:2 ratio for the pane contents?

apps/beeai-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+'use client';
```
kapetr: Just curious—why was this necessary? 'use client' is meant to separate the client and server parts of the app, so typically, it should only be in one file. But I see it was added to MainNav, so this was probably needed before that update?

kapetr (approved):

apps/beeai-ui/src/components/MainContentView/MainContentView.tsx
PetrBulanek: I guess it served a purpose before, but not anymore. I've looked through it and it doesn't appear anywhere, so I've removed the option altogether.

apps/beeai-ui/src/components/MainContentView/MainContentView.tsx
PetrBulanek: I would like to move it into layout component, but that would actually mean a lot more changes. The main component that scrolls with the whole page content is MainContentView, again it had some reasoning behind it, which is currently not there anymore as there were lots of quick changes. However we want the AppFooter scroll with the content and moving it to the layout component would mean change the scrolling context and update all other components that relate to it. This would have some additional impacts on various components, so it's up for a broader discussion if that makes sense. If we do redo it, I'd be a little worried about redoing it again in a while with some other UI tweak. For now, I'd probably leave it like this.

apps/beeai-ui/src/components/MainNav/MainNav.module.scss
PetrBulanek: I dare to disagree. :-)

I've worked on projects where there were different approaches to writing CSS for media queries. But most of the time, the first way was always to write a media query rule directly to a specific class, as you suggest here. And then the second approach, where you always wrote media query rules for each breakpoint individually and grouped the CSS classes under it.

And the other way has always been clearer. Especially for larger components where there are more CSS classes, it's noticeable. You just have the rules for responsive behavior in one place and you don't have to search all over the file for it.

I prefer this:

```
.classname {}

.other-classname {}

.third-classname {}

@include breakpoint-up(md) {
  .classname {}

  .other-classname {}
}

@include breakpoint-up(lg) {
  .other-classname {}

  .third-classname {}
}
```

over this:

```
.classname {
  @include breakpoint-up(md) {}
}

.other-classname {
  @include breakpoint-up(md) {}

  @include breakpoint-up(lg) {}
}

.third-classname {
  @include breakpoint-up(lg) {}
}
```

Aside from clarity, it's usually less typing.

apps/beeai-ui/src/components/SplitPanesView/SplitPanesView.tsx
PetrBulanek: In fact, I don't see any benefit here. I am just unnecessarily creating some structure (object) where from my point of view it is not needed. I don't see any duplication there. Or am I missing something? :)

apps/beeai-ui/src/hooks/useToTopButton.ts
PetrBulanek: It's a pretty random offset that I found ok and functional. I'm not trying to match it to the size of anything, so I think it would be just confusing to name it something else. Besides, I'm not adding it in this PR, it's been there before. :-)

apps/beeai-ui/src/components/MainNav/MainNav.module.scss
kapetr: I'm on the fence on this and leaning a little bit towards the @penge's approach because I prefer having all the rules of a class in one place including media queries, so I don't have to scroll somewhere to see how it behaves on another device (and meanwhile forget what were the original rules).

But both approaches have their merits and the difference in small files (and we want to keep them small) is not that significant, so I wouldn’t strongly enforce either :)

apps/beeai-ui/src/modules/compose/sequential/SequentialSetup.tsx
PetrBulanek: I am not sure. This code has been here before. I didn't add it as part of this PR.

PetrBulanek: cc. @kapetr

apps/beeai-ui/src/components/MainNav/MainNav.module.scss
penge: I think the second example is much better. (easier to maintain, readable, scales better)

But I guess it's a matter of preference.

apps/beeai-ui/src/modules/compose/sequential/SequentialSetup.tsx
penge: I thought it's new. OK then.

apps/beeai-ui/src/components/SplitPanesView/SplitPanesView.module.scss
PetrBulanek: There was actually a mistake I didn't notice. On smaller screen this part `(100vw - #{$max-container})` could actually be negative value, that eventually causes this column to grow, which is not desirable. I fixed it with wrapping it with max fn `max(100vw - #{$max-container}, 0px)`. Thanks!

apps/beeai-ui/src/components/Tooltip/Tooltip.tsx
PetrBulanek: The Next.js app build required it. I didn't look deeper into it.

But now I see that the component is not used anywhere, so maybe it was caused by some forgotten import that is no longer there. I removed it. The 'use client' in MainNav should be enough.

apps/beeai-ui/src/components/MainNav/MainNav.module.scss
PetrBulanek: Got it. So it seems to be mostly about personal preference. So I guess we don't enforce it either way. Alternatively, I'll adapt if we keep the components/styles as small as possible to keep it clear anyway. :-)

penge (approved):

PR #441, opened 2025-03-21
chore(web): add preview task for beeai-web

PetrBulanek:


apps/beeai-web/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "NODE_OPTIONS=\"--no-experimental-global-navigator\" next dev",
     "build": "NODE_OPTIONS=\"--no-experimental-global-navigator\" next build",
-    "start": "NODE_OPTIONS=\"--no-experimental-global-navigator\" next start",
+    "preview": "NODE_OPTIONS=\"--no-experimental-global-navigator\" next start",
```
penge, line 8: I would definitely stick to the name next uses and match it:
```
start: "... start",
```

The PR could be then named:
```
chore(web): add start task
```

PetrBulanek: I was actually matching it with the same task we have in beeai-ui https://github.com/i-am-bee/beeai/blob/1cda1bf499dabd79fa4f53511c28facea2682239/apps/beeai-ui/package.json#L12, so there is no mental overhead and you just use same name for both projects.

penge: That is interesting! LD
I would personally not try to match it. One is Vite, the other one is Next.

penge (approved):

PR #445, opened 2025-03-21
docs: update how to compose

jenna-winkler:


docs/how-to/compose-agents.mdx
```diff
@@ -1,14 +1,39 @@
...
+    <Step title="Run your workflow">
+        Click the **Run** button to execute the composition.
+    </Step>
+</Steps>
 
-In the meantime, bee 🐝 curious—dive into our [GitHub repository](https://github.com/i-am-bee/beeai) to explore the code or even lend a hand! Your contributions help the community grow and make this documentation bloom faster.
+## CLI
 
-Thanks for your patience, and bee-lieve us—we appreciate your support in keeping our hive buzzing with open-source goodness!
+You can compose agents directly from your terminal by running the sequential-workflow agent.
```
tomkis: @jenna-winkler 

Would it be possible to further explain the concept of instruction?

Basically user should know that sequential workflow gives each agent instruction and output from previous agent is added to the following agent alongside the instruction.

tomkis: Also they don't necessarily need to provide the payload.

You can run the sequential workflow just by `beeai run sequential-workflow` and then follow the CLI UI to provide the data. The CLI UI mimics the way it works in the GUI.

tomkis (changes requested):

docs/how-to/compose-agents.mdx
jenna-winkler: @tomkis done!

tomkis (approved):

PR #490, opened 2025-03-31
docs(maintainers): Create MAINTAINERS.md

jenna-winkler:


penge (commented):
I have checked the names, their order, and the GitHub usernames, and **they match**.

One thing I noticed:
There is a reference to 'Project Governance,' but the document is titled 'Governance,' so aligning it to just 'Governance' may be preferable.

matoushavlena (approved):

PR #504, opened 2025-04-03
feat(beeai-cli): improve beeai-server waiting and management

JanPokorny:
- detect local process and its creation time
- wait for server startup if it's recently spawned
- wait for API in cases where it may not be available yet


jezekra1 (changes requested):
Nice! :)
please add `psutil` to pyproject

jezekra1 (approved):

PR #515, opened 2025-04-07
fix: prevent code block opening

QuantGeekDev:
Hey everyone. Thanks for creating beeai, there a lot of interesting approaches taken throughout the codebase and it's been fun to read. I wish my first contribution was more significant, but here's a quick fix on the quickstart docs.

Before screenshot:
image

After screenshot:
image


JanPokorny:
Hello, thank you for pointing this out, I see that we fixed it in the meantime

PR #516, opened 2025-04-08
docs: update copyright

jenna-winkler:


JanPokorny:
@jenna-winkler Is the copyright year intentionally missing?

jenna-winkler:
> @jenna-winkler Is the copyright year intentionally missing?

@JanPokorny good catch I checked with the LF because this is the string they provided me with but they said that wasn’t intentional we can include year

JanPokorny:
@jenna-winkler OK nice, it's easier if I don't have to hack NWA to _not_ include year 😁 In that case I'll just update the NWA config in this PR so that the checks pass

Issue #455, opened 2025-03-24
Agent Installation


tomkis:
We are changing the paradigm of agents availability.

Once platform is installed all the agents will not be available until installed, they will show up in the agents catalog but user has to first install the agent in order to use it.

The same user experience can be achieved through both UI and CLI.

**List of Agents**

The list of agents should have na icon indicating whether agent is installed or not. This information has to be provided through the platform API.

There are three possible states:

- Installed
- Not Installed
- Installing
- Error

Upon clicking, the icon should turn into loading state indicating that downloading of agent is in progress. There's no need to visualize numerical progress, it's just state.

**Agent Detail**

When agent is being installed, the detail of the agent should reflect that information and should not allow to launch until the installation is finished. 

Not-installed agent should not allow launching as well.

No need to visualise the progress, just the state of downloading.

Installation of agent may fail and this should be somehow communicated with the user.

**Acceptance Criteria**
- [ ] Agent List GUI in local experience shows three state icon whether agent is installed / installing / not-installed
- [ ] Clicking on uninstalled agent results on initiating installation of the agent
- [ ] Multiple agents can be installed in parallel
- [ ] Clicking on installed / installing agent won't result in any action
- [ ] Once installation of the agent is finished, the status in the list changes from Installing to Installed
- [ ] Failing of installation should be communicated to the user
- [ ] Installation of agent should happen in the background, when browser is killed, the installation continues
- [ ] CLI list of agents should show state of the agent (Installing, Installed, Not Installed)
- [ ] CLI should allow `beeai install agent-name`
- [ ] CLI should handle cases where user attempts to either install currently installing agent or already installed agent gracefully
- [ ] CLI - installing agent should block the process unlike GUI
- [ ] CLI - Failing of installation should be communicated to the user

PR #531, opened 2025-04-09
feat(ui): add agent installation

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai/issues/455


apps/beeai-ui/src/modules/agents/detail/AgentLaunchButton.tsx
```diff
@@ -37,29 +43,76 @@ interface Props {
...
+  const uiType = ui?.type;
+  const sharedProps: ButtonBaseProps = {
+    kind: 'primary',
+    size: 'md',
+    className: classes.button,
+  };
+
+  const isNotInstalled = status === ProviderStatus.NotInstalled;
+  const isInstalling = status === ProviderStatus.Installing;
+  const isInstallError = status === ProviderStatus.InstallError;
```
kapetr: This is repeated in 3 files, it might be worth moving into `useAgentStatus`.

apps/beeai-ui/src/modules/phoenix/api/queries/usePhoenix.ts
```diff
@@ -23,7 +23,6 @@ import { phoenixKeys } from '../keys';
 export function usePhoenix() {
   const query = useQuery({
     queryKey: phoenixKeys.all(),
-    refetchInterval: 5_000,
```
kapetr, line 26: I know we talked about this, but I'm not sure we want remove it completely (maybe just for local dev?). I believe this regular status check is still a valid feature. 
cc @pilartomas

apps/beeai-ui/src/modules/providers/hooks/useMonitorProviderStatus.ts
```diff
@@ -37,41 +37,56 @@ export function useMonitorProvider({ id }: Props) {
 
   const { data: provider, refetch: refetchProvider } = useProvider({
     id,
-    refetchInterval: (data) => (data?.status === 'ready' ? false : CHECK_PROVIDER_STATUS_INTERVAL),
+    refetchInterval: (data) => {
+      const status = data?.status;
+      const isReady = status === ProviderStatus.Ready;
+
+      return isReady ? false : CHECK_PROVIDER_STATUS_INTERVAL;
```
kapetr: Should the status be checked for `installing` status here instead? It looks like the check continues even after an error occurs.

kapetr (approved):

apps/beeai-ui/src/modules/agents/detail/AgentLaunchButton.tsx
PetrBulanek: Good idea, I'll add the helpers directly to the hook.

apps/beeai-ui/src/modules/phoenix/api/queries/usePhoenix.ts
PetrBulanek: I'll remove it just for local development.

apps/beeai-ui/src/modules/providers/hooks/useMonitorProviderStatus.ts
PetrBulanek: With the new agent lifecycle, I could have simplified the whole thing a lot. Thanks for the nudge! :-)

kapetr (approved):

PR #606, opened 2025-05-01
feat(deploy): add helm chart

jezekra1:


chandra-potula:
I have tested the code changes on OpenShift environment and with minor patch I was able to deploy the solutions as containers. So looking forward for the merger of these code changes soon to main repository.

PR #608, opened 2025-05-02
docs: Update README.md

jenna-winkler:


README.md
```diff
@@ -7,7 +7,7 @@
   BeeAI
```
jezekra1: I believe @geneknit wanted to put Platform here
```suggestion
  BeeAI Platform
```

PR #632, opened 2025-05-13
feat(llm-gateway): implement

JanPokorny:
- add a LLM gateway to beeai-server
- wire all agents to use the LLM gateway
- connect the gateway to the existing env var setup system (temp solution until we have a proper DB for storing per-user settings)
- add watsonx support
- add RITS support

### DIFF
--- a/apps/beeai-server/src/beeai_server/routes/llm.py
+++ b/apps/beeai-server/src/beeai_server/routes/llm.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import re
 import time
 import uuid
 from typing import Dict, List, Literal, Optional, Union, AsyncGenerator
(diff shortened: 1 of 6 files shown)


apps/beeai-server/src/beeai_server/routes/llm.py
```diff
@@ -0,0 +1,212 @@
...
+            model=request.model,
+            choices=[
+                ChatCompletionResponseChoice(
+                    message=ChatCompletionMessage(content=output.get_text_content()),
+                    finish_reason=output.finish_reason,
+                )
+            ],
+        )
+
+    except Exception as e:
```
jezekra1: [nitpick] This shouldn't be necessary, server errors are covered in `application.py`. Lets' improve the `extract_messages` utility if the error message is not nice enough :)

jezekra1 (approved):

jezekra1 (approved):

PR #657, opened 2025-05-20
Updated default model for ollama setup to granite

### LINKED ISSUE #656: Update recommended model to Granite
### Issue

The default recommended model for Ollama setup is llama, but it should be granite.

### Current behavior

When running the `beeai env setup`, the recommended model currently is `llama3.1:8b`.

### Expected behavior

When running the `beeai env setup`, the recommended model should be either `granite3.2:8b` or `granite3.3:8b`.

vedem1192:
Small update to beeai-cli to point to granite3.2:8b during env setup with Ollama.

Closes #656

### DIFF
--- a/apps/beeai-cli/src/beeai_cli/commands/env.py
+++ b/apps/beeai-cli/src/beeai_cli/commands/env.py
@@ -108,7 +108,7 @@ async def setup() -> bool:
             ),
             Choice(
                 name="IBM watsonx".ljust(25) + "🚧 experimental",
-                value=("watsonx", None, "ibm/granite-3-8b-instruct"),
+                value=("watsonx", None, "ibm/granite-3-3-8b-instruct"),
             ),
             Choice(
                 name="Ollama".ljust(25) + "💻 local", value=("Ollama", "http://localhost:11434/v1", "granite3.3:8b")
--- a/apps/beeai-cli/src/beeai_cli/commands/env.py
+++ b/apps/beeai-cli/src/beeai_cli/commands/env.py
@@ -110,7 +110,9 @@ async def setup() -> bool:
                 name="IBM watsonx".ljust(25) + "🚧 experimental",
                 value=("watsonx", None, "ibm/granite-3-8b-instruct"),
             ),
-            Choice(name="Ollama".ljust(25) + "💻 local", value=("Ollama", "http://localhost:11434/v1", "llama3.1:8b")),
+            Choice(
+                name="Ollama".ljust(25) + "💻 local", value=("Ollama", "http://localhost:11434/v1", "granite3.2:8b")
+            ),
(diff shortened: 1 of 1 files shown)


apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -110,7 +110,9 @@ async def setup() -> bool:
                 name="IBM watsonx".ljust(25) + "🚧 experimental",
                 value=("watsonx", None, "ibm/granite-3-8b-instruct"),
             ),
-            Choice(name="Ollama".ljust(25) + "💻 local", value=("Ollama", "http://localhost:11434/v1", "llama3.1:8b")),
+            Choice(
+                name="Ollama".ljust(25) + "💻 local", value=("Ollama", "http://localhost:11434/v1", "granite3.2:8b")
+            ),
```
penge, line 115: How about we go for `granite3.3:8b`?

tomkis: @geneknit can u elaborate please? we use 3.3-8b for granite-chat agent.

vedem1192: @penge Sure, I'll update

@tomkis It might be used it for chat, but as mentioned by @geneknit , it would look better if the default proposed model to pull from Ollama is Granite and not Llama. Also, keeping in mind we want as many people as possible to get their hands on Granite, it's good to give it visibility.

penge (approved):
LGTM!

**Note:** We might want to update IBM watsonx model to `ibm/granite-3-3-8b-instruct` as well.

apps/beeai-cli/src/beeai_cli/commands/env.py
JanPokorny, line 1: Have you tested how well the built-in agents work with Granite? It wouldn't be a very good look for Granite if some of them were broken

vedem1192: @JanPokorny 
I didn't initially, but just ran all current Beeai agents against granite3.3:8b. All worked as expected but the `marketing_strategy` agent. Seeing this, I tried running that agent with llama3.1:8b, which also resulted in an error. I'll open an issue for this, but seeing how that issue appears to be agent agnostic, I think we can go ahead with granite3.3:8b

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -110,7 +110,9 @@ async def setup() -> bool:
                 name="IBM watsonx".ljust(25) + "🚧 experimental",
                 value=("watsonx", None, "ibm/granite-3-8b-instruct"),
```
JanPokorny: Can you also update this one to 3.3?

penge (approved):

PR #658, opened 2025-05-20
feat(ui): continue with ui rework

kapetr:
- first agent as landing
- agent detail improvements
- enable settings by feature flag


apps/beeai-ui/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,38 @@
...
+    console.warn('Unable to parse feature flags!', err);
+    features = Object.fromEntries(Object.keys(FeatureName).map((feature) => [feature, false])) as Record<
+      FeatureName,
+      boolean
+    >;
+  }
+  return features;
+}
+
+export const FEATURE_FLAGS = parseFeatureFlags(import.meta.env.VITE_FEATURE_FLAGS);
```
penge: It would be great to see `.env.example` with an example structure.

apps/beeai-ui/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,38 @@
...
+    features = Object.fromEntries(
+      Object.keys(FeatureName).map((feature) => [feature, feature in allFeatures ? !!allFeatures[feature] : false]),
+    ) as Record<FeatureName, boolean>;
+  } catch (err) {
+    console.warn('Unable to parse feature flags!', err);
+    features = Object.fromEntries(Object.keys(FeatureName).map((feature) => [feature, false])) as Record<
+      FeatureName,
+      boolean
+    >;
+  }
```
penge: Can **Zod** help us with the parsing? It could define the schema, parse, and produce the errors too.

apps/beeai-ui/src/components/layouts/AppSidebar.tsx
```diff
@@ -16,14 +16,26 @@
...
-      <AgentsNav />
+      <div className={classes.root}>
+        <AgentsNav />
+
+        {FEATURE_FLAGS.Settings && (
+          <div className={classes.footer}>
+            <UserNav />
+          </div>
+        )}
+      </div>
```
penge: I am missing the correlation between `Settings` << >> `UserNav`. How about the feature is named `Navigation` or `UserNavigation`?

apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
```diff
@@ -37,7 +34,8 @@ export function AgentDetailPanel() {
 
   if (!agent) return null;
 
-  const { name, description } = agent;
+  const { description, metadata } = agent;
+  const documentationUrl = metadata.links?.find(({ type }) => type === 'documentation')?.url;
```
penge: Helper:

```ts
function getLinkUrl<T extends components['schemas']['LinkType']>(
  links: components['schemas']['Metadata']['links'], type: T
): string | undefined {
  return links?.find((link) => link.type === type)?.url;
}
```

Use:

```ts
const documentationUrl = getLinkUrl(metadata.links, "documentation");
```

What do you think?

But it would be ideal, if Links would be a Record. Because then, we could do:

```ts
metadata.links?.Documentation?.url
```

apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
```diff
@@ -55,13 +53,15 @@ export function AgentDetailPanel() {
                 <>
                   {description && <MarkdownContent className={classes.description}>{description}</MarkdownContent>}
 
-                  <AgentMetadata agent={agent} />
+                  {metadata.author && <span>By {metadata.author.name}</span>}
```
penge: This is a very detail. But could be nice to have as:
```tsx
const authorName = metadata.author.name;

// ...

{authorName && <span>By {authorName}</span>}
```

Just to follow the same pattern as we have for the `documentationUrl`. It also reads nice, and makes sure "name" is defined.

apps/beeai-ui/src/modules/agents/components/AgentTags.tsx
```diff
@@ -29,8 +30,21 @@ interface Props {
...
+  });
+
+  const tags = [
+    framework,
+    avg_run_time_seconds && `${avg_run_time_seconds}s/run (avg)`,
+    avg_run_tokens && `${avg_run_tokens} tokens/run (avg)`,
+    license,
+  ]
+    .map((value) => (value ? <AgentTag key={value} name={value} /> : null))
+    .filter(isNotNull);
```
penge: Swapping these two lines could help:
```tsx
    .filter(isNotNull)
    .map((value) => <AgentTag key={value} name={value} />);
```

apps/beeai-ui/src/components/layouts/AppSidebar.tsx
kapetr: Oh yes, this is kind of interim solution and will be subject to change. But you're right, I'll rename it to reflect the current state, the worst thing, we'll revert the name back, if it really controls just the Settings link.

penge (approved):

apps/beeai-ui/src/App.tsx
```diff
@@ -44,10 +45,7 @@ export function App() {
                 <BrowserRouter>
                   <Routes>
                     <Route element={<AppLayout />}>
-                      <Route
-                        path={routeDefinitions.home()}
-                        element={<Navigate to={routeDefinitions.agents()} replace />}
-                      />
+                      <Route path={routeDefinitions.home()} element={<Landing />} />
```
vedem1192, line 48: Hi! This change seems to have created the following behavior :

When locally running `beeai ui`, the landing page becomes `http://localhost:8333/agents/sequential_workflow/run` instead of the previous `http://localhost:8333/agents`. Since in my case I don't have the agent by default, this is the view I see : 


Screenshot 2025-05-22 at 9 59 56 AM

Is this the expected behavior?

vedem1192: @kapetr I can open an issue for this, but I wanted to confirm the expected behavior beforehand 😅

kapetr: It is the expected behaviour except we should filter out the agents without ui :) So feel free to open an issue, thanks!

PR #666, opened 2025-05-21
feat(platform): rewrite for k8s, add persistence

jezekra1:


apps/beeai-cli/src/beeai_cli/commands/telemetry.py
```diff
@@ -12,19 +12,28 @@
...
+        sharing_enabled = not disable
+        if sharing_enabled != telemetry_config["sharing"]:
+            save_telemetry_config(False)
+            helm_chart = HelmChart(
+                {"metadata": {"name": "beeai", "namespace": "default"}},
+                api=await kr8s.asyncio.api(kubeconfig=KUBECONFIG),
+            )
+            if await helm_chart.exists():
+                await helm_chart.patch({"spec": {"set": {"telemetry.sharing": str(sharing_enabled).lower()}}})
+    console.print(get_telemetry_config())
```
JanPokorny: We can simplify this since we're switching from brew service (which can't be parametrized) to a Lima service (which can). The user can disable telemetry when starting the platform using: `beeai platform start --set telemetry.sharing=false` (this also works for updating an existing platform). So this whole command is perhaps not necessary anymore.

apps/beeai-server/docker-compose.yaml
JanPokorny, line 1: Is this necessary? I'd prefer if we unified the dev setup on Lima + `telepresence`

apps/beeai-server/scripts/bootstrap_dev.sh
JanPokorny, line 1: Why does this need a separate install of DB and service/deployment, wouldn't telepresence work with `beeai platform start`?

JanPokorny: Also, scripts can directly register as Mise tasks: https://mise.jdx.dev/tasks/file-tasks.html

apps/beeai-server/src/beeai_server/application.py
```diff
@@ -129,54 +125,40 @@ def mount_routes(app: FastAPI):
...
+    #                 count += 1
+    #         yield Observation(
+    #             value=count,
+    #             attributes={
+    #                 "status": status,
+    #             },
+    #         )
 
-    meter.create_observable_gauge("providers_by_status", callbacks=[scrape_providers_by_status])
+    # meter.create_observable_gauge("providers_by_status", callbacks=[scrape_providers_by_status])
```
JanPokorny, line 153: Forgotten commented out code

apps/beeai-server/src/beeai_server/bootstrap.py
```diff
@@ -14,171 +14,34 @@
...
+def setup_database_engine(config: Configuration) -> AsyncEngine:
+    return create_async_engine(str(config.persistence.db_url.get_secret_value()), isolation_level="SERIALIZABLE")
+
+
+async def setup_kubernetes_client():
+    ns_path = Path("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+    namespace = None
+    if await ns_path.exists():
+        namespace = (await ns_path.read_text()).strip()
+    return await kr8s.asyncio.api(namespace=namespace)
```
JanPokorny: The file should always exist (since we're always in kubernetes), and if not, it should crash on that

```suggestion
    return await kr8s.asyncio.api(namespace=(await Path("/var/run/secrets/kubernetes.io/serviceaccount/namespace").read_text()).strip())
```

apps/beeai-server/src/beeai_server/service_layer/telemetry.py
```diff
@@ -14,33 +14,28 @@
...
-        await self.sync()
-
-    async def sync(self):
-        await self._repository.sync()
-        if self._config.collector_managed:
-            await self._manager.reload()
+        raise NotImplementedError
+        # config = await self._repository.get()
+        # await self._repository.set(config=config.model_copy(update={"sharing_enabled": sharing_enabled}))
+        # await self._manager.reload()
```
JanPokorny: Forgotten commented out code?

apps/beeai-server/src/beeai_server/utils/github.py
```diff
@@ -13,20 +13,11 @@
...
-import shutil
-from collections import defaultdict
 from enum import StrEnum
-from os import PathLike
-from backports.tarfile import ExtractError
 from typing import Any
 
-import anyio
-import anyio.to_thread
-import backports.tarfile as tarfile
```
JanPokorny, line 27: Remove `backports-tarfile` from dependencies

apps/beeai-server/src/beeai_server/domain/models/provider.py
```diff
@@ -0,0 +1,174 @@
...
+#         if not with_dummy_env:
+#             self.check_env(env)
+#
+#         required_env_vars = {var.name for var in self.env if var.required}
+#         env = {
+#             **self._global_env,
+#             **({var: "dummy" for var in required_env_vars} if with_dummy_env else {}),
+#             **(self.extract_env(env=env)),
+#         }
+#         port = str(await find_free_port())
```
JanPokorny: Forgotten commented out code?

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
```diff
@@ -0,0 +1,237 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class KubernetesProviderDeploymentManager(IProviderDeploymentManager):
+    def __init__(self, api: kr8s.Api):
+        self._api = api
+
+    def _get_k8s_name(self, provider_id: UUID, kind: str | None = None):
+        return f"beeai-provider-{provider_id}" + (f"-{kind}" if kind else "")
```
JanPokorny, line 44: AFAIK this can be at most 63 characters, so long provider names may break this

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
```diff
@@ -0,0 +1,237 @@
...
+        )
+        deployment_hash = hashlib.sha256(combined_manifest.encode()).hexdigest()[:63]
+        deployment_manifest["metadata"]["labels"]["deployment-hash"] = deployment_hash
+
+        deployment = Deployment(deployment_manifest, api=self._api)
+        try:
+            existing_deployment = await Deployment.async_get(deployment.metadata.name, api=self._api)
+            if existing_deployment.metadata.labels["deployment-hash"] == deployment_hash:
+                if existing_deployment.replicas == 0:
+                    await deployment.async_scale(1)
```
JanPokorny: You could instead use `kr8s.asyncio.Api` for an async API -- which allows just `await deployment.scale(1)` etc. without the prefix. Also it initializes asynchronously (things like looking for service credentials), but that's a detial.

apps/beeai-cli/src/beeai_cli/commands/telemetry.py
jezekra1: aha I thought we were still going to use brew services, in that case it's better to remove it

apps/beeai-server/docker-compose.yaml
jezekra1: oops, this is leftover from my previous setup

apps/beeai-server/scripts/bootstrap_dev.sh
jezekra1: Not sure, would have to try it out - probably would need to somehow disable the beeai-server that comes with the chart which might interfere with locally running instance.

This is not actually that much overhead and it's maybe nicer that it's separate - you can develop and then test it immediately by building and deploying to a separate namespace without interrupting the dev setup.

apps/beeai-server/src/beeai_server/bootstrap.py
jezekra1: It doesn't exist during development on local machine, should I add a separate `namespace` and if it's not set then default to this path and crash potentially?

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
jezekra1: Good find, I calculated this, `provider_id` is uuid4, so it's fine, but adding a trim won't hurt :D

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
jezekra1: I don't think I can use the Api objects without `async` even with async api, there is some mixin that's syncifying all normal methods
image

apps/beeai-server/scripts/bootstrap_dev.sh
jezekra1: hm, found out there is a `--replace` flag in telepresence, that might do the job!

JanPokorny: Yeah `telepresence` can completely disable the pod you are telepresenting into, that's what I was thinking of for the dev setup: `beeai platform start` and then replace the container(s) you are developing

apps/beeai-server/src/beeai_server/bootstrap.py
JanPokorny: Oh right, it does not work when telepresenting -- in that case we may want to add a config option for the namespace to deploy to, so that we can set it for local dev (also potentially useful to have in general)

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
JanPokorny: It works here: https://github.com/i-am-bee/beeai-platform/blob/f50d0e3656b5ae9d1f7c99b7b5d882184d51a995/apps/beeai-cli/src/beeai_cli/commands/platform.py#L184

apps/beeai-cli/src/beeai_cli/commands/telemetry.py
jezekra1: removed

apps/beeai-server/scripts/bootstrap_dev.sh
jezekra1: leaving this to separate PR

apps/beeai-server/src/beeai_server/application.py
jezekra1: added TODO

apps/beeai-server/src/beeai_server/bootstrap.py
jezekra1: added optional `k8s_namespace` to configuration

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
jezekra1: aha, it needs to be imported from `kr8s.asyncio.objects` not `kr8s.objects`

apps/beeai-server/src/beeai_server/infrastructure/kubernetes/provider_deployment_manager.py
jezekra1: I left it without trim since we're extracting provider id again from the name just below. Let's just rely on the uuid length, it's even reflected in the parameter type

Issue #645, opened 2025-05-16
Recommended model visualised in Agent UI


tomkis:
There's recommended model in agent metadata we should visualise that underneath the chat / handoff prompt in the Agent UI. The component will eventually evolve into a dropdown to allow model overriding.

jenna-winkler:
It is my understanding that today model is hard coded. As a next step, we want to leverage this parameter of ACP metadata. Can we have it so that the first model defined is the default for the user in the UI? And the user can toggle to change it if they want?

```
    "recommended_models": [
      "llama3.3:70b-instruct-fp16"
    ]
```

tomkis:
> It is my understanding that today model is hard coded. As a next step, we want to leverage this parameter of ACP metadata. Can we have it so that the first model defined is the default for the user in the UI? And the user can toggle to change it if they want?
> 
> ```
>     "recommended_models": [
>       "llama3.3:70b-instruct-fp16"
>     ]
> ```

That would be the low hanging fruit.

However we were thinking of something more advanced, basically answer things like:

- What if user wants to provide family of models?
- What if user wants to define different models for different tasks (reasoning model, "dumb model"...)

jenna-winkler:
Important consideration that came from community feedback - not every agent requires an LLM, so we need to consider cases where there is no LLM cc @tomkis @anafucs

anafucs:
Should we just remove the model from input field?

jenna-winkler:
Yes this is what I am thinking for when no model is defined

PR #667, opened 2025-05-21
feat(ui): display agent model, update chat landing

kapetr:
Refs #645

### DIFF
--- a/apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
+++ b/apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
@@ -42,10 +42,10 @@
 }
 
 .heading {
+  @include font-family(serif);
   font-size: rem(42px);
   line-height: math.div(60, 42);
   font-weight: 300;
-  @include font-family(serif);
 }
 
 .content {
--- a/apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
+++ b/apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
@@ -17,6 +17,10 @@
 .root {
   block-size: 100%;
   display: flex;
+
+  &.isNew {
+    align-items: center;
+  }
 }
 
 .holder {
@@ -26,7 +30,22 @@
 }
 
 .header {
-  padding-block: $spacing-03;
+  .isNew & {
+    padding-block: $spacing-03;
+    display: flex;
+    flex-direction: column;
+    gap: $spacing-06;
+    h1 {
+      margin: 0;
+    }
+  }
+}
+
+.heading {
+  font-size: rem(42px);
+  line-height: math.div(60, 42);
+  font-weight: 300;
+  @include font-family(serif);
 }
 
 .content {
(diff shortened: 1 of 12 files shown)


apps/beeai-ui/src/modules/runs/components/AgentModel.tsx
```diff
@@ -0,0 +1,40 @@
...
+import classes from './AgentModel.module.scss';
+
+export function AgentModel() {
+  const { data, isPending } = useListVariables();
+
+  if (isPending) {
+    return null;
+  }
+
+  const model = data?.env['LLM_MODEL'] ?? AGENT_DISPLAY_MODEL_TEMP;
```
penge: Can be simplified without the use of a string literal:
```ts
const model = data?.env.LLM_MODEL ?? AGENT_DISPLAY_MODEL_TEMP;
```

apps/beeai-ui/src/modules/runs/components/AgentModel.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { isGraniteModel } from '../utils';
+import classes from './AgentModel.module.scss';
+
+export function AgentModel() {
+  const { data, isPending } = useListVariables();
+
+  if (isPending) {
+    return null;
+  }
```
penge: Maybe we can simplify, and check only for data (status is irrelevant):
```tsx
const { data } = useListVariables();

if (!data) {
  return null;
}

const model = data.env.LLM_MODEL ?? AGENT_DISPLAY_MODEL_TEMP;
```

apps/beeai-ui/src/modules/runs/components/InputBar.module.scss
```diff
@@ -58,6 +58,12 @@
   column-gap: $spacing-04;
 }
 
+.actionBarLeft {
```
penge: Because we are using CSS logical properties in this project, a more proper name would be:
```css
.actionBarStart
```

apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
```diff
@@ -26,7 +30,22 @@
...
+      margin: 0;
+    }
+  }
+}
+
+.heading {
+  font-size: rem(42px);
+  line-height: math.div(60, 42);
+  font-weight: 300;
+  @include font-family(serif);
```
penge: Let's put include first:
```scss
.heading {
  @include font-family(serif);
  // ...
}
```

**Reasons:** Easier to scan, Common convention, Reduces override risk.

apps/beeai-ui/src/modules/runs/components/AgentModel.tsx
kapetr: The point is that fetch will fail in production deployment, because `/variables` endpoint is restricted - `AGENT_DISPLAY_MODEL_TEMP` needs to be displayed in that case. 
I'll add the explaining comment, it's temporary solution, until we have the current agent model from API.

penge: Ah.... I see. Then some explanatory would be great!

penge (approved):

PR #676, opened 2025-05-22
feat 635 --> playground

### LINKED ISSUE #635: Move compose playground as a separate UI
The Compose Playground is primarily used for demo purposes and can be confusing for business users, who may not need or understand the composition of agents. This overlap in functionality creates friction for non-technical users, while still being a valuable tool for developers.

Move the Compose Playground to a separate UI, accessible through a dedicated URL that is not linked from the main app. It should be accessible via the CLI using the command `beeai compose`. This approach would keep the main app cleaner for business users while still providing developers with the tools they need.

An alternative would be to keep all compose agents as out-of-the-box (OOTB) agents with a custom UI, reducing the need for a separate playground. However, feedback indicates that users often struggle to understand the concept of composition agents, potentially due to suboptimal messaging, making this approach less favorable.

vedem1192:
Hey all!
Opening a PR for issue #635. 

### What does this do?

This separates the "Compose Playground" from the rest of the application. It's still there, but not directly accessible from the UI tab. A new CLI command has been added to start the playground, which now live at `/playground`.

### Added feature 

New CLI command : `beeai playground`

### Notes 

All issue subtasks are done, aside from the video update. This will require some input from the design team. 

Closes #635


vedem1192:
@kapetr should be good now! thanks :D

tomkis (approved):

PR #706, opened 2025-05-27
fix(server): logs command not working

jezekra1:


JanPokorny:
@jacobtomlinson I dug up some old messages with Radek, back then he said that the issue was that `kr8s` was re-using a single `httpx` client inside and it was causing issues as different requests on the same client were influencing each other. Not sure if that helps, but it's all I could find.

jacobtomlinson:
Thanks for getting back to me @JanPokorny! That sounds like the kind of thing that could be reported upstream to `httpx`.

PR #733, opened 2025-06-03
feat(ui): tweak ui after rework

PetrBulanek:
Refs: https://github.com/i-am-bee/beeai-platform/issues/707

### DIFF
--- a/apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
+++ b/apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
@@ -27,6 +27,7 @@ import type { AgentPageParams } from '../types';
 import { getAvailableAgentLinkUrl } from '../utils';
 import classes from './AgentDetailPanel.module.scss';
 import { AgentTags } from './AgentTags';
+// import { AgentTools } from './AgentTools';
 
 export function AgentDetailPanel() {
   const { agentName } = useParams<AgentPageParams>();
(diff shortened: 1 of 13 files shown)


apps/beeai-ui/src/modules/agents/components/AgentTags.tsx
```diff
@@ -46,7 +46,7 @@ export function AgentTags({ agent, className }: Props) {
     .filter(isNotNull)
     .map((value) => <AgentTag key={value} name={value} />);
 
-  return <TagsList tags={tags} className={className} />;
+  return tags.length > 0 ? <TagsList tags={tags} className={className} /> : null;
```
penge: How about `<TagsList>` decides what to render based on the _length_ of the **_tags_**?

kapetr (approved):

apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
```diff
@@ -51,27 +52,37 @@ export function AgentDetailPanel() {
...
+                  <>
+                    <SkeletonText paragraph={true} lineCount={5} />
+                  </>
+                )}
+              </div>
             </TabPanel>
+
+            {/* <TabPanel>
+              <AgentTools />
+            </TabPanel> */}
```
penge, line 85: Do we need this?

PetrBulanek: I was about to implement the tools list there, but it has its own separate issue https://github.com/i-am-bee/beeai-platform/issues/671, which I think @kapetr is working on, so I left just the stub component there.

PR #734, opened 2025-06-03
docs: Deployment with Helm

jenna-winkler:


jenna-winkler:
@jezekra1 @tomkis I tested deploying my own instance of the platform using Helm and was successful. please check my accuracy in this documentation 🙏

docs/deployment/kubernetes.mdx
```diff
@@ -1,5 +1,189 @@
...
+description: "Deploy BeeAI on a Kubernetes cluster using Helm charts"
 ---
 
+This guide walks you through deploying the BeeAI Platform to Kubernetes using Helm. You'll configure agents and LLM providers, then deploy everything with a single command.
+
+## Quick Start
+
+The fastest way to deploy BeeAI with default agents and Ollama:
+
+### Step 1: Get the Helm Chart
```
tomkis: Let's remove the cloning part. because you can literally.

`helm install beeai oci://ghcr.io/i-am-bee/beeai-platform/beeai-platform-chart/beeai-platform:0.2.0 -f agents.yaml -f llm-config.yaml -f secret.yaml`

Where the oci image is in packages on github.

docs/deployment/kubernetes.mdx
```diff
@@ -1,5 +1,189 @@
...
+# Clone the BeeAI repository to get the Helm chart
+git clone https://github.com/i-am-bee/beeai-platform.git
+cd beeai-platform/helm
+```
+
+### Step 2: Install Dependencies
+
+```bash
+# Update Helm chart dependencies
+helm dependency build beeai-platform
```
tomkis: this is not needed cc @jezekra1

docs/deployment/kubernetes.mdx
```diff
@@ -1,5 +1,189 @@
...
+```
+
+### Step 3: Deploy with Default Configuration
+
+```bash
+# Deploy with Ollama configuration
+helm install my-beeai-platform ./beeai-platform \
+  -f ../agent-registry.yaml \
+  --set variables.LLM_API_BASE=http://host.docker.internal:11434/v1 \
```
tomkis: This will be hard to maintain, @jezekra1 i am thinking of making sure this is installable with some defaults? maybe we provide some .env example?

docs/deployment/kubernetes.mdx
```diff
@@ -1,5 +1,198 @@
...
+```shell
+beeai list
+beeai run chat hi
+```
+
+## Configuration Options
+
+### Security Settings
+
+For production deployments, ensure proper security:
```
jezekra1: nitpick: I'm not that confident in the security yet without proper user and auth methods, so not sure if I'd use this phrasing

docs/deployment/kubernetes.mdx
```diff
@@ -1,5 +1,198 @@
...
+kubectl get events --sort-by=.lastTimestamp
+```
+
+**Can't access agents:**
+```bash
+# Verify LLM configuration
+kubectl logs deployment/beeai-platform | grep -i llm
+
+# Test LLM connectivity from within cluster
+kubectl run test-curl --image=curlimages/curl -it --rm --restart=Never -- \
```
jezekra1: this is not necessary, it only tests that the cluster has access to the internet, which is almost always the case :slightly_smiling_face: You probably saw this when debugging [this issue](https://github.com/i-am-bee/beeai-platform/discussions/742), but there I tested host.docker.internal to check whether the cluster has access to host, that's only applicable for local installation

docs/deployment/openshift.mdx
```diff
@@ -3,3 +3,422 @@ title: "OpenShift"
...
+
+```bash
+helm install -f openshift-config.yaml beeai \
+  oci://ghcr.io/i-am-bee/beeai-platform/beeai-platform-chart/beeai-platform:latest \
+  -n beeai-platform
+```
+
+### Step 6: Create OpenShift Route
+
+Create a route to expose the BeeAI platform:
```
jezekra1: Here we're showing how to create a route for openshift, but for kubernetes deployment it's not clear how you can access it. We don't have an ingress in the helm chart yet (it belongs there) but I'd at least mention that you need to create an ingress pointed to the beeai-platform-svc service

jezekra1 (changes requested):
I don't want to promote any "production" public deployments because we really didn't have the time to stop yet and think about the consequences.
We don't have users or ratelimits yet and the /chat/completions endpoint is wide open, that's like free LLM for everyone if it's available publicly.

I haven't tried openshift so it would be nice if someone can replicate it and test that it's working :)

jezekra1:
> Would it make sense to include this warning in the deployment overview page? @jezekra1

Yes, I hope we'll secure this soon :)

jezekra1 (approved):

PR #754, opened 2025-06-05
feat(ui): handle agent 'starting' state 

### LINKED ISSUE #750: Handle the agent 'starting' state when launched for the first time in the UI
On the first run, the agent is downloaded and added to the platform, which can take a non-trivial amount of time. We should track its provider status and show appropriate visual feedback in the UI.

This is a work-in-progress proposal where the handoff view simply replaces the loading message. It would be great if you could review this and could you also please prepare a suitable solution for the chat view @anafucs ? 
(we should probably start counting the elapsed time once the agent is started)
 
https://github.com/user-attachments/assets/3d89880e-b96b-4af5-9a4f-7c689905f681

kapetr:
Closes #750

Implements visual feedback for the agent `starting` state in the hand-off view and adds the necessary context for supporting it in other views.

To be done:
- Get revision of copy (and possibly visual) for the hand-off view
- Get design for the chat view
- Implement the starting state in the chat view


apps/beeai-ui/src/modules/providers/api/index.ts
```diff
@@ -27,6 +27,12 @@ export async function listProviders() {
   return ensureData(response);
 }
 
+export async function readProvider(id: string) {
```
penge, line 30: I'd go with `getProvider`, for consistency with the `GET` method.

penge (approved):

apps/beeai-ui/src/modules/providers/api/index.ts
PetrBulanek: By some previous convention, we prefix all GET queries to a single entity as read* (readAgent, readRun, ...) and to a multiple entities as list* (listAgents, listProviders, ...), so changing to get* would actually create an inconsistency here.

PR #756, opened 2025-06-05
feat(ui): add base components for sources

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/652


apps/beeai-server/src/beeai_server/api/routes/ui.py
```diff
@@ -23,3 +28,62 @@
...
+    title: str | None = None
+    description: str | None = None
+    favicon_url: str | None = None
+
+
+# We plan to switch from Vite to Next.js for beeai-ui (https://github.com/i-am-bee/beeai-platform/pull/760), so then we can move this endpoint there.
+@router.post("/source", response_model=SourceMetadata)
+def get_source_metadata(url: str):
+    try:
+        response = requests.get(url, timeout=10)
```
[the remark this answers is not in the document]
tomkis: @jezekra1 can we proxy ACP and if it contains source then MitM and fetch the  preview before proxying back to the client?

apps/beeai-server/src/beeai_server/api/routes/ui.py
```diff
@@ -23,3 +28,62 @@
...
+    title: str | None = None
+    description: str | None = None
+    favicon_url: str | None = None
+
+
+# We plan to switch from Vite to Next.js for beeai-ui (https://github.com/i-am-bee/beeai-platform/pull/760), so then we can move this endpoint there.
+@router.post("/source")
+def get_source_metadata(url: str) -> SourceMetadata:
+    try:
+        response = requests.get(url, timeout=10)
```
jezekra1: let's use async fastapi handler with `httpx` instead of requests

apps/beeai-server/src/beeai_server/api/routes/ui.py
```diff
@@ -23,3 +28,62 @@
...
+
+class SourceMetadata(BaseModel):
+    title: str | None = None
+    description: str | None = None
+    favicon_url: str | None = None
+
+
+# We plan to switch from Vite to Next.js for beeai-ui (https://github.com/i-am-bee/beeai-platform/pull/760), so then we can move this endpoint there.
+@router.post("/source")
+def get_source_metadata(url: str) -> SourceMetadata:
```
jezekra1: you can use `HttpUrl` here instead of `urlparse`, this will parse the url and validate that the scheme is `http` or `https`

apps/beeai-server/src/beeai_server/api/routes/ui.py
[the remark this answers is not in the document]
jezekra1: I don't think it's a good idea to keep adding features to the proxy. You'll have a different experience when using agents through the platform vs. using them directly which is something I'd like to avoid.
This is a UI feature not ACP feature

Also it would hold the requests for up to 10s even if you don't care about the previews.
Previews should be asynchronous - render the response and then show spinner on sources and load icons asynchronously

jezekra1 (dismissed):

apps/beeai-server/src/beeai_server/api/routes/ui.py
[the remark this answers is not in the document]
tomkis: It doesn't necessary has to be part of the proxy itself. We can trigger the call asynchronously (on demand by UI) but ideally there's a way to download the content through the server in a transaparent way (eg. based on run_id) which kinda means "proxying the ACP somehow"

apps/beeai-server/src/beeai_server/api/routes/ui.py
```diff
@@ -23,3 +28,62 @@
...
+    origin = f"{parsed_url.scheme}://{parsed_url.netloc}"
+
+    if favicon_path.startswith("http"):
+        return favicon_path
+    elif favicon_path.startswith("//"):
+        return f"{parsed_url.scheme}:{favicon_path}"
+    elif favicon_path.startswith("/"):
+        return origin + favicon_path
+    else:
+        return f"{origin}/{favicon_path}"
```
penge:
```python
parsed_url = urlparse(url)
origin = f"{parsed_url.scheme}://{parsed_url.netloc}/"

return urljoin(origin, favicon_path)
```

?

PetrBulanek:
After discussion, I have for the time being removed the entire endpoint for scraping information from the url for security reasons.

tomkis (approved):

apps/beeai-ui/src/modules/runs/sources/contexts/SourcesProvider.tsx
```diff
@@ -0,0 +1,52 @@
...
+import type { ActiveMessageKey, ActiveSourceKey } from './types';
+
+interface Props {
+  sourcesData: SourcesData;
+}
+
+export function SourcesProvider({ sourcesData, children }: PropsWithChildren<Props>) {
+  const { sourcesPanelOpen } = useApp();
+  const [activeMessageKey, setActiveMessageKey] = useState<ActiveMessageKey>(null);
+  const [activeSourceKey, setActiveSourceKey] = useState<ActiveSourceKey>(null);
```
kapetr, line 32: I’d consider making this just one state object variable to avoid ending up in an inconsistent state.

apps/beeai-ui/src/contexts/App/AppProvider.tsx
```diff
@@ -15,25 +15,57 @@
...
+
+  const hideAgentDetail = useCallback(() => {
+    setAgentDetailOpen(false);
+  }, []);
+
+  const showSourcesPanel = useCallback(() => {
+    setSourcesPanelOpen(true);
+    hideAgentDetail();
+  }, [hideAgentDetail]);
+
```
kapetr, line 45: If we ever need to add another right panel, we might want to consider a more generic way to manage their open states.

apps/beeai-ui/src/modules/runs/contexts/chat/ChatProvider.tsx
```diff
@@ -114,6 +135,37 @@ export function ChatProvider({ agent, children }: PropsWithChildren<Props>) {
...
+              citationTransformGroup.sources.push(newSource);
+            } else {
+              message.contentTransforms.push(createCitationTransform({ source: newSource }));
+            }
+          });
+
+          break;
+
+        default:
+          break;
```
kapetr, line 163: This looks redundant to me.

PR #760, opened 2025-06-05
feat(ui): switch to nextjs

kapetr:


apps/beeai-ui/tasks.toml
```diff
@@ -20,7 +20,7 @@ outputs = { auto = true }
 ["beeai-ui:check:tsc"]
 depends = ["setup:pnpm"]
 dir = "{{config_root}}/apps/beeai-ui"
-run = "pnpm tsc --noEmit -p tsconfig.app.json"
+run = "pnpm tsc --noEmit -p tsconfig.json"
```
kapetr, line 23: @tomkis I’m not sure what the reasoning was behind splitting the `tsconfig` into two separate files, and whether it’s still relevant after the migration to Next.js. For now, I've used a single `tsconfig.json`.

apps/beeai-ui/src/styles/_breakpoints.scss
PetrBulanek, line 1: Since we have to copy it manually, I guess we could get rid of the map.merge below and have it all defined in one place to make it clearer?

apps/beeai-ui/src/styles/components/_button.scss
```diff
@@ -5,6 +5,8 @@
 
 @use 'styles/common' as *;
 
+@use '@carbon/styles/scss/components/button';
```
PetrBulanek: Why is this needed?

apps/beeai-ui/src/App.tsx
```diff
@@ -1,58 +0,0 @@
...
-import { Agent } from '#pages/agents/Agent.tsx';
-import { Landing } from '#pages/Landing.tsx';
-import { NotFound } from '#pages/NotFound.tsx';
-import { AgentRunPage } from '#pages/runs/AgentRunPage.tsx';
-import { Settings } from '#pages/Settings.tsx';
-import { routeDefinitions } from '#utils/router.ts';
-
-export function App() {
-  return (
-    <ErrorBoundary FallbackComponent={ErrorFallback}>
```
PetrBulanek, line 29: If we don't use the `ErrorBoundary` anymore, we can probably get rid of the `ErrorFallback` and `ErrorLayout` components.

apps/beeai-ui/src/components/TransitionLink/TransitionLink.tsx
```diff
@@ -3,34 +3,30 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Slot } from '@radix-ui/react-slot';
-import type { HTMLProps } from 'react';
+import type { LinkProps } from 'next/link';
+import Link from 'next/link';
```
PetrBulanek: Can be merged into one import.

apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
```diff
@@ -3,23 +3,24 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { ArrowUpRight } from '@carbon/icons-react';
+'use client';
+
 import { SkeletonText, Tab, TabList, TabPanel, TabPanels, Tabs } from '@carbon/react';
-import { useParams } from 'react-router';
+import { ArrowUpRight } from '@carbon/react/icons';
```
PetrBulanek: Probably not any issue, but all other icon imports are from `@carbon/icons-react`.

apps/beeai-ui/src/modules/agents/components/AgentDetailPanel.tsx
```diff
@@ -3,23 +3,24 @@
...
 
 import { MarkdownContent } from '#components/MarkdownContent/MarkdownContent.tsx';
 import { SidePanel } from '#components/SidePanel/SidePanel.tsx';
 import { useApp } from '#contexts/App/index.ts';
+import { useAgentNameFromPath } from '#hooks/useAgentNameFromPath.ts';
 
 import { useAgent } from '../api/queries/useAgent';
-import type { AgentPageParams } from '../types';
```
PetrBulanek, line 16: Seems we can also remove the whole file as it is not used anywhere now.

apps/beeai-ui/src/app/agents/[name]/page.tsx
Comment (PetrBulanek) [inline apps/beeai-ui/src/app/agents/[name]/page.tsx:1]: I think we can fetch the agent on server here and pass it into `AgentRun` to render it statically. Farther, we could remove the isPending branch and show 404 (maybe with some special description) if the agent is not present.

apps/beeai-ui/src/modules/agents/list/AgentsView.tsx
PetrBulanek, line 1: We don't use this component at all. Probably delete it?

apps/beeai-ui/.env.example
```diff
@@ -1,13 +1,16 @@
...
-VITE_API_SERVER_TARGET=
+NEXT_PUBLIC_APP_FAVICON_SVG=
 
 # Default: 'http://localhost:6006'
-VITE_PHOENIX_SERVER_TARGET=
+NEXT_PUBLIC_PHOENIX_SERVER_TARGET='http://localhost:6006'
+
+# Custom navigation structure in JSON format 
+# See ./src/modules/nav/schema.ts for the schema definition
+NEXT_PUBLIC_NAV_ITEMS_JSON=
```
PetrBulanek: `NEXT_PUBLIC_NAV_ITEMS_JSON` is not used anywhere. There is only `process.env.NEXT_PUBLIC_NAV_ITEMS`, so I guess it is a naming mismatch.

apps/beeai-ui/package.json
```diff
@@ -5,11 +5,11 @@
   "version": "0.1.0",
   "type": "module",
   "scripts": {
-    "dev": "vite",
-    "build": "tsc -b && vite build",
-    "check": "prettier --log-level silent --check src && eslint src && tsc --noEmit -p tsconfig.app.json && stylelint src/**/*.css src/**/*.scss",
+    "dev": "NODE_OPTIONS=\"--no-experimental-global-navigator\" next dev",
+    "build": "NODE_OPTIONS=\"--no-experimental-global-navigator\" next build",
+    "preview": "pnpm build && NODE_OPTIONS=\"--no-experimental-global-navigator\" next start",
```
PetrBulanek: As we now use Next.js I think it is good idea to rename it back to `start` (I'd do it also in the beeai-web app).

apps/beeai-ui/src/app/layout.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export default function RootLayout({
+  children,
+}: Readonly<{
+  children: React.ReactNode;
+}>) {
+  return (
+    <html lang="en" suppressHydrationWarning>
+      <head>
+        <link rel="icon" type="image/svg+xml" href="/bee.svg" />
```
PetrBulanek, line 50: I think this is redundant to the icon declaration in metadata export. I would do the same for beeai-web.

apps/beeai-ui/src/app/page.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { listAgents } from '#modules/agents/api/index.ts';
+import { sortAgentsByName } from '#modules/agents/utils.ts';
+import { routes } from '#utils/router.ts';
+
+export default async function LandingPage() {
+  let firstAgentName;
+
+  try {
+    const response = await listAgents();
+    const agents = response?.agents.sort(sortAgentsByName);
```
PetrBulanek: We should filter out agents without supported ui.

apps/beeai-ui/src/components/ExpandButton/ExpandButton.module.scss
PetrBulanek, line 1: I got this stylelint error:

```Unexpected duplicate selector "&", first used at line 25 (no-duplicate-selectors)Stylelint[no-duplicate-selectors](https://stylelint.io/user-guide/rules/no-duplicate-selectors)```

File: apps/beeai-ui/src/components/MarkdownContent/components/CitationLink/InlineCitationTooltipContent.tsx
Comment (PetrBulanek) [inline apps/beeai-ui/src/components/MarkdownContent/components/CitationLink/InlineCitationTooltipContent.tsx:1]: The `Source` and `SourceButton` components use next/image. Is there any difference?

File: apps/beeai-ui/src/components/MarkdownContent/components/Img.tsx [hunk 11]
```diff
@@ -13,7 +13,8 @@ import classes from './Img.module.scss';
 export function Img({ node, className, ...props }: ImgHTMLAttributes<HTMLImageElement> & ExtraProps) {
   return (
     <span className={classes.root}>
-      <img {...props} className={clsx(classes.img, className)} />
+      {/* eslint-disable-next-line @next/next/no-img-element */}
+      <img {...props} className={clsx(classes.img, className)} alt={props.alt ?? 'Markdown image'} />
```
PetrBulanek: I don't think the "Markdown image" has any value. I'd leave it with blank string or suppress the rule here.

apps/beeai-ui/src/components/layouts/Navigation.tsx
```diff
@@ -45,7 +45,7 @@ export function Navigation() {
 
   return (
     <MainNav
-      linkComponent={TransitionLink}
+      linkComponent={TransitionLink as ComponentType}
```
PetrBulanek: Is the typecasting needed? I don't have any warning/error once omitted.

apps/beeai-ui/src/index.ts
```diff
@@ -11,15 +11,21 @@ export * from './components/layouts/Container';
 export type * from './components/MainContentView/MainContentView';
 export * from './components/MainContentView/MainContentView';
 export * from './components/MainNav/MainNav';
+export * from './components/MainNav/TracesTooltip';
```
PetrBulanek: I don't see `TracesTooltip`,  `usePhoenix ` and `router` being use in beeai-web app. Why re-exporting?

apps/beeai-ui/src/modules/compose/ComposeLanding.module.scss
PetrBulanek, line 1: Any shenanigans with CSS specificity, or why this change?

apps/beeai-ui/src/modules/compose/components/AgentListOption.tsx
```diff
@@ -23,7 +23,7 @@ export function AgentListOption({ agent, onClick }: Props) {
   const { display_name } = getAgentUiMetadata(agent);
 
   return (
-    <li className={classes.root} role="option" onClick={onClick}>
+    <li className={classes.root} role="option" aria-selected onClick={onClick}>
```
PetrBulanek: Doesn't this apply to all items now?

apps/beeai-ui/src/modules/settings/SettingsView.tsx
PetrBulanek, line 1: Are those spaces necessary? :-)

apps/beeai-ui/src/pages/agents/Agent.tsx
PetrBulanek, line 1: I would go through the components that were only showing up in detail and consider removing them, or moving them directly to beeai-web if some are still in use there.

apps/beeai-ui/src/hooks/useUpdateSearchParams.ts
```diff
@@ -0,0 +1,30 @@
...
+export function useUpdateSearchParams() {
+  const router = useRouter();
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+
+  const updateSearchParams = useCallback(
+    (params: Record<string, string | number | boolean>) => {
+      const newSearchParams = new URLSearchParams(searchParams ?? undefined);
+      Object.entries(params).forEach(([key, value]) => {
+        if (value !== undefined && value !== null) {
```
PetrBulanek: Same as `value != null`, or better use the `isNotNull(value)` helper.

apps/beeai-ui/src/pages/api/[...path].ts
Comment (PetrBulanek) [inline apps/beeai-ui/src/pages/api/[...path].ts:1]: Later on we can review this, not sure we need all of this code. This file is a bit cryptic to me. Not needed to do it now. :-)

apps/beeai-ui/next.config.ts
PetrBulanek, line 1: A lot of this is the same as in the beeai-web config. We could go through it later and reuse those common things to make it more manageable.

apps/beeai-ui/eslint.config.mjs
```diff
@@ -0,0 +1,23 @@
+import { FlatCompat } from '@eslint/eslintrc';
```
PetrBulanek, line 6: My eslint server is crashing with couple of errors. At least I noticed `@eslint/eslintrc` is missing from package.json.

And some now not used packages are still in pacakge.json (eslint-plugin-react-refresh, eslint-plugin-react-hooks).

Have you tried to lint all the files to check everything is still the same?

apps/beeai-ui/src/hooks/useAgentNameFromPath.ts
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { usePathname } from 'next/navigation';
+
+export function useAgentNameFromPath() {
+  const pathname = usePathname();
+  const parts = pathname?.split('/');
+
+  if (parts?.at(1) === 'agents') {
```
penge: Use `useParams` instead.

apps/beeai-ui/src/hooks/useIsNavSectionActive.ts
```diff
@@ -3,14 +3,14 @@
...
 import { useCallback } from 'react';
-import { useLocation } from 'react-router';
 
 import type { NavSectionName } from '#utils/router.ts';
 
 export function useIsNavSectionActive() {
-  const { pathname } = useLocation();
-  const section = pathname.split('/').at(1);
+  const pathname = usePathname();
+  const section = pathname?.split('/').at(1);
```
penge, line 13: This will break if routing changes. Let's use nextjs suggested ways.

apps/beeai-ui/src/hooks/useUpdateSearchParams.ts
penge: Why not just inverting the condition?

`if (value)`

apps/beeai-ui/src/modules/phoenix/api/queries/usePhoenix.ts
```diff
@@ -15,7 +15,7 @@ export function usePhoenix() {
     refetchInterval: PROD_MODE ? 5_000 : false,
     enabled: Boolean(PHOENIX_SERVER_TARGET),
     queryFn: () =>
-      fetch(PHOENIX_SERVER_TARGET, { mode: 'no-cors' })
+      fetch(PHOENIX_SERVER_TARGET!, { mode: 'no-cors' })
```
penge: Let's avoid `!`

apps/beeai-ui/src/utils/commands.ts
```diff
@@ -27,3 +27,5 @@ export default {
     },
   },
 };
+
+export default commands;
```
penge, line 31: Why using this style since there is no type attached to the variable? (just like before)

apps/beeai-ui/src/utils/constants.ts
```diff
@@ -3,6 +3,20 @@
...
+
+export const APP_FAVICON_SVG = process.env.NEXT_PUBLIC_APP_FAVICON_SVG ?? '/bee.svg';
+
+export const PHOENIX_SERVER_TARGET = process.env.NEXT_PUBLIC_PHOENIX_SERVER_TARGET ?? 'http://localhost:6006';
+
+export const NAV_ITEMS = parseNav(process.env.NEXT_PUBLIC_NAV_ITEMS);
+
+export const API_URL = process.env.API_URL;
+
+export const PROD_MODE = process.env.PROD;
```
penge: `IS_PROD`

apps/beeai-ui/src/styles/components/_button.scss
kapetr: I guess it was just some debug leftover.

apps/beeai-ui/src/pages/api/[...path].ts
Comment (kapetr) [reply apps/beeai-ui/src/pages/api/[...path].ts:1]: Agree, that's a copy/paste from `bee-ui` stripped of unnecessary stuff.

apps/beeai-ui/src/components/MarkdownContent/components/CitationLink/InlineCitationTooltipContent.tsx
kapetr: In the end I think we'll have to use `img` in all the files, since these will be images from external urls.

apps/beeai-ui/src/components/ExpandButton/ExpandButton.module.scss
kapetr: I don't get this error, do you @penge ?

apps/beeai-ui/src/components/layouts/Navigation.tsx
kapetr: It wasn't. The weird thing is, it was needed in `beeai-web` in the identical usage :) Anyway, the `linkComponent` prop is not needed anymore, since both apps use the same component now, so I'm removing it.

apps/beeai-ui/src/index.ts
kapetr: I merged the `Navigation` component, which was using that, from `beeai-web` to `beeai-ui` and this was left behind.

apps/beeai-ui/src/components/ExpandButton/ExpandButton.module.scss
penge: I can check.

apps/beeai-ui/src/modules/compose/ComposeLanding.module.scss
kapetr: No, I was just getting errors that these ARIA attributes aren’t allowed on the li element. Looking at it now, from an accessibility perspective, it would be better to use proper element/role.

kapetr: It seems that setting `role='option'` was all that was needed.

apps/beeai-ui/src/modules/compose/components/AgentListOption.tsx
kapetr: Ah, good catch, this was left unfinished.

kapetr: `role='option'` was actually wrongly used here, removing it altogether.

apps/beeai-ui/src/modules/settings/SettingsView.tsx
kapetr: No idea how they got there :)

apps/beeai-ui/src/hooks/useUpdateSearchParams.ts
kapetr: Unchecked code from AI 🙈

apps/beeai-ui/next.config.ts
kapetr: Let's do that later on.

apps/beeai-ui/src/utils/commands.ts
kapetr: [[no-anonymous-default-export]](https://github.com/import-js/eslint-plugin-import/blob/v2.31.0/docs/rules/no-anonymous-default-export.md)

apps/beeai-ui/src/utils/constants.ts
kapetr: Neither is set, I'm changing it to `process.env.NODE_ENV === 'production'`.

apps/beeai-ui/src/hooks/useIsNavSectionActive.ts
kapetr: Why would it break?

apps/beeai-ui/eslint.config.mjs
kapetr: It's not crashing for me. But I removed the mentioned eslint plugins and added `@eslint/eslintrc`. It was probably not needed, since it was already added from `beeai-web` to the catalogue.

Could you check this as well @penge?🙏

apps/beeai-ui/src/hooks/useIsNavSectionActive.ts
penge: Because it relies on specific file structure.

kapetr: That's irrelevant, look at the usage

penge: Then write `pathname?.split('/').at(2);` and it should work :D

kapetr: Why would I do that?

jezekra1 (changes requested):
The following need to be updated:
[ ] mise tasks for UI build
[ ] remove UI logic and endpoint from platform server
[ ] `beeai ui` now opens localhost:8333, it should be a different port when served separately
[ ] build pipeline (add docker image for next)
[ ] helm chart - add UI deployment
[ ] feature flags configuration in helm - not sure how it's propagated now

apps/beeai-ui/src/hooks/useIsNavSectionActive.ts
penge: Because it's irrelevant :D

apps/beeai-ui/.env.example
```diff
@@ -1,13 +1,16 @@
+API_URL='http://127.0.0.1:8333'
```
jezekra1, line 1: don't forget to set this to `http://beeai-platform-svc:8333` in helm

PR #762, opened 2025-06-06
feat: added sidebar variant

jayolee:
### Fixes
- fixed agent nav's title aligned to the end on Settings page (/components/AppHeader/AppHeader.module.scss)

### Updates
- added sidebar variant env variable and description to the README
- added toggle-below-header sidebar. Will share the design in a separate channel.

### Notes to the reviewers
- To view the new variant, add `VITE_APP_SIDEBAR_VARIANT=toggle-below-header` to `.env`
- For `toggle-below-header`, I moved the agent nav control into the `AgentHeader`, which I think is not an ideal solution. The main reason is the position of the "New Session" button: which now is relative to the length of the agent name. I felt this approach is cleaner than others, but I'm open to suggestions.
- The nav is always in the header on Settings page.

### DIFF
--- a/apps/beeai-ui/README.md
+++ b/apps/beeai-ui/README.md
@@ -36,6 +36,17 @@ Example:
 
 For more details, see the [schema](./src/modules/nav/schema.ts).
 
+
 ## ENV
 
 See [`.env.example`](./.env.example).
+
+
+## Sidebar toggle position
+
+By default, the hamburger menu used to toggle the sidebar is located in the header.
+Its position can be configured using `VITE_APP_SIDEBAR_VARIANT` variable in `.env` file.
+
+Two options are currently supported.
+- `toggle-in-header` (default): Displays the hamburger menu inside the header, next to the app name.
+- `toggle-below-header`: Displays the hamburger menu below the header with the agent name. The new session button appears next to the menu.
--- a/apps/beeai-ui/src/components/AgentsNav/AgentsNav.module.scss
+++ b/apps/beeai-ui/src/components/AgentsNav/AgentsNav.module.scss
@@ -20,6 +20,15 @@
   row-gap: $spacing-04;
   padding-inline: $spacing-05;
   @include scrollbar();
+
+  &.toggleBelowHeader {
+    padding-block-start: rem(48px);
+
+    .heading {
+      border-block-start: rem(1px) solid $border-subtle-00;
+      padding-block-start: rem(16px);
+    }
+  }
 }
 
 .heading {
@@ -52,14 +61,5 @@
 
   &.isActive {
     background-color: $layer-02;
-  }
-}
-
-.root.toggleBelowHeader {
-  padding-block-start: rem(48px);
-
-  .heading {
-    border-block-start: rem(1px) solid $border-subtle-00;
-    padding-block-start: rem(16px);
   }
 }
--- a/apps/beeai-ui/src/components/AgentsNav/AgentsNav.module.scss
+++ b/apps/beeai-ui/src/components/AgentsNav/AgentsNav.module.scss
@@ -54,3 +54,12 @@
     background-color: $layer-02;
   }
 }
+
+.root.toggleBelowHeader {
+  padding-block-start: rem(48px);
+
+  .heading {
+    border-block-start: rem(1px) solid $border-subtle-00;
+    padding-block-start: rem(16px);
+  }
+}
--- a/apps/beeai-ui/src/components/AppHeader/AppHeader.tsx
+++ b/apps/beeai-ui/src/components/AppHeader/AppHeader.tsx
@@ -21,12 +21,14 @@ import { MainNav } from '#components/layouts/MainNav.tsx';
 import { useAgent } from '#modules/agents/api/queries/useAgent.ts';
 import type { AgentPageParams } from '#modules/agents/types.ts';
 import { getAgentDisplayName } from '#modules/agents/utils.ts';
-import { APP_NAME, NAV, SIDEBAR_VARIANT } from '#utils/vite-constants.ts';
+import { isSidebarToggleBelowHeader } from '#modules/sidebar/isSidebarToggleBelowHeader.ts';
+import { NAV } from '#utils/vite-constants.ts';
 
 import { Container } from '../layouts/Container';
 import { AgentDetailButton } from './AgentDetailButton';
 import classes from './AppHeader.module.scss';
 import { AppHeaderNav } from './AppHeaderNav';
+import { AppName } from './AppName';
 
 interface Props {
   className?: string;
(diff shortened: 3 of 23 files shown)


apps/beeai-ui/README.md
```diff
@@ -36,6 +36,17 @@ Example:
 
 For more details, see the [schema](./src/modules/nav/schema.ts).
 
+
 ## ENV
 
 See [`.env.example`](./.env.example).
+
+
+## Sidebar toggle position
```
penge, line 45: It would look more natural to place this after "Custom navigation". Suggested final order:
- Custom navigation
- Sidebar toggle position
- ENV

apps/beeai-ui/src/components/AgentsNav/AgentsNav.module.scss
```diff
@@ -54,3 +54,12 @@
...
 }
+
+.root.toggleBelowHeader {
+  padding-block-start: rem(48px);
+
+  .heading {
+    border-block-start: rem(1px) solid $border-subtle-00;
+    padding-block-start: rem(16px);
+  }
+}
```
penge: It is better to co-locate with base styles:

```scss
.root {
  // ...

  &.toggleBelowHeader {
    padding-block-start: rem(48px);
  }
}

.heading {
  // ...

  .toggleBelowHeader & {
    border-block-start: rem(1px) solid $border-subtle-00;
    padding-block-start: rem(16px);
  }
}
```

apps/beeai-ui/src/components/AppHeader/AppHeader.tsx
```diff
@@ -35,18 +35,21 @@ interface Props {
...
   const { agentName } = useParams<AgentPageParams>();
   const { data: agent } = useAgent({ name: agentName ?? '' });
+  const toggleBelowHeader = SIDEBAR_VARIANT === 'toggle-below-header';
 
   return (
     <header className={clsx(classes.root, className)}>
       <Container size="full">
         <div className={clsx(classes.holder, { [classes.hasNav]: NAV.length > 0 })}>
-          <MainNav />
+          {toggleBelowHeader && agent ? <span className={classes.appName}>{APP_NAME}</span> : <MainNav />}
```
penge: Introducing an `<AppName>` component could improve organization and house the styles for `.appName`.

apps/beeai-ui/src/components/AppHeader/SidebarButton.tsx
```diff
@@ -42,3 +47,29 @@ export function SidebarButton() {
...
   );
 }
+
+export function SidebarButtonWAgentName() {
+  const { agentName } = useParams<AgentPageParams>();
+  const { data: agent } = useAgent({ name: agentName ?? '' });
+  const { setNavigationOpen, navigationOpen } = useApp();
+
+  return agent ? (
+    <div className={clsx(classes.root, classes.belowHeader)}>
```
penge: There's nothing in this component that suggests the use of `.belowHeader`. A better name might be `.wAgentName`?

apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
```diff
@@ -26,7 +26,10 @@
 .holder {
   display: grid;
   grid-template-rows: max-content 1fr max-content;
-  row-gap: $spacing-06;
+
+  > *:not(:last-child) {
+    margin-block-end: $spacing-06;
+  }
```
penge, line 28: How is this style **_within_** Chat related to the sidebar variant? Should there be different spacings for different sidebar variants?

apps/beeai-ui/src/modules/runs/chat/Chat.tsx
```diff
@@ -79,14 +79,13 @@ export function Chat() {
...
-        {isNew ? (
+        {isNew && (
           <div className={classes.header}>
             <AgentIcon size="xl" />
             <AgentGreeting agent={agent} />
           </div>
-        ) : (
-          <AgentHeader className={classes.header} onNewSessionClick={onClear} />
         )}
+        <AgentHeader className={classes.header} onNewSessionClick={onClear} hideButton={isNew} />
```
penge, line 88: I find it confusing that `<AgentHeader>` is used to render **fixed** `<NewSessionButton>` outside the `<Chat>` (and also that `<NewSessionButton>` has the responsibility to render `<MainNav>`).

Maybe `<AgentHeader>` could be renamed to `<ChatHeader>` and be displayed only when the chat is **not** new, and only **_within_** the Chat. For the other scenario, we could use `<ChatHeader fixed>`. Thoughts?

apps/beeai-ui/src/modules/runs/components/NewSessionButtton.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  if (fixed) return <FixedButton>{Button}</FixedButton>;
+
+  return Button;
+}
+
+function FixedButton({ children }: PropsWithChildren) {
+  const { navigationOpen } = useApp();
+  if (!(window && document)) return null;
+  return createPortal(
```
penge, line 59: Should work just fine without the portal, since the position is **fixed**. Is there some specific reason to use portal?

apps/beeai-ui/src/utils/vite-constants.ts
```diff
@@ -23,3 +23,5 @@ export const NAV = parseNav(__NAV__);
 export const PHOENIX_SERVER_TARGET = __PHOENIX_SERVER_TARGET__;
 
 export const PROD_MODE = import.meta.env.PROD;
+
+export const SIDEBAR_VARIANT = __SIDEBAR_VARIANT__;
```
penge, line 27: Let's have `SIDEBAR_VARIANT` strongly typed. Here's the schema we could use:

```ts
const sidebarVariantSchema = z.enum(['toggle-in-header', 'toggle-below-header']);
```

apps/beeai-ui/vite.config.ts
```diff
@@ -10,12 +10,13 @@ import { loadFile } from './src/utils/files/loadFile';
 const DEFAULT_ENV = {
   VITE_APP_NAME: 'BeeAI',
   VITE_APP_FAVICON_SVG: '/bee.svg',
+  VITE_APP_SIDEBAR_VARIANT: 'toggle-in-header',
```
penge: Deserves to be strongly typed.

penge (changes requested):

apps/beeai-ui/src/modules/runs/components/NewSessionButtton.tsx
jayolee: The content outside the chat's scrollable area is hidden.

apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
jayolee: Because the layout uses grid template, I left an empty div when the header is fixed to not break the template and also to avoid introducing different grid template per scenario. 
This adds the top margin to the chat area which is the row gap of this template when the nav is `position: fixed`. By changing it to margin, I have control over each spacing - means I can remove the top gap.

apps/beeai-ui/README.md
jayolee: I've put it after ENV because sidebar toggle position is defined by ENV - Do you think it's still better to put it before ENV?

penge: Oh, I see. I was concerned about not having those together:
- Custom navigation
- Sidebar toggle position

As they both are some customization things.

Placing ENV before makes sense. Thanks for explaining.

Maybe we could put ENV as the very first. Mmm? But, marking as resolved anyways.

apps/beeai-ui/src/modules/runs/chat/Chat.tsx
jayolee: Could you elaborate more on "For the other scenario, we could use <ChatHeader fixed>"?
FYI this component is also used for HandOff UI

apps/beeai-ui/README.md
jayolee: Yeah I also thought about it hmm
I think either way it make sense. Custom navigation does not have anything related to env, and right now ENV section is short so reading from custom nav -> env -> and sidebar also makes sense.

penge: I like the reasoning. Let's keep it that way.

PR #766, opened 2025-06-09
fix(ui): alignment in header

PetrBulanek:


apps/beeai-ui/src/components/AppHeader/AppHeader.tsx
```diff
@@ -43,10 +43,14 @@ export function AppHeader({ className }: Props) {
           <MainNav />
 
           {NAV.length > 0 && <AppHeaderNav items={NAV} />}
+
           {!NAV.length && agent && (
             <>
               <p className={classes.agentName}>{getAgentDisplayName(agent)}</p>
-              <AgentDetailButton />
+
+              <div className={classes.agentDetailButton}>
```
penge: The name `.agentDetailButton` suggests it refers to the **button** itself, not a wrapper or container. A better name might be `.agentDetailButtonContainer`.

penge (approved):

apps/beeai-ui/src/modules/runs/chat/Chat.module.scss
```diff
@@ -17,16 +17,15 @@
...
-  &.isNew {
-    align-items: center;
-  }
 }
 
 .holder {
   display: grid;
   grid-template-rows: max-content 1fr max-content;
   row-gap: $spacing-06;
+  .isNew & {
```
penge, line 26: Doesn't linter complain about a missing empty line above?

PetrBulanek: It does not. I don't think we have any rules for that. :)

penge (approved):

PR #775, opened 2025-06-10
docs: Update kubernetes.mdx with ArgoCD note

jenna-winkler:


JanPokorny:
This is actually not accurate, ArgoCD has no problem using YAML values, given that you pass them in properly -- see https://argo-cd.readthedocs.io/en/latest/user-guide/helm/#values-files

IMO ArgoCD is out of scope of our docs, the only change I would make here is rename `config.yaml` to a more conventional name `values.yaml` so that it's better googleable.

PR #776, opened 2025-06-10
Spdx License Identifier

markstur:
Update source file licenses to use SPDX-License-Identifier instead of comment block for LF.


JanPokorny:
I applied this manually since it had more conflicts already 😁 Thanks!

PR #784, opened 2025-06-12
feat: configurable base path

grahamwhiteuk:
Allows the base path to be configurable.

@penge is working on an update to the UI component such that the whole platform can be configured to a specific base url.

### DIFF
--- a/apps/beeai-server/src/beeai_server/application.py
+++ b/apps/beeai-server/src/beeai_server/application.py
@@ -98,13 +98,13 @@ def register_global_exception_handlers(app: FastAPI):
         return await http_exception_handler(request, exception)
 
 
-def mount_routes(app: FastAPI):
+def mount_routes(app: FastAPI, base_path: str = "/"):
     static_directory = pathlib.Path(__file__).parent.joinpath("static")
     if not static_directory.joinpath("index.html").exists():  # this check is for running locally
         raise RuntimeError("Could not find static files -- ensure that beeai-ui is built: `mise build:beeai-ui`")
 
     ui_app = FastAPI(docs_url=None, redoc_url=None, openapi_url=None)
-    ui_app.mount("/", NoCacheStaticFiles(directory=static_directory, html=True))
+    ui_app.mount(base_path, NoCacheStaticFiles(directory=static_directory, html=True))
     ui_app.add_exception_handler(
         404,
         lambda _req, _exc: FileResponse(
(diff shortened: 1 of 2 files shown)


jezekra1:
Can you please explain why you need this?
It has implications for UI, CLI and docs as well

You can achieve the same thing using an nginx ingress rules for non-standard configurations

penge:
@jezekra1

**Reason:**
We'd like to serve the UI and Docs from subdirectories like `/granite/playground/` and `/granite/docs/`, in environments where we don't control the root path.

**Specifically, we will use:**
https://www.ibm.com/granite/playground/
https://www.ibm.com/granite/docs/

> Ingress rules can handle routing, but the UI and Docs need to know the base path at build time to correctly resolve static assets and internal links. Without a configurable base path, they break when not served from /.

apps/beeai-server/src/beeai_server/application.py
```diff
@@ -122,7 +122,7 @@ def mount_routes(app: FastAPI):
     server_router.include_router(ui_router, prefix="/ui", tags=["ui"])
 
     app.mount("/healthcheck", lambda: "OK")
-    app.include_router(server_router, prefix="/api/v1", tags=["provider"])
+    app.include_router(server_router, prefix=f"{base_path}api/v1", tags=["provider"])
     app.mount("/", ui_app)
```
JanPokorny, line 126: Isn't `base_path` a misnomer when we don't use it for the UI base `/` or `/healthcheck`? It seems inconsistent when it prefixes only some of the routes.

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -128,13 +128,21 @@ class Configuration(BaseSettings):
...
         oci_registry = defaultdict(OCIRegistryConfiguration)
         oci_registry.update(self.oci_registry)
         self.oci_registry = oci_registry
         return self
 
+    @model_validator(mode="after")
+    def _base_path_ends_in_slash(self):
+        if not self.base_path.endswith("/"):
+            self.base_path += "/"
+        return self
```
JanPokorny, line 144: This is a bit cumbersome. I think you could do this more elegantly by either joining using `os.path.join` (which handles `/` automatically) or by defining the field like `Field(default="/", pattern=r".*/$")`.

jezekra1:
closing this for inactivity

PR #792, opened 2025-06-13
feat(beeai-cli): bundle lima

JanPokorny:
Publish platform-specific wheels bundling Lima VM executables. Switch to uv as a build backend. Add an alias to make `uvx beeai-cli platform start` etc. work.


apps/beeai-cli/src/beeai_cli/commands/platform.py
```diff
@@ -82,34 +93,60 @@ def _lima_yaml(k3s_port: int = 16443, beeai_port: int = 8333, phoenix_port: int
...
 
+@functools.cache
+def _limactl_exe():
+    bundled_limactl = importlib.resources.files("beeai_cli") / "data" / "limactl"
+    return bundled_limactl if bundled_limactl.is_file() else "limactl"
+
+
 def _validate_driver(vm_driver: VMDriver | None) -> VMDriver:
+    is_windows = platform.system() == "Windows" or shutil.which("wsl.exe")
+    has_lima = (importlib.resources.files("beeai_cli") / "data" / "limactl").is_file() or shutil.which("limactl")
```
jezekra1, line 111: Can we issue a warning on fallback to `limactl`? - this should be an unsupported configuration

Podman-compose example
```
$ podman compose --log-level=error --progress=quiet run --rm --quiet-pull foobar
>>>> Executing external compose provider "/usr/libexec/docker/cli-plugins/docker-compose". Please refer to the documentation for details. <<<<
```

Also the file should never be missing right? Does it make sense to start writing multi-platform tests for CLI at this point? :)

apps/beeai-cli/src/beeai_cli/commands/platform.py
```diff
@@ -82,34 +93,60 @@ def _lima_yaml(k3s_port: int = 16443, beeai_port: int = 8333, phoenix_port: int
...
+    has_docker = shutil.which("docker")
```
jezekra1, line 112: Same with docker - maybe a warning that the docker driver is experimental (except on windows)?

jezekra1 (approved):

PR #793, opened 2025-06-13
docs: improve installation instructions

JanPokorny:
> [!WARNING]
> Merge only after releasing the bundled `limactl` -- Linux instructions have changed.

Put all installation instructions in a single list. (Adds a bit of redundancy in code but it looks better visually.) Unify all platforms under `uv`.


jenna-winkler:
This layout is way better thanks for doing this

PR #796, opened 2025-06-16
feat(ui): show only agents with supported ui types

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/699

### DIFF
--- a/apps/beeai-ui/src/modules/agents/api/queries/useListAgents.ts
+++ b/apps/beeai-ui/src/modules/agents/api/queries/useListAgents.ts
@@ -22,22 +22,18 @@ import { listAgents } from '..';
 import { agentKeys } from '../keys';
 import type { Agent, ListAgentsParams } from '../types';
 
-interface Props {
-  params?: ListAgentsParams;
-}
-
-export function useListAgents({ params }: Props = {}) {
+export function useListAgents({ onlyUiSupported, sort }: ListAgentsParams = {}) {
   const query = useQuery({
-    queryKey: agentKeys.list({ params }),
+    queryKey: agentKeys.list(),
     queryFn: listAgents,
     select: (data) => {
       let agents = data?.agents as Agent[];
 
-      if (params?.onlySupportedUis) {
+      if (onlyUiSupported) {
         agents = agents.filter(isAgentUiSupported);
       }
 
-      if (params?.sort) {
+      if (sort) {
         agents = agents.sort(sortAgentsByName);
       }
 
--- a/apps/beeai-ui/src/modules/agents/list/AgentsView.tsx
+++ b/apps/beeai-ui/src/modules/agents/list/AgentsView.tsx
@@ -35,7 +35,7 @@ export function AgentsView() {
     error,
     refetch,
     isRefetching,
-  } = useListAgents({ params: { onlySupportedUis: true, sort: true } });
+  } = useListAgents({ onlyUiSupported: true, sort: true });
 
   const { watch } = useFormContext<AgentsFiltersParams>();
   const filters = watch();
--- a/apps/beeai-ui/src/components/AgentsNav/AgentsNav.tsx
+++ b/apps/beeai-ui/src/components/AgentsNav/AgentsNav.tsx
@@ -30,7 +30,7 @@ export function AgentsNav() {
   const { pathname } = useLocation();
   const { transitionTo } = useViewTransition();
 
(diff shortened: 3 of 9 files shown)


apps/beeai-ui/src/modules/agents/list/AgentsView.tsx
```diff
@@ -26,15 +26,18 @@ import { AgentCard } from '../components/AgentCard';
 import { AgentsFilters } from '../components/AgentsFilters';
 import { AgentsList } from '../components/AgentsList';
 import { ImportAgents } from '../components/ImportAgents';
+import { useSupportedAgents } from '../hooks/useSupportedAgents';
 import type { AgentsFiltersParams } from '../providers/AgentsFiltersProvider';
 
 export function AgentsView() {
   const { data, isPending, error, refetch, isRefetching } = useListAgents();
+  const agents = useSupportedAgents({ agents: data });
```
penge: It seems to me that the already existing `useListAgents()` hook could accept an optional `onlyUiSupported` parameter, and even a `sort` parameter — to keep things centralized and avoid introducing an additional hook.

Reasons why NOT to use an extra hook, and rely on `useListAgents()` instead:
1. **Centralized logic** – keeps all agent-fetching logic in one place.
2. **Fewer re-renders and dependencies** – avoids extra computation after the initial hook resolves.
3. **Cleaner and more readable** – see the example below.
4. **Less boilerplate** – fewer imports, fewer variables, and less data transformation.

### Example

```tsx
const { data: agents } = useListAgents({ onlyUiSupported: true, sort: true });
```

PetrBulanek: This is definitely a good point. Thanks!

apps/beeai-ui/src/modules/agents/api/queries/useListAgents.ts
```diff
@@ -16,15 +16,33 @@
...
+      if (params?.onlySupportedUis) {
+        agents = agents.filter(isAgentUiSupported);
+      }
+
+      if (params?.sort) {
+        agents = agents.sort(sortAgentsByName);
+      }
+
+      return agents;
+    },
```
penge: Introduces a **fragile coupling** between `queryKey` and `select`.

1. If one side (`queryKey` or `select`) is updated without the other, the hook can return inconsistent data — a breeding ground for subtle bugs.
2. The `queryKey` should encode only the inputs that affect the **actual** _fetch_. Because `listAgents()` ignores `params`, adding them to the `queryKey` just pollutes the cache key.
3. Nesting everything under `params` adds unnecessary boilerplate and hurts DX.
4. The key name `onlySupportedUis` is awkward — it should be **singular**: `onlyUiSupported` (you're filtering agents, not UIs, and "UI" is not a countable plural here).

### Better approach

1. Include only fetch-affecting inputs in `queryKey`. **Here:** drop `params`.
2. Keep options flat. **Here:** expose `onlyUiSupported` and `sort` directly instead of nesting them.
3. Use clear and consistent naming. **Here:** rename `onlySupportedUis` → `onlyUiSupported`.

penge (changes requested):

apps/beeai-ui/src/modules/agents/api/queries/useListAgents.ts
PetrBulanek: Thanks for the detailed feedback! :-)

Just to clarify a couple of points:

The params object is intended to eventually be passed directly to the listAgents API call, once backend support is in place. So while it’s currently only used client-side (in select), the idea was to structure things now in a way that’s ready for future API expansion. I see the point about avoiding coupling between queryKey and select, so I’ll adjust it accordingly for now and reintroduce params into it when it actually affects the fetch.

The params nesting is somewhat subjective. In this case, I chose to nest it to clearly separate filtering-specific params from other hook/query options like enabled, refetchInterval that might be passed in the future. Once params gets passed to the API, I think this separation will be more justified and helpful for maintainability.

penge: I understand your point about thinking about the future or thinking of params as parameters for the API call (which I would still prefer not to expose as `params`, but as destructured keys for easy use from the outside).

About the cache key for the future uses, let's bring it back when API actually supports it, so we don't get into some issues with the cache.

penge (approved):

PR #798, opened 2025-06-16
feat(ui): add and show displayName for agents

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/675

### DIFF
--- a/apps/beeai-ui/src/modules/agents/utils.ts
+++ b/apps/beeai-ui/src/modules/agents/utils.ts
@@ -75,6 +75,10 @@ export function getAvailableAgentLinkUrl<T extends AgentLinkType | AgentLinkType
 }
 
 export function getAgentDisplayName(agent: Agent) {
-  const { name, metadata } = agent;
-  return metadata.name ?? name;
+  const {
+    name,
+    metadata: { annotations },
+  } = agent;
+
+  return annotations?.display_name ?? name;
 }
--- a/agents/community/aider/src/aider_agent/agent.py
+++ b/agents/community/aider/src/aider_agent/agent.py
@@ -65,6 +65,7 @@ server = Server()
             "**Bash/Shell Scripting Assistance** – Automates script writing, optimization, and debugging.",
         ],
         ui={"type": "hands-off", "user_greeting": "Define your programming task"},
+        annotations={"display_name": "Aider"},
         env=[
             {
                 "name": "LLM_MODEL",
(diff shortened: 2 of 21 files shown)


apps/beeai-ui/src/modules/agents/utils.ts
```diff
@@ -75,6 +75,10 @@ export function getAvailableAgentLinkUrl<T extends AgentLinkType | AgentLinkType
...
 
 export function getAgentDisplayName(agent: Agent) {
-  const { name, metadata } = agent;
-  return metadata.name ?? name;
+  const {
+    name,
+    metadata: { annotations },
+  } = agent;
+
+  return annotations?.display_name ?? name;
```
penge: ### Issues

1. **Hidden fallback:** At first glance, `getAgentDisplayName(agent)` suggests it always returns `display_name`. The fallback to `name` isn't immediately obvious.
2. **Excessive indirection:** A simple one-liner is extracted into a utility function used across many files, adding mental overhead and development friction.
3. **Import bloat:** Importing `getAgentDisplayName` in multiple files clutters the codebase and reduces component readability.
4. **Risk of inconsistent usage:** Developers might forget to use the function, leading to inconsistent display logic across the app.

---

### Suggestion

The _cleanest and most efficient_ approach is to augment the React Query hook directly.  
For a field that prefers `display_name` but falls back to `name`, introduce a new field called `label`.

---

### Benefits of augmenting the hook

1. **Centralized logic:** The fallback is handled once inside the React Query hook. All consumers receive extended Agent objects automatically.
2. **Improved readability:** Components simply use `agent.label`, with no need to understand or call a helper function.
3. **No redundant imports:** No need to import a utility function in every file — the data comes ready to use.
4. **Scales cleanly:** Additional fields can be added later without affecting consuming code.
5. **Type-safe and declarative:** With a type like `ExtendedAgent`, it's immediately clear which fields are available — and why.

---

### Example

In the hook:

```ts
.map((agent) => ({
  ...agent,
  label: agent.metadata?.annotations?.display_name ?? agent.name,
}))
```

In components:

```tsx
{agent.label}
```

penge (changes requested):
I'd suggest removing `getAgentDisplayName` and enriching the `Agent` object directly in the React Query hook — for example, by adding a `label` field.

tomkis:
This has to be reworked with the new Annotations metadata concept.

PR #821, opened 2025-06-19
feat(server): add embeddings endpoint

xjacka:


apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -365,6 +365,18 @@ async def setup(
             if "Hello" not in response_text:
                 err_console.print(format_error("Error", "Model did not provide a proper response."))
                 return False
+
+            if provider_name in ["watsonx", "OpenAI", "Ollama"]:
```
JanPokorny: Cohere also supports embeddings, IDK about the others.

JanPokorny: I'd replace this hardcoded condition by adding another field in the provider list at the top, `recommended_embedding_model`, which would be `None` if the provider can't do embedding. Then I'd just ask `f"For embedding, do you want to use the recommended model {recommended_embedding_model}?"`, and continue to model selection only if user says no.

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -365,6 +365,18 @@ async def setup(
...
+                selected_embedding_model = None
+                if available_models and len(available_models) >= 1:
+                    selected_embedding_model = await inquirer.fuzzy(
+                        message="Select an embedding model (type to search):",
+                        choices=sorted(["None", "Write model name", *available_models]),
```
JanPokorny: - You can use `Choice(name="None", value=None)` so you don't have to rely on string comparison (and avoid edge cases like a model literally called `None` 😁 )
- `"Write model name"` does not make sense here, either the model provider does not support listing models (watsonx) in which case this step is skipped (since `available_models` is empty), or the model provider lists all models and there is no point in writing in a custom value (Ollama, OpenAI).

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -365,6 +365,18 @@ async def setup(
...
+                    ).execute_async()
+
+                if selected_embedding_model == "Write model name" or len(available_models) == 0:
+                    await inquirer.text(message="Write a model name to use:").execute_async()
```
JanPokorny:
```suggestion
                    await inquirer.text(message="Write a model name to use for embedding:").execute_async()
```

apps/beeai-server/src/beeai_server/api/routes/embeddings.py
```diff
@@ -0,0 +1,99 @@
...
+    model: str
+    usage: Dict[str, int] = {
+        "prompt_tokens": int,
+        "total_tokens": int,
+        "completion_tokens": int,
+    }
+    data: List[EmbeddingsDataItem]
+
+
+@router.post("")
```
JanPokorny: I'd aim for OpenAI-compatible API, thus put this together with the LLM proxy, under `/llm/embeddings` -- to correspond to `/llm/chat/completions` with `/llm` as the OpenAI API base.

apps/beeai-cli/src/beeai_cli/commands/env.py
xjacka: This will be changed when we refactor providers. You'll be able to set up multiple providers, even for the embedding models. We'll change this soon.

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -366,16 +384,56 @@ async def setup(
...
-            if provider_name in ["watsonx", "OpenAI", "Ollama"]:
-                selected_embedding_model = None
-                if available_models and len(available_models) >= 1:
-                    selected_embedding_model = await inquirer.fuzzy(
-                        message="Select an embedding model (type to search):",
-                        choices=sorted(["None", "Write model name", *available_models]),
-                    ).execute_async()
+            selected_embedding_model = None
+            if (
+                provider_name in ["watsonx", "OpenAI", "Ollama", "Cohere"]
```
JanPokorny: Perhaps here it should be `recommended_embedding_model is not None`? Or why are there recommended embedding models defined for providers not listed here?

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -366,16 +384,56 @@ async def setup(
...
+                and await inquirer.confirm(
+                    message="Do you want to set up embedding model?", default=True
```
JanPokorny:
```suggestion
                    message="Do you want to set up an embedding model?", default=True
```

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -366,16 +384,56 @@ async def setup(
...
+                    and selected_embedding_model not in available_models
+                ):
+                    try:
+                        await run_command(
+                            ["ollama", "pull", selected_embedding_model],
+                            "Pulling the selected embedding model",
+                            check=True,
+                        )
+                    except Exception as e:
+                        console.print(f"[red]Error while pulling embrdding model: {e!s}[/red]")
```
JanPokorny:
```suggestion
                        console.print(f"[red]Error while pulling the embedding model: {e!s}[/red]")
```

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -366,16 +384,56 @@ async def setup(
...
+                            ),
+                            default=True,
+                        ).execute_async()
+                    )
+                    else (
+                        await inquirer.fuzzy(
+                            message="Select an embedding model (type to search):",
+                            choices=sorted(available_models),
+                        ).execute_async()
+                        if available_models and len(available_models) > 1
```
JanPokorny: It's missing these lines from the equivalent code above

```suggestion
                        if available_models and len(available_models) > 1
                        else available_models[0]
                    	if available_models and len(available_models) == 1
```

But actually not sure about this now that we mix embedding and chat models, maybe we want to drop this shortcut and always show a list of models even when there's just one, so that users know what's available

jezekra1 (approved):

PR #824, opened 2025-06-19
Add procrastinate

jezekra1:
future TBD: Optional configuration to run workers separately in helm chart


apps/beeai-server/src/beeai_server/jobs/crons/provider.py
```diff
@@ -78,3 +94,24 @@ async def check_registry(configuration: Configuration, provider_service: Provide
 
     if errors:
         raise ExceptionGroup("Exceptions occurred when reloading providers", errors)
+
+
+if get_configuration().provider.auto_remove_enabled:
+
+    @blueprint.periodic(cron="* * * * * */5")
```
jezekra1, line 101: This will run every 5 seconds - creating few thousand records in a database over the span of 1 hour.
Cleanup job will cleanup the records older than 1 hour every 10 minutes.

I don't think it's a big deal considering postgres performance with millions of rows, just a heads up that this is happening.

Previously all crons were running in the background using asyncio, but that prevents scaling platform server horizontally - we don't want the same cron to be running in each pod.

JanPokorny (approved):

PR #826, opened 2025-06-20
feat(beeai-server): add docling integration

xjacka:


jezekra1:
closing because we'll use official docling image instead

PR #830, opened 2025-06-24
Add vector stores

jezekra1:


apps/beeai-server/src/beeai_server/api/schema/vector_stores.py
```diff
@@ -0,0 +1,20 @@
...
+
+    name: str = Field(..., description="Name of the vector store")
+    dimension: int = Field(..., description="Dimension of the vectors to be stored")
+    model_id: str
+
+
+class SearchRequest(BaseModel):
+    """Request to search a vector store."""
+
+    query_vector: list[float] = Field(None, description="Vector to search for")
```
xjacka, line 19: `query` and `max_num_results` to be aligned with OpenAI?

apps/beeai-server/src/beeai_server/domain/models/common.py
```diff
@@ -0,0 +1,34 @@
...
+from pydantic import Field, AfterValidator
+
+
+def validate_metadata(metadata: dict[str, str]) -> dict[str, str]:
+    if len(metadata) > 16:
+        raise ValueError("Metadata must be less than 16 keys.")
+    if any(len(v) > 64 for v in metadata.keys()):
+        raise ValueError("Metadata keys must be less than 64 characters.")
+    if any(len(v) > 512 for v in metadata.values()):
+        raise ValueError("Metadata values must be less than 64 characters.")
```
xjacka:
```suggestion
        raise ValueError("Metadata values must be less than 512 characters.")
```

JanPokorny:
Are we targeting OpenAI API compatibility (https://platform.openai.com/docs/api-reference/vector-stores)? Maybe we should (for simplicity), in which case we should keep the API together with `/chat/completions` and `/embeddings` on the same API base

jezekra1:
It's not possible to use the openai API how we designed it, it's more high level - works with files not vectors. We wanted to give users more control over the data, so it's more like "simplified perplexity" api

PR #840, opened 2025-06-25
chore: use spdx-license-identifier instead of comment block in source files

markstur:
Update source file licenses to use SPDX-License-Identifier instead of comment block for LF.

SPDX-License_Identifier has become the standard for concise and machine-readable license comments.
This will allow license-scanner to recognize that our source files have license headers as recommended.


markstur:
This was an attempt to get e2e to run (original PR was from a fork which does not work), but this push fixed the old PR so I'll close this new one.

PR #844, opened 2025-06-25
feat(ui): add trajectory metadata components

PetrBulanek:
Refs: https://github.com/i-am-bee/beeai-platform/issues/651


apps/beeai-ui/src/modules/runs/trajectory/components/TrajectoryItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+          >
+            <ChevronDown />
+          </IconButton>
+        )}
+
+        {tool_name && (
+          <h3 className={classes.name}>
+            {/* <span className={classes.icon}>
+            <Icon />
+          </span> */}
```
kapetr, line 47: Do we expect to have the icon?

PetrBulanek: Yes, we do, at least for the known tools it was part of the design.

PR #860, opened 2025-06-30
feat(ui): align agent message role with api

kapetr:
Just aligning the agent message role with the API - currently it does not make any difference, since the UI doesn't exchange agent messages with the API, but this will be necessary once session history is implemented.

### DIFF
--- a/apps/beeai-ui/src/modules/runs/chat/types.ts
+++ b/apps/beeai-ui/src/modules/runs/chat/types.ts
@@ -3,13 +3,13 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { CreateRunStreamRequest, TrajectoryMetadata } from '../api/types';
+import type { TrajectoryMetadata } from '../api/types';
 import type { SourceReference } from '../sources/api/types';
 import type { Role } from '../types';
 
 interface Message {
   key: string;
-  role: CreateRunStreamRequest['input'][number]['role'];
+  role: Role | string;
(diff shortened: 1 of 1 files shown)


apps/beeai-ui/src/modules/runs/chat/types.ts
```diff
@@ -3,22 +3,22 @@
...
 
-import type { TrajectoryMetadata } from '../api/types';
+import type { CreateRunStreamRequest, TrajectoryMetadata } from '../api/types';
 import type { SourceReference } from '../sources/api/types';
 import type { Role } from '../types';
 
 interface Message {
   key: string;
-  role: Role;
+  role: CreateRunStreamRequest['input'][number]['role'];
```
PetrBulanek: We use "Role.User" and "Role.Agent | string" in UserMessage and AgentMessage specific interfaces, so I think we should use "Role | string" in the base Message interface. Once the acp-sdk is imported and used, we can probably use type from the sdk and not to implement it ourselves s we do for now.

PetrBulanek (changes requested):

apps/beeai-ui/src/modules/runs/chat/types.ts
kapetr: I thought this would validate our `Role` type against any changes in the schema, but it turns out it won’t, since the schema definition includes a generic `string` type. I'll change it back as you suggest.

PR #863, opened 2025-06-30
feat(ui): file upload enabled by agent 'input_content_types'

### LINKED ISSUE #819: Restrict File Upload UI to Agents That Declare File Support
## Problem Statement

Currently, the file upload button appears for all agents, even those that can't process files, which creates a confusing user experience. We need a way to only show the file upload UI when it’s actually usable.

## Proposed Solution

Leverage ACP's input/output content types as proposed in the https://github.com/i-am-bee/acp/pull/161

UI will only show the file upload button if the agent has declared one or more supported MIME types.

The content types can also be leveraged to inform end users which file types are accepted by the agent (e.g., "text/plain", "image/*").

## Impact 

- Prevents misleading UX
- Gives agent developers explicit control
- Opens future possibility of tailoring UI messaging based on declared MIME types

kapetr:
Closes #819 

Note that schema.d.ts was edited manually for now, until `acp-sdk` is updated to a version that includes `input_content_types`.


apps/beeai-ui/src/api/schema.d.ts
PetrBulanek, line 1: Removing the TrajectoryMetadata is probably a mistake, right?

PetrBulanek (changes requested):

apps/beeai-ui/src/api/schema.d.ts
kapetr: It looks like it’s missing from the OpenAPI schema, this happened after generating types from it. Let’s wait for the new version of `acp-sdk`, I’ve updated the PR description.

PR #870, opened 2025-07-01
refactor(ui): agent run

penge:
Simplify the `<AgentRun>` component by replacing nested ternaries with clear `if` statements.


PetrBulanek:
I am currently working on refactoring hands-off/chat ui into common components and also to use shared provider as there are a lot of same stuff and adding features (files, sources, trajectories) meant code duplications. Your suggestion is actually part of that. You can check it here https://github.com/i-am-bee/beeai-platform/blob/ui/hands-off-ui-rework/apps/beeai-ui/src/modules/runs/components/AgentRun.tsx

penge:
@PetrBulanek Thanks for letting me know. You may then easily refactor, right? (moving out what's inside the conditions) Or is there a reason not to merge?

PR #885, opened 2025-07-03
docs: Create gui-components.mdx

jenna-winkler:


docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
+---
+title: "GUI Components"
+description: "Customize how your agents appear and behave in the BeeAI platform interface"
+---
+
+GUI components in the BeeAI platform define how your agents appear and behave in the user interface. Through the `PlatformUIAnnotation` metadata, you can customize the visual presentation and interaction patterns for your agents.
+
+## Configuration
+
+The `PlatformUIAnnotation` is the primary configuration object that defines how your agent appears in the GUI. It's specified within your agent's metadata annotations.
```
PetrBulanek: Shall we add a link to the acp documentation (https://agentcommunicationprotocol.dev/core-concepts/agent-manifest#schema-metadata) to the "agent's metadata"?

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+```python
+from acp_sdk.models.platform import PlatformUIAnnotation, PlatformUIType
```
PetrBulanek: I think we should add proper imports for Metadata, Annotations and AgentToolInfo as well.

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+The chat interface (`PlatformUIType.CHAT`) provides a conversational experience.
+
+<Info>
+    Chat UI is best suited for conversational agents, customer support bots, and multi-turn dialogues.
+</Info>
+
+### Hands-off UI
+
+The hands-off interface (`PlatformUIType.HANDS_OFF`) is designed for autonomous agents that work independently:
```
PetrBulanek: I think it's actually `PlatformUIType.HANDSOFF` (without the underscore).

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
...
+            name="Calculator",
+            description="Perform mathematical calculations"
+        ),
+        AgentToolInfo(
+            name="File Processor",
+            description="Read and analyze uploaded files"
+        )
+    ]
+)
+```
```
PetrBulanek: In my opinion, we should move this next to the other `PlatformUIAnnotation` stuff.

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+## Advanced Features
+
+### Dynamic Content Rendering
+
+The platform automatically renders various content types within your agent responses:
+
+#### Markdown Support
```
PetrBulanek, line 171: We also render Markdown images if present inside the message. :-)

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
...
+- Code blocks with syntax highlighting
+- Tables and structured data
+- Links and references
+
+### Trajectory Visualization
+
+The platform provides built-in trajectory visualization to show the step-by-step process of agent execution:
+
+```python
+from acp_sdk.models.models import TrajectoryMetadata
```
PetrBulanek, line 184: Add `MessagePart` import.

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+```python
+# Citations are automatically rendered when included in message metadata
+MessagePart(
+    content="According to recent studies...",
+    metadata=CitationMetadata(
+        url="https://example.com/study",
+        title="Recent Research Study",
+        description="Comprehensive analysis of the topic"
+    )
```
PetrBulanek: It should also provide `start_index` and `end_index` to correctly render in GUI.

PetrBulanek (changes requested):

docs/concepts/gui-components.mdx
jenna-winkler: Good call, just added a note

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,243 @@
+---
+title: "GUI Components"
+description: "Customize how your agents appear and behave in the BeeAI platform interface"
+---
+
+GUI components in the BeeAI platform define how your agents appear and behave in the user interface. Through the `PlatformUIAnnotation` metadata, you can customize the visual presentation and interaction patterns for your agents.
```
tomkis: I'd add word graphical here so there's no confusion with CLI UI which does not support this yet.

```suggestion
GUI components in the BeeAI platform define how your agents appear and behave in the graphical user interface. Through the `PlatformUIAnnotation` metadata, you can customize the visual presentation and interaction patterns for your agents.
```

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+### Hands-off UI
+
+The hands-off interface (`PlatformUIType.HANDSOFF`) is designed for autonomous agents that work independently:
+
+- **Task Execution**: Focus on agent actions rather than conversation
+- **Result Display**: Clear presentation of final outcomes
+
+<Info>
+    Hands-off UI is ideal for automated workflows where there is a clear output.
```
tomkis:
```suggestion
    Hands-off UI is ideal for automated workflows where there is a clear output. For example, consider a researcher agent: it receives a clearly defined input which is the research topic and then it produces a well-structured output which is the resulting research.
```

docs/concepts/gui-components.mdx
```diff
@@ -0,0 +1,243 @@
...
+
+GUI components in the BeeAI platform define how your agents appear and behave in the user interface. Through the `PlatformUIAnnotation` metadata, you can customize the visual presentation and interaction patterns for your agents.
+
+<Note>
+To fully customize your agent's appearance and behavior in the interface, you'll be using metadata defined by the Agent Communication Protocol (ACP). For details on how metadata annotations work, see the [ACP metadata documentation](https://agentcommunicationprotocol.dev/core-concepts/agent-manifest#schema-metadata).
+</Note>
+
+## Configuration
+
+The `PlatformUIAnnotation` is the primary configuration object that defines how your agent appears in the GUI. It's specified within your agent's metadata annotations.
```
tomkis: @jenna-winkler this seems a bit off here - because as a reader it's hard to me understand why Configuration is anyhow different from Display Properties. 

The idea here is that `PlatformUIAnnotation` is extension of ACP to serve things that are not part of the protocol, while the other section (Display Properties) covers standardized parts of the protocol. They both do serve for definition of how agent appear in the GUI.

The way this is documented here just makes me wonder why configuration is to define how agent appears in GUI when the display properties are doing effectively the same and why they are separated.

jenna-winkler: Great point, making adjustments now

tomkis (approved):

Issue #904, opened 2025-07-08
IBM Cloud Postgresql support


flekmatik:
**Pre-requisities**
- [x] I am using the newest version of the platform (`uv tool upgrade beeai-cli`)

**Describe the bug**
When using external DB hosted in IBM cloud getting error: `ERROR: extensions can only be installed in schema 'ibm_extension'`. The issue is with command `CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog`

**To Reproduce**

1. Setup with IBM Cloud Postgresql managed instance
2. Start beeai-platform

**Expected behavior**
No error, running beeai-platform

**Logs / Screenshots / Code snippets**
```
CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
....
sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error) <class 'asyncpg.exceptions.FeatureNotSupportedError'>: extensions can only be installed in schema 'ibm_extension'
```

**Additional context**
`CREATE EXTENSION IF NOT EXISTS plpgsql` should work fine in and out of IBM Cloud

JanPokorny:
Closing due to lack of follow up

PR #905, opened 2025-07-08
fix: ibm cloud support (#904)

flekmatik:


apps/beeai-server/src/beeai_server/infrastructure/persistence/migrations/alembic/versions/procrastinate.sql
```diff
@@ -1,6 +1,6 @@
 -- Procrastinate Schema
 
-CREATE EXTENSION IF NOT EXISTS plpgsql WITH SCHEMA pg_catalog;
+CREATE EXTENSION IF NOT EXISTS plpgsql;
```
jezekra1, line 3: Can you please add a section for postgres in `helm/README.md` under `## External Services`?

It should mention all the requirements for the external database - the commands you run as a sql query such as:

`CREATE EXTENSION IF NOT EXISTS plpgsql;`, vector, memory setup etc

jezekra1 (changes requested):

PR #930, opened 2025-07-16
feat(ui): support custom basePath

penge:
Support **custom** `basePath`.

Ensure the **correct** _API URL_ is constructed when called from either the **server** or the **client**.

<br>

### How to test (development)

1. Set:

```
NEXT_PUBLIC_BASE_PATH='/granite/playground'
```

<br>

2. Run:

```
$ mise beeai-ui:run:dev
```

<br>

3. Open: http://localhost:3000/granite/playground

<br>

---

<br>

### How to test (build)

1. Set:

```
NEXT_PUBLIC_BASE_PATH='/granite/playground'
```

<br>

2. Run:

```
$ mise beeai-ui:build:nextjs
$ mise beeai-ui:run
```

<br>

3. Open: http://localhost:3000/granite/playground

<br><br><br>

### Expectations

- API calls use the correct base path and work (both "ACP" and "API").
- Favicon correctly set.

### DIFF
--- a/apps/beeai-ui/src/utils/api/getBaseUrl.ts
+++ b/apps/beeai-ui/src/utils/api/getBaseUrl.ts
@@ -1,0 +1,11 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { API_URL, BASE_PATH } from '#utils/constants.ts';
+
+export function getBaseUrl(suffix = '') {
+  const baseUrl = typeof window !== 'undefined' ? BASE_PATH : API_URL;
+  return baseUrl + suffix;
+}
--- a/apps/beeai-ui/.env.example
+++ b/apps/beeai-ui/.env.example
@@ -1,6 +1,10 @@
-API_URL='http://127.0.0.1:8333'
-
 ############# OPTIONAL #############
+
+# Default: 'http://127.0.0.1:8333'
+API_URL=
+
+# Default: ''
+NEXT_PUBLIC_BASE_PATH=
 
 # Default: 'BeeAI'
 NEXT_PUBLIC_APP_NAME=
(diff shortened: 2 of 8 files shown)


apps/beeai-ui/src/utils/api/getBaseUrl.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { API_URL, BASE_PATH } from '#utils/constants.ts';
+
+export function getBaseUrl(suffix = '') {
+  const baseUrl = typeof window !== 'undefined' ? BASE_PATH : API_URL;
+  return baseUrl + suffix;
+}
```
kapetr, line 11: Maybe consider a different name, since it doesn’t actually return the base URL when a suffix is provided?

kapetr (approved):

apps/beeai-ui/src/utils/api/getBaseUrl.ts
penge: Let me see.

PetrBulanek: I'm just not sure what the function is for anyway? In current contexts where this is called, it will always return the API_URL, if I'm not mistaken?

penge: Actually, in the ACP, where we do provide a suffix, it still is a "base" in that context.

But primarily, I wanted to make the function name consistent with the `baseUrl` attribute:

```
baseUrl: getBaseUrl(...)
```

kapetr: This is wrong, but we'll refactor with @PetrBulanek

PR #932, opened 2025-07-16
feat(ui): add feature_flags env

kapetr:


apps/beeai-ui/src/contexts/AppConfig/AppConfigProvider.tsx
PetrBulanek, line 1: I was wondering if we could remove this provider and include the feature flags under the AppProvider, that would simplify things a bit. I don't think it's necessary to have it separate. What do you think?

apps/beeai-ui/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { z } from 'zod';
+
+import { objectFromEntries } from './helpers';
+
+const featureNames = ['Variables', 'Providers'] as const;
+const featureFlagsDefaults = objectFromEntries(featureNames.map((key) => [key, false]));
```
PetrBulanek: You can use `[key, false] as const` for better type inference.

apps/beeai-ui/src/app/layout.tsx
PetrBulanek, line 1: How about moving this `JSON.parse` call into the`parseFeaterFlags` helper function and `process.env.FEATURE_FLAGS` to the constants file and just import it here as `FEATURE_FLAGS`?

Result: `parseFeatureFlags(FEATURE_FLAGS)`

apps/beeai-ui/src/app/layout.tsx
```diff
@@ -52,7 +53,7 @@ export default function RootLayout({
         <script dangerouslySetInnerHTML={{ __html: darkModeScript }} />
       </head>
       <body>
-        <Providers>
+        <Providers featureFlags={parseFeatureFlags(JSON.parse(process.env.FEATURE_FLAGS ?? ''))}>
```
penge: Throws if `FEATURE_FLAGS` is unset or invalid JSON.

apps/beeai-ui/src/contexts/AppConfig/AppConfigProvider.tsx
kapetr: Agree!

apps/beeai-ui/src/utils/feature-flags.ts
kapetr: I can, but it doesn't make any difference in the end, if I didn't miss something.

PetrBulanek (approved):

apps/beeai-ui/src/app/layout.tsx
```diff
@@ -52,7 +53,7 @@ export default function RootLayout({
         <script dangerouslySetInnerHTML={{ __html: darkModeScript }} />
       </head>
       <body>
-        <Providers>
+        <Providers featureFlags={parseFeatureFlags(FEATURE_FLAGS)}>
```
penge: How about we parse `FEATURE_FLAGS` in `'#utils/constants.ts'` just like `NAV_ITEMS`?

**Reasons:**

- Remove 1 import for parsing.
- Simplify use of `FEATURE_FLAGS` as already parsed.
- Having all parsing at one place (in constants).

apps/beeai-ui/.env.example
```diff
@@ -1,4 +1,5 @@
 API_URL='http://127.0.0.1:8333'
+FEATURE_FLAGS='{"Variables":true,"Providers":true}'
```
penge: It seems **_optional_**.

Let's move under:
```
############# OPTIONAL #############
```

And specify the **default**.

apps/beeai-ui/next.config.ts
```diff
@@ -79,6 +79,9 @@ const nextConfig: NextConfig = {
     // https://github.com/vercel/next.js/issues/64921
     cssChunking: false,
   },
+  devIndicators: {
+    position: 'bottom-right',
+  },
```
penge, line 84: Not related to the PR – but ok. Would deserve a separate commit for organization at least.

apps/beeai-ui/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const featureNames = ['Variables', 'Providers'] as const;
+const featureFlagsDefaults = objectFromEntries(featureNames.map((key) => [key, false] as const));
+
+const featureFlagsSchema = z
+  .object(objectFromEntries(featureNames.map((key) => [key, z.boolean().optional().default(false)])))
+  .strict();
+
+export type FeatureFlags = z.infer<typeof featureFlagsSchema>;
+export type FeatureName = (typeof featureNames)[number];
```
penge, line 18: Maybe I am missing something but this seems too complicated. Why not having something like?

```ts
const featureFlagsSchema = z.object({
  Variables: z.boolean().optional().default(false),
  Providers: z.boolean().optional().default(false),
}).strict();
```

<br>

I saw somewhere **'Theme'** but that is just an extension to the provided feature flags.

apps/beeai-ui/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function parseFeatureFlags(data?: string) {
+  if (data) {
```
penge, line 21: I personally don't like the nesting because of `data`. We could either change it to a guard clause, or remove completely.

apps/beeai-ui/src/modules/settings/SettingsView.tsx
```diff
@@ -6,42 +6,54 @@
...
       </Container>
     </MainContent>
   );
 }
+
+const ITEMS: { title: string; component: ComponentType; featureName?: FeatureName }[] = [
+  { title: 'Variables', component: VariablesView, featureName: 'Variables' },
+  { title: 'Agent providers', component: ProvidersView, featureName: 'Providers' },
+  { title: 'Theme', component: ThemeView },
+];
```
penge, line 59: Maybe we could do just this:

```ts
const featureFlagsToDisplay: Partial<FeatureFlags> = {
  Variables: true,
};

type Items = /* ... */

const ITEMS: Items = [
  ...(featureFlagsToDisplay.Variables
    ? [{ title: 'Variables', component: VariablesView }]
    : []),
  { title: 'Theme', component: ThemeView },
];
```

Basically we use same exact object structure `FeatureFlags` has, so we don't have to extract keys type, and can use it to control what to show or not to show in the UI. We can also drop the `useMemo`.

apps/beeai-ui/src/app/layout.tsx
kapetr: I don't think it's right to do that. The reason I didn’t extract it into a constant originally is because `FEATURE_FLAGS` is a runtime env - it’s only available on the server. It would have a different value on the client, and parsing it there would be unnecessary.

apps/beeai-ui/src/modules/settings/SettingsView.tsx
penge, line 59: How about?

```ts
const items = useMemo(() => getItemsToDisplay(featureFlags), [featureFlags]);

// ...
// ...
// ...

const itemsToConsider: {
  title: string;
  component: ComponentType;
  featureName?: keyof FeatureFlags;
}[] = [
  { title: 'Variables', component: VariablesView, featureName: 'Variables' },
  { title: 'Providers', component: ProvidersView, featureName: 'Providers' },
  { title: 'Theme', component: ThemeView },
];

const getItemsToDisplay = (featureFlags: FeatureFlags) =>
  itemsToConsider.filter(({ featureName }) => !featureName || featureFlags[featureName]);
```

apps/beeai-ui/src/utils/feature-flags.ts
kapetr: I considered that, but I wanted to have clear feature list in one variable, sice they have same type and default value.

penge: How about?

```ts
const flag = z.boolean().optional().default(false);

const featureFlagsSchema = z
  .object({
    Variables: flag,
    Providers: flag,
  })
  .strict();
  ```

Comment (penge) [reply apps/beeai-ui/src/utils/feature-flags.ts:18]: Or how about?

```ts
import { z } from 'zod';

const featureNames = ['Variables', 'Providers'] as const;
const featureFlagsSchema = z
  .object(Object.fromEntries(featureNames.map((key) => [key, z.boolean().catch(false)])))
  .strict();

export type FeatureFlags = z.infer<typeof featureFlagsSchema>;

export function parseFeatureFlags(data?: string) {
  try {
    const featureflags = JSON.parse(data ?? '{}');
    const result = featureFlagsSchema.parse(featureflags);

    return result;
  } catch (error) {
    console.error(error);
  }
}
```

**Summary:**
- keeps the initial idea of having listed names in `featureNames`
- simplifies creation of the schema and fallback to `false` for all flags
- removes `if (data)`
- drops requirement of having `featureFlagsDefaults`

PR #938, opened 2025-07-17
fix(ui): align provider import with current API

kapetr:


apps/beeai-ui/src/modules/providers/api/index.ts
```diff
@@ -32,7 +30,5 @@ export async function registerManagedProvider({ body }: { body: RegisterProvider
   const response = await api.POST('/api/v1/providers', {
     body,
   });
-  const stream = events(ensureResponse(response));
```
PetrBulanek, line 35: I think `ensureResponse` is not used anywhere else, so it could be removed.

PR #939, opened 2025-07-17
Fix upload button

jezekra1:


agents/community/aider/src/aider_agent/agent.py
```diff
@@ -21,6 +21,7 @@
 
 
 @server.agent(
+    input_content_types=["none"],
```
jezekra1, line 24: hacky content type - `[]` does not pass TS sdk validation

PR #941, opened 2025-07-18
fix(server): file_type column migration integrity error

jezekra1:


pilartomas (approved):

apps/beeai-cli/src/beeai_cli/commands/platform.py
```diff
@@ -305,7 +305,14 @@ async def start(
         elif status != "running":
             await run_command(
                 {
-                    VMDriver.lima: [_limactl_exe(), "--tty=false", "start", "--memory=8", vm_name],
+                    VMDriver.lima: [
+                        _limactl_exe(),
+                        "--tty=false",
+                        "start",
+                        "--memory=8",
+                        "--mount=/tmp/beeai:w",
```
jezekra1, line 313: Import images was not working after upgrade due to different lima configuration.
Using this runtime argument (only applies after platform start from a stopped state).

`beeai platform import` after upgrading from previous version will work only after platform restart (also laptop reboot).

PR #944, opened 2025-07-18
feat(server): add separate embedding provider setup

jezekra1:


apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -66,126 +66,394 @@ async def remove_env(
...
+                    Choice(name="au-syd", value="au-syd"),
+                ],
             ).execute_async()
-            if re.match(r"^https://[a-z0-9.-]+\.rits\.fmaas\.res\.ibm\.com/.*$", api_base):
-                provider_name = "RITS"
-                if not api_base.endswith("/v1"):
-                    api_base = api_base.removesuffix("/") + "/v1"
+        }.ml.cloud.ibm.com"""
 
+    if api_base == env["LLM_API_BASE"]:
```
JanPokorny, line 444: It's possible that you would use a different API key here, so perhaps make this an explicit confirmation?

apps/beeai-server/src/beeai_server/service_layer/deployment_manager.py
```diff
@@ -21,6 +21,8 @@ def global_provider_variables(configuration: Configuration):
         "OTEL_EXPORTER_OTLP_ENDPOINT": str(configuration.telemetry.collector_url),
         "PLATFORM_URL": f"http://{configuration.platform_service_url}",
         "EMBEDDING_MODEL": "dummy",
+        "EMBEDDING_API_KEY": "dummy",
+        "EMBEDDING_API_BASE": f"http://{configuration.platform_service_url}/api/v1/llm/embeddings",
```
JanPokorny: I wouldn't include the `/embeddings` in the API base, the same way LLM api base does not include `/chat/completions`

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -66,126 +66,394 @@ async def remove_env(
...
-async def setup(
-    use_true_localhost: typing.Annotated[bool, typer.Option(hidden=True)] = False,
-    verbose: typing.Annotated[bool, typer.Option("-v")] = False,
-) -> bool:
-    """Interactive setup for LLM provider environment variables"""
-    with verbosity(verbose):
-        # Ping BeeAI platform to get an error early
-        async with httpx.AsyncClient() as client:
-            await client.head(str(Configuration().host))
+def _get_llm_providers():
```
JanPokorny: This could probably be a constant `LLM_PROVIDERS=...`

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -66,126 +66,394 @@ async def remove_env(
...
+        Choice(name="Perplexity".ljust(25), value=("Perplexity", "https://api.perplexity.ai", None)),
+        Choice(
+            name="together.ai".ljust(25) + "🆓 has a free tier",
+            value=("Together", "https://api.together.xyz/v1", "deepseek-ai/DeepSeek-R1"),
+        ),
+        Choice(name="Other (RITS, vLLM, ...)".ljust(25) + "🔧 provide API URL", value=("Other", None, None)),
+    ]
+
+
+def _get_embedding_providers():
```
JanPokorny: This could probably be a constant `EMBEDDING_PROVIDERS=...`

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -199,279 +467,151 @@ async def setup(
...
+        console.print("[bold]Setting up LLM provider...[/bold]")
+        if not (llm_env := await _configure_llm()):
             return False
+        embedding_env = {}
+        if await inquirer.confirm(
+            message="Do you want to configure an embedding provider?", default=True
+        ).execute_async():
+            console.print("[bold]Setting up embedding provider...[/bold]")
+            if not (embedding_env := await _configure_embedding(llm_env)):
+                return False
```
JanPokorny, line 626: I don't think this should return false, see `ensure_llm_env` -- embedding model is not strictly needed for most agents to work.

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -66,126 +66,394 @@ async def remove_env(
...
+        Choice(
+            name="Ollama".ljust(25) + "💻 local",
+            value=("Ollama", "http://localhost:11434/v1", "nomic-embed-text:latest"),
+        ),
+        Choice(
+            name="OpenAI".ljust(25),
+            value=("OpenAI", "https://api.openai.com/v1", "text-embedding-3-small"),
+        ),
+        Choice(name="Other (vLLM, ...)".ljust(25) + "🔧 provide API URL", value=("Other", None, None)),
+    ]
```
JanPokorny: Sort embedding providers alphabetically (but "Other" goes last) in code directly, same as LLM providers, and remove `sorted` call further down

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -66,126 +66,394 @@ async def remove_env(
...
+
+
+async def _configure_llm() -> dict[str, str] | None:
+    """Configure LLM provider and return (provider_name, api_base, api_key, model, extra_config)"""
```
JanPokorny: Docstring doesn't seem accurate

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -66,126 +66,394 @@ async def remove_env(
...
+    return {
+        "LLM_API_BASE": api_base,
+        "LLM_API_KEY": api_key,
+        "LLM_MODEL": selected_model,
+        **extra_config,
+    }
+
+
+async def _configure_embedding(env: dict[str, str]) -> dict[str, str] | None:
+    """Configure embedding provider and return (provider_name, api_base, api_key, model, extra_config)"""
```
JanPokorny: Docstring doesn't seem accurate

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -199,279 +467,151 @@ async def setup(
...
-                console.print(f"[red]Error setting up Ollama model: {e!s}[/red]")
-                return False
-
-            selected_model = modified_model
-
-        try:
-            with console.status("Checking if the model works...", spinner="dots"):
+    # Load available models
+    try:
+        if provider_name in ["Anthropic", "watsonx"]:
```
JanPokorny: Anthropic does not support embeddings

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -199,279 +467,151 @@ async def setup(
...
-                        return False
+            )
+            else (
+                await inquirer.fuzzy(
+                    message="Select an embedding model (type to search):",
+                    choices=sorted(available_models),
+                ).execute_async()
+                if available_models and len(available_models) > 1
+                else available_models[0]
+                if available_models and len(available_models) == 1
```
JanPokorny: Looking back, we should probably show the list even when there is just one model, since now that model might not be appropriate for embedding (and same the other way around with LLM setup above)

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -199,279 +467,151 @@ async def setup(
...
+
+    return {
+        "EMBEDDING_API_BASE": api_base,
+        "EMBEDDING_API_KEY": api_key,
+        "EMBEDDING_MODEL": selected_model,
+        **extra_config,
+    }
+
+
+@app.command("setup", help="Interactive setup for LLM and embedding provider environment variables")
```
JanPokorny: It grabs the docstring so `help=` is not needed

apps/beeai-server/src/beeai_server/api/routes/embeddings.py
```diff
@@ -25,18 +24,18 @@ class EmbeddingsRequest(pydantic.BaseModel):
 
     model: str
     input: list[str] | str
-    encoding_format: Literal["float"]
+    encoding_format: str | None = None
```
JanPokorny: You probably want this

```suggestion
    encoding_format: Literal["float"] = "float"
```

Watsonx does not support the base64 format and I did not want to bother polyfilling it since afaik it's not used in our agents, I'd keep the arg like this in the meantime so that the error message is clear when somebody tries to use base64

apps/beeai-cli/src/beeai_cli/commands/env.py
```diff
@@ -199,279 +467,151 @@ async def setup(
...
+                [_ollama_exe(), "pull", selected_model],
+                "Pulling the selected embedding model",
+                check=True,
+            )
         except Exception as e:
-            console.print(format_error("Error", f"Error during model test: {e!s}"))
+            console.print(f"[red]Error while pulling embedding model: {e!s}[/red]")
+            return None
+
+    return {
```
JanPokorny, line 598: We could also do a little test of the provider here, similar to how we test LLMs

apps/beeai-cli/src/beeai_cli/commands/env.py
jezekra1: It's returning False only if you said yes to "Do you want to configure an embedding provider?" and then the setup failed. I think it's best if it explicitly fails then, next time you will say "no" or do you think we should still move forward in this case?

apps/beeai-cli/src/beeai_cli/commands/env.py
jezekra1: there is no sorted call, that's for available models only

apps/beeai-server/src/beeai_server/api/routes/embeddings.py
jezekra1: voyage supports only "base64" here or `null` (not float):
https://docs.voyageai.com/reference/embeddings-api

so it needs to be at least nullable, but I wouldn't make opinions here, the error message should be clear from the embedding backend - maybe you know about it and you designed your agents that way

apps/beeai-cli/src/beeai_cli/commands/env.py
jezekra1: leaving this as potential future improvement :D 
I think it's rare and this will be refactored soon altogether

apps/beeai-cli/src/beeai_cli/commands/env.py
JanPokorny: Yeah in that case it's probably fine

apps/beeai-server/src/beeai_server/api/routes/embeddings.py
JanPokorny: I don't want to "leak" the underlying backend that much, plus it's confusing if agent works with one configuration of the proxy and not the other, so I prefer unifying the interface around common stuff that works. Imagine if someone builds their agent against OpenAI and then someone else tries to use it using Watsonx, they would have no idea why it does not work.

OK, Voyage is weird, in that case we need a little change to map `"float" -> None` to make the call universal, but I'd still keep `"float" | None` as the only allowed values.

jezekra1: ok makes sense, added the remap

PR #945, opened 2025-07-18
docs: rag

jenna-winkler:


docs/concepts/rag.mdx
```diff
@@ -0,0 +1,205 @@
...
+
+#### 3. Vector Search Tool
+
+```python
+from beeai_framework.tools import Tool
+
+class VectorSearchTool(Tool):
+    """Tool for searching through indexed documents."""
+    
+    def __init__(self, vector_store_id: str):
```
jezekra1, line 133: The vector search tool cannot be simplified this way

```
    name = "VectorSearch"
    description = (
        "Search for relevant documents and information from uploaded files using semantic search. "
        "This tool finds the most relevant content based on meaning rather than exact keyword matching."
    )
```

descriptions are crucial and I think that the Input/Output schemas are also required (they tell the LLM how to use the tool)

jenna-winkler:
will wait until SDK is available

PR #946, opened 2025-07-21
refactor(ui): rework ui messages

PetrBulanek:


apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -99,29 +92,27 @@ function AgentRunProvider({ agent, children }: PropsWithChildren<Props>) {
       }
 
       updateLastAgentMessage((message) => {
-        message.content = applyContentTransforms({
-          rawContent: message.rawContent,
-          transforms: message.contentTransforms,
-        });
+        message.parts = sortMessageParts(message.parts);
+        message.content = processMessageContent(message);
```
kapetr: I’m wondering if rendering the final content should be the responsibility of a View component, rather than being stored in `UIMessage`. The `content` duplicates the data in `parts` and could be constructed from them when needed. 
But I understand it's a convenient optimization. At minimum, we would need to add an indicator of a visible content present (`hasContent`) to the `UIMessage`, because it is checked in a few places.

apps/beeai-ui/src/modules/runs/hands-off/HandsOffOutputView.tsx
```diff
@@ -19,7 +19,7 @@ export function HandsOffOutputView() {
   const { agent, input, isPending, cancel, clear } = useAgentRun();
   const { messages } = useMessages();
   const message = messages.find(isAgentMessage);
-  const hasOutput = Boolean(message?.content);
+  const hasOutput = message ? Boolean(message.content) : false;
```
kapetr: Any reason for the update? Just wondering, the result seems the same...

apps/beeai-ui/src/modules/trajectories/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+    id: uuid(),
+    message: message ?? undefined,
+    toolName: tool_name ?? undefined,
+  };
+
+  return part;
+}
+
+export function hasViewableTrajectoryParts(trajectory: UITrajectoryPart) {
+  const nonViewableProperties = ['kind', 'id'] as NonViewableTrajectoryProperty[];
```
kapetr: This could be a constant outside the function.

kapetr (approved):

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
PetrBulanek: You re right. In the original code I didn't have `content` at all and I always built it on demand using getMessageContent directly in the UI components. It was in about five places, so then I was like, I don't totally like that it's always being called again, so yeah, this is more of an optimization.

What do you mean about adding`hasContent`? Right now the `content` is rebuilt for every message Part that comes in the stream, so it should always be up to date. Maybe I just missed something. :-)

apps/beeai-ui/src/modules/runs/hands-off/HandsOffOutputView.tsx
PetrBulanek: I guess it's a by-product of refactoring, when I probably needed to edit it for something, but it doesn't make sense now. I'm reverting it back.

PR #958, opened 2025-07-22
feat(ui): align agent ui fields, add author and contributors

### LINKED ISSUE #682: Add author email metadata to GUI
## Description 

Today, we only show author name, but there is also the ability for developers to define email and url, our GUI should support these additional optional fields if filled out. 
  
 ```
  author={
            "name": "John Smith",
            "email": "jsmith@example.com"
        },
```

## Designs

[Figma](https://www.figma.com/design/hNNBTrcvQ6UIQQuxZAR3HX/BeeAI-dev-ready--external-?node-id=998-4601&p=f&t=9a5IvOdp1E7BZXjb-0)

### PR DESCRIPTION
**Pull request by kapetr**
Closes #682 #700

### DIFF
--- a/apps/beeai-ui/src/modules/agents/api/types.ts
+++ b/apps/beeai-ui/src/modules/agents/api/types.ts
@@ -12,7 +12,7 @@ export interface Agent extends Omit<AgentCard, 'provider'> {
   provider: Omit<Provider, 'agent_card'> & {
     metadata?: AgentCardProvider;
   };
-  ui: UiExtensionParams;
+  ui: UIExtensionParams;
 }
 
 export type AgentExtension = NonNullable<Agent['capabilities']['extensions']>[number];
@@ -27,7 +27,7 @@ export interface AgentTool {
   description: string;
 }
 
-export interface UiExtensionParams {
+export interface UIExtensionParams {
   ui_type?: UiType;
   user_greeting?: string;
   display_name: string;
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: apps/beeai-ui/src/modules/agents/api/types.ts [hunk 1]
```diff
@@ -53,9 +50,25 @@ export interface UiExtension extends AgentExtension {
...
+export interface AgentAuthor {
+  name?: string;
+  email?: string;
+}
+
+export interface AgentContributor extends AgentAuthor {
+  name?: string;
+  email?: string;
+  url?: string;
+}
```
PetrBulanek: This should probably be handled/validated by the SDK, but I think at least the name should be mandatory.

apps/beeai-ui/src/modules/agents/utils.ts
```diff
@@ -45,27 +38,19 @@ export function isAgentUiSupported(agent: Agent) {
   return ui_type && SupportedUis.includes(ui_type);
 }
 
-// TODO: a2a
-type AgentLinkType = 'homepage' | 'documentation' | 'source-code';
-
 export function getAvailableAgentLinkUrl<T extends AgentLinkType | AgentLinkType[]>(
-  metadata: AgentMetadata,
+  links: UiExtensionParams['links'],
```
PetrBulanek: I prefixed the `ui` types with uppercase `UI`, so I'd probably do the same thing here. :-)

apps/beeai-ui/src/modules/agents/api/types.ts
kapetr: It's not validated at the moment, so I wanted to cover all possible data. There's even a fallback to display the email, if the name is not provided, which seemed reasonable to me - for the current stat of things. But let me know, I get your point, I guess we could (and maybe should) be more strict.

apps/beeai-ui/src/modules/agents/utils.ts
kapetr: Ah of course :)

PR #960, opened 2025-07-23
feat: decoupling of a2a sdk from the UI layer

tomkis:


apps/beeai-ui/src/api/a2a/client.ts
```diff
@@ -0,0 +1,94 @@
...
+    return [];
+  }
+
+  const parts = message.parts
+    .map((part) => {
+      const processedPart = match(part)
+        .with({ kind: 'text' }, (part) => processTextPart(part, message.messageId))
+        .with({ kind: 'file' }, processFilePart)
+        .otherwise((otherPart) => {
+          throw new Error(`Unsupported part - ${otherPart.kind}`);
```
PetrBulanek, line 35: I am not sure, we should throw here as it could kill the whole stream. Maybe just log a warning?

apps/beeai-ui/src/api/a2a/part-processors.ts
```diff
@@ -0,0 +1,67 @@
...
+    const trajectoryPart = createTrajectoryPart(trajectory);
+
+    return trajectoryPart;
+  }
+
+  const citation = extractCitation(metadata);
+
+  if (citation) {
+    if (text !== '') {
+      throw new Error('Text part should be empty when citation is present');
```
PetrBulanek, line 39: Is this correct? I guess the text could be theoretically non empty and be valid.

apps/beeai-ui/src/api/a2a/client.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      for await (const event of stream) {
+        match(event).with({ kind: 'status-update' }, (event) => {
+          const messageParts = handleStatusUpdate(event);
+
+          messageSubject.next(messageParts);
+        });
+      }
+
+      messageSubject.complete();
```
PetrBulanek, line 71: Should we wrap this in a try-catch and handle possible error?

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -168,53 +90,102 @@ function AgentRunProvider({ agent, children }: PropsWithChildren<Props>) {
...
+            updateLastAgentMessage((message) => {
+              match(part)
+                .with({ kind: UIMessagePartKind.File }, (part) => {
+                  const transformedPart = transformFilePart(part, message);
+
+                  if (transformedPart) {
+                    message.parts.push(transformedPart);
+                  } else {
+                    message.parts.push(part);
+                  }
```
PetrBulanek, line 161: Shouldn't we always return the original ui part here as well?

apps/beeai-ui/src/modules/runs/hooks/useRunAgent.ts
```diff
@@ -1,207 +0,0 @@
...
-          break;
-        case 'failed':
-          handleFailed(event);
-
-          break;
-        case 'rejected':
-          handleFailed(event);
-
-          break;
-      }
```
PetrBulanek, line 131: We are not handling this cases anywhere, which is probably wrong. We should at least handle the error states (failed, rejected).

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -168,53 +90,105 @@ function AgentRunProvider({ agent, children }: PropsWithChildren<Props>) {
...
-    stopAgent();
-  }, [stopAgent]);
+  const cancel = useCallback(async () => {
+    if (pendingRun.current && pendingSubscription.current) {
+      updateLastAgentMessage((message) => {
+        message.status = UIMessageStatus.Aborted;
+      });
+
+      pendingSubscription.current();
+      await pendingRun.current.cancel();
```
kapetr, line 100: Wouldn't it make sense to cancel all subscriptions automatically in `cancel` method of `ChatRun`?

kapetr:
@tomkis Should we start handling `update_kind` to display only `final_answer` on the UI, or is that still a subject to change? Currently we are still displaying all the `text` messages including `thought`, etc.

kapetr (approved):

PetrBulanek (approved):

Issue #896, opened 2025-07-07
Introduce Beeai SDK that simplifies wrapping of agents (A2A)


tomkis:
Things to consider
- A2A doees not support multi agent per server unlike ACP
- Skills in A2A are not logical units that would allow to route the logic and hence effectively simulating multi agents
- One of the important questions is how opinionated we want to be, for instance; do we want to keep the messages as we used to do with ACP or do we want to completely switch over to A2A Messages?
- How to leverage A2A extensions to implement custom extensions that have been introduced with ACP

PR #962, opened 2025-07-23
High level a2a sdk

jezekra1:
Ref: #896


apps/beeai-sdk/src/beeai_sdk/server/utils.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import asyncio
+from asyncio import CancelledError
+from contextlib import suppress
+
+
+async def cancel_task(task: asyncio.Task[None] | None):
```
pilartomas: nitpick: it is strange to accept non in `cancel_task` function.

apps/beeai-sdk/src/beeai_sdk/server/agent.py
```diff
@@ -0,0 +1,358 @@
...
+                            state=TaskState.completed
+                            | TaskState.failed
+                            | TaskState.canceled
+                            | TaskState.auth_required
+                            | TaskState.input_required
+                        )
+                    ):
+                        break
+        except Exception as ex:
+            print(ex)
```
pilartomas: log and fail the task

apps/beeai-sdk/src/beeai_sdk/server/agent.py
```diff
@@ -0,0 +1,358 @@
...
+        except CancelledError:
+            await TaskUpdater(event_queue, task_id=context.task_id, context_id=context.context_id).cancel()
```
pilartomas: Shouldn't be Cancelled error caught first? will the execution ever reach this?

apps/beeai-sdk/src/beeai_sdk/server/agent.py
```diff
@@ -0,0 +1,358 @@
...
+                    case TaskStatusUpdateEvent(
+                        status=TaskStatus(
+                            state=TaskState.completed
+                            | TaskState.failed
+                            | TaskState.canceled
+                            | TaskState.auth_required
+                            | TaskState.input_required
+                        )
+                    ):
+                        break
```
pilartomas: What about artifact-related events?

apps/beeai-sdk/src/beeai_sdk/server/agent.py
```diff
@@ -0,0 +1,358 @@
...
+
+    return decorator
+
+
+class Executor(AgentExecutor):
+    def __init__(self, execute_fn: AgentFunction, queue_manager: QueueManager) -> None:
+        self._execute_fn = execute_fn
+        self._queue_manager = queue_manager
+        self._running_tasks: dict[str, asyncio.Task] = {}
+        self._cancel_queues: dict[str, EventQueue] = {}
```
pilartomas, line 199: I was also thinking about an expiration period (good for hosted solutions) but that isn't suitable for all types of agents I suppose. Maybe optional based on env var? Agent would simply fail after it expires.

apps/beeai-sdk/src/beeai_sdk/server/app.py
```diff
@@ -0,0 +1,59 @@
...
+        **kwargs,
+    )
+
+    app.add_middleware(
+        CORSMiddleware,
+        allow_origins=["https://beeai.dev"],
+        allow_methods=["*"],
+        allow_headers=["*"],
+        allow_credentials=True,
+    )
```
pilartomas: Would be nice to get rid of this if we can.

apps/beeai-sdk/src/beeai_sdk/server/agent.py
jezekra1: Artifact events should never be final in this implementation
(a TaskUpdate failed/canceled/completed event will always follow).

But I can use `TaskStatusUpdateEvent.final` - that's actually more appropriate than listing all states individually

apps/beeai-sdk/src/beeai_sdk/server/agent.py
jezekra1: let's leave it for a separate task, could be tricky to manage lifecycle of all the objects and queues with expiration

apps/beeai-sdk/src/beeai_sdk/server/app.py
jezekra1: I can remove it but CORS is required to make it work with our UI - we need to add some config anyway

pilartomas: Next-based UI should proxy the request, rendering CORS unnecessary.

pilartomas: ACP had it not for the UI but for mintlify docs page.

apps/beeai-sdk/src/beeai_sdk/server/logging.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import logging
+
+from uvicorn.logging import DefaultFormatter
+
+logger = logging.getLogger("acp")
```
pilartomas: value

apps/beeai-sdk/src/beeai_sdk/server/agent.py
jezekra1: There is a clause in `_run_generator`:
```
        except CancelledError:
            await task_updater.cancel()
```
which handles explicit "cancel" call to the agent through the queue.

But I found that during streaming, the actual "execute function" task is also cancelled ([src](https://github.com/a2aproject/a2a-python/blob/f22b384d919e349be8d275c8f44bd760d627bcb9/src/a2a/server/request_handlers/default_request_handler.py#L149))

So because of this the event is never propagated from the agent through the long_running_event_queue.
I'll add a coment explaining this

This clause is hit by `test_run_cancel_stream`

apps/beeai-sdk/src/beeai_sdk/server/app.py
jezekra1: @PetrBulanek @kapetr can you chime in why was the CORS proxy required? I remember it was in the example agents

apps/beeai-sdk/examples/awaiting_agent.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from collections.abc import AsyncGenerator
+
+from a2a.client import create_text_message_object
+from a2a.types import Message, TaskState, TaskStatus
```
JanPokorny, line 7: It seems a bit confusing that "TaskStatus" and "TaskState" are two different things, can we come up with a better naming?

apps/beeai-sdk/examples/cli.py
```diff
@@ -175,47 +180,48 @@ async def cli(base_url: str, context_id: str, history: bool) -> None:
...
-
-            if history and task_id:
-                print("========= history ======== ")
-                task_response = await client.get_task(
-                    a2a.types.GetTaskRequest(
-                        id=str(uuid.uuid4()), params=a2a.types.TaskQueryParams(id=task_id, history_length=10)
-                    )
-                )
-                print(task_response.model_dump_json(include={"result": {"history": True}}))
+    # TODO: dead code?
```
JanPokorny: Yeah I think for now you can delete this plus all the unused stuff above (+ the --history cli arg), we can always add this functionality when we have the need. It's mostly just copypasted from the official CLI which is quite verbose.

apps/beeai-sdk/examples/awaiting_agent.py
```diff
@@ -0,0 +1,26 @@
...
+
+from beeai_sdk.server.server import Server
```
JanPokorny: Importing from `server.server` seems ugly, can we rename the module to `backend` or similar?

apps/beeai-sdk/src/beeai_sdk/server/app.py
JanPokorny: This was so that the mintlify-based docs on beeai.dev could send requests to localhost agents

apps/beeai-sdk/src/beeai_sdk/version.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+from importlib.metadata import version
+
+__version__ = version("beeai-sdk")
```
JanPokorny: Don't we want to have this in `__init__.py` so it's `beeai_sdk.__version__`, not `beeai_sdk.version.__version__`?

apps/beeai-sdk/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
```
JanPokorny: This should not be here

apps/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
```
JanPokorny: This should not be here

apps/beeai-server/src/beeai_server/api/routes/settings.py
```diff
@@ -0,0 +1,3 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
```
JanPokorny: Created by mistake?

apps/beeai-sdk/pyproject.toml
```diff
@@ -24,32 +31,40 @@ build-backend = "uv_build"
...
+asyncio_mode = "auto"
 
 [dependency-groups]
 dev = [
     "pyright>=1.1.403",
+    "pytest>=8.4.1",
+    "pytest-asyncio>=1.1.0",
+    "pytest-httpx>=0.35.0",
     "ruff>=0.12.3",
 ]
```
JanPokorny: Bit unrelated but could you move this up, right below dependencies? It's now mixed down here among the tool configs

apps/beeai-sdk/tests/e2e/config.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+
+class Config:
+    PORT = 8000
```
JanPokorny: Unused file

JanPokorny (changes requested):

apps/beeai-sdk/src/beeai_sdk/server/app.py
pilartomas: It is a common practice in the architecture we have. It technically isn't "cors" proxy since it resides on the same origin as the web but avoiding cors issues is usually one of it's primary purposes (another would be auth).

jezekra1: Found out the issue - the a2a JS sdk is always fetching the card in constructor:
https://github.com/a2aproject/a2a-js/blob/main/src/client/client.ts#L56

In python the client can be constructed directly from the card (obtained by other means - fetch /api/v1/providers) and does not re-fetch it again.

Currently direct agent card request is proxied to the source agent - I will change it on the server to return the static card with patched url - but this is a potential for contribution to A2A:
- Side-effects in constructor are very wierd
- No way to instantiate client without refetch of agent card
- Not handling custom agent card urls (we assume the same, but we are not official a2a client)

apps/beeai-sdk/examples/awaiting_agent.py
jezekra1: I don't think we should rename the underlying structures, but also I kind of already introduced `ArtifactChunk` to clearly mark chunks from whole artifacts.

apps/beeai-sdk/src/beeai_sdk/server/utils.py
jezekra1: the typical usage pattern is:
```
task = None
try:
    task = asyncio.create_task(...)
finally:
    cancel_task(task)
```
I'd like to avoid the extra `None` check everytime it's used, I'll just rename it to `cancel_task_if_exists`

apps/beeai-sdk/examples/awaiting_agent.py
jezekra1: I'd rather stick to server/client terminology. I'll reexport it

apps/beeai-sdk/examples/awaiting_agent.py
JanPokorny: Oh right that's directly from A2A, did not notice

apps/beeai-sdk/__init__.py
JanPokorny: @jezekra1 You still need to delete this one 😁

PR #970, opened 2025-07-25
feat(beeai-sdk): port citation and trajectory from ACP

JanPokorny:


apps/beeai-sdk/src/beeai_sdk/a2a_extensions/ui/trajectory.py
```diff
@@ -0,0 +1,57 @@
...
+    - message: A reasoning step or thought in the agent's decision process.
+    - tool_name: Name of the tool that was executed.
+    - tool_input: Input parameters passed to the tool.
+    - tool_output: Output or result returned by the tool.
+    """
+
+    message: str | None = None
+    tool_name: str | None = None
+    tool_input: str | None = None
+    tool_output: str | None = None
```
JanPokorny: Maybe we want to make this more general? Especially since "Trajectory" has nothing to do with tool calling, referring to "tool" explicitly seems unnecessary.

I propose switching to `title` and `content` only, with `content` being Markdown-enabled, which would result in an accordion component like this:

Image

tomkis: @michael-desmond what's your take on this?

jezekra1 (approved):

apps/beeai-sdk/src/beeai_sdk/a2a_extensions/ui/trajectory.py
JanPokorny: I updated the PR to reflect what I had in mind

PR #965, opened 2025-07-24
feat(beeai-sdk): implement agent detail extension

JanPokorny:

PR #972, opened 2025-07-28
feat(ui): sync agent details extension with beeai-sdk

kapetr:
Refs #965

### DIFF
--- a/agents/ui/hello/src/extensions.py
+++ b/agents/ui/hello/src/extensions.py
@@ -8,25 +8,28 @@ class BeeAIUITool(BaseModel):
     name: str
     description: str
 
+class AgentDetailsContributor(BaseModel):
+    name: str
+    email: str | None = None
+    url: str | None = None
+
 class BeeAIUI(AgentExtension):
-    def __init__(self, ui_type: str, user_greeting: str, display_name: str, tools: list[BeeAIUITool]):
+    def __init__(self, ui_type: str, user_greeting: str, tools: list[BeeAIUITool], framework: str, license: str, programming_language: str, 
+                 source_code_url: str, container_image_url: str | None = None, author: AgentDetailsContributor | None = None, 
+                 contributors: list[AgentDetailsContributor] | None = None, starter_prompts: list[str] |None = None ):
         super().__init__(
             uri="beeai_ui",
             params={
                 "ui_type": ui_type,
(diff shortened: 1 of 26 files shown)


agents/ui/hello/src/extensions.py
```diff
@@ -8,25 +8,28 @@ class BeeAIUITool(BaseModel):
     name: str
     description: str
 
+class AgentDetailsContributor(BaseModel):
```
PetrBulanek, line 11: I am not sure about the naming convention in Python, but shouldn't this be also prefixed with `BeeAI`, so maybe something like `BeeAIAgentContributor`?

apps/beeai-ui/src/modules/agents/api/types.ts
```diff
@@ -28,21 +28,18 @@ export interface AgentTool {
 }
 
 export interface UIExtensionParams {
-  ui_type?: UiType;
+  ui_type?: string;
   user_greeting?: string;
-  display_name: string;
```
PetrBulanek, line 33: How so? We are removing the display name?

apps/beeai-ui/src/modules/agents/api/types.ts
```diff
@@ -28,21 +28,18 @@ export interface AgentTool {
 }
 
 export interface UIExtensionParams {
-  ui_type?: UiType;
+  ui_type?: string;
```
PetrBulanek: I understand it can be any string, but don't we want to narrow it down to what is supported on the UI here? Possibly at least `UiType | string`?

PetrBulanek (changes requested):

apps/beeai-ui/src/modules/agents/api/types.ts
JanPokorny: Display name is part of core A2A agent card

agents/ui/hello/src/extensions.py
JanPokorny: @PetrBulanek No need to worry about code style here, we'll soon switch to the real SDK anyway

kapetr: It will be available under this name in beeai-sdk, see #965

PetrBulanek (approved):

PR #973, opened 2025-07-28
feat(ui): redesign messages and streaming from top

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/664


apps/beeai-ui/src/components/LineClampText/LineClampText.tsx
```diff
@@ -14,16 +16,28 @@ import classes from './LineClampText.module.scss';
...
+  buttonClassName,
+  as = 'span',
+  children,
+}: PropsWithChildren<Props>) {
   const id = useId();
   const textRef = useRef<HTMLElement>(null);
   const [isExpanded, setIsExpanded] = useState(false);
   const [showButton, setShowButton] = useState(false);
 
+  const Component = as;
```
kapetr: This seems like non-standard use of the `as` prop. I'd consider restricting its type to `'span' | 'div'` (and possibly naming it a bit differently?). Or could any other element type make sense here, given it's used for several elements even nested in each other?

apps/beeai-ui/src/components/LineClampText/LineClampText.tsx
```diff
@@ -74,23 +88,39 @@ export function LineClampText({ lines, className, buttonClassName, children }: P
...
-            {isExpanded ? 'View less' : 'View more'}
-          </ExpandButton>
-        </span>
+        <Component className={clsx(classes.button, buttonClassName)}>
+          {iconButton ? (
+            <IconButton
+              kind="tertiary"
+              onClick={() => setIsExpanded((state) => !state)}
+              aria-controls={id}
+              aria-expanded={isExpanded}
```
kapetr: Just a minor detail - these props are identical in both cases, it might make sense to extract them, or even move this into a helper component.

apps/beeai-ui/src/hooks/useIsScrolled.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  useEffect(() => {
+    const observeElement = observeElementRef.current;
+
+    const observer = new IntersectionObserver(
+      ([entry]) => {
+        setIsScrolled(!entry.isIntersecting);
+      },
+      { root: scrollElementRef.current },
+    );
```
kapetr: Maybe `useIntersectionObserver` could be used?

apps/beeai-ui/src/modules/runs/components/RunSubmit.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { useAgentRun } from '../contexts/agent-run';
+
+interface Props {
+  disabled?: boolean;
+}
+
+export function RunSubmit({ disabled }: Props) {
+  const { isPending, cancel } = useAgentRun();
+  const { isPending: isFileUploadPending } = useFileUpload();
```
kapetr: I'd consider decoupling this component from contexts. It seems a bit confusing that disabled comes as prop, but actually depends on context props used here.

apps/beeai-ui/src/components/LineClampText/LineClampText.tsx
PetrBulanek: Agree. I suggest using it as a boolean `useBlockElement`. What do you think?

kapetr: Sounds good!

kapetr (approved):

PR #979, opened 2025-07-29
feat(chat-agent): requirement agent + file support

aleskalfas:


agents/official/beeai-framework/chat/src/chat/tools/files/file_generator.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class FileGeneratorTool(
+    Tool[FileGeneratorInput, ToolRunOptions, FileGeneratorToolOutput]  # type: ignore
+):
+    """
+    Creates a new file and writes the provided content into it.
+    """
+
+    name: str = "FileGenerator"
```
jezekra1: wouldn't `FileWriter` be a better name for the LLM?

agents/official/beeai-framework/chat/src/chat/tools/files/utils.py
```diff
@@ -0,0 +1,175 @@
...
+
+    # smallest positive integer that isn’t taken
+    i = 1
+    while i in used:
+        i += 1
+
+    return f"{base}({i}){ext}"
+
+
+def is_base64(s: str) -> bool:
```
jezekra1: unused function

agents/official/beeai-framework/chat/src/chat/agent.py
```diff
@@ -124,49 +149,100 @@ async def chat(message: Message, context: Context):
...
-        ChatModelParameters(temperature=0),
+    extracted_files = await extract_files(
+        history=messages[context.context_id], incoming_message=message
     )
+    input = to_framework_message(message)
 
     # Configure tools
-    tools: list[AnyTool] = [
+    tools = [
+        ActTool(),
```
jezekra1: can you clarify what is this tool for? It's a tool called before other tools are called?

agents/official/beeai-framework/chat/src/chat/tools/files/file_generator.py
aleskalfas: The idea is that this tool does just one atomic operation creates/generates file. In the future there will be an another tool (etc. `FileEditor`) that will just update existing files. The goal is to provide the most simple tools that will be easy to use by small LLMs. Maybe the best name would be `FileCreator` (resp. `FileUpdater`).

aleskalfas: Fixed

agents/official/beeai-framework/chat/src/chat/tools/files/utils.py
aleskalfas: removed

agents/official/beeai-framework/chat/src/chat/agent.py
aleskalfas: I've added some comments

PR #1009, opened 2025-08-04
feat(beeai-sdk): implement platform client

JanPokorny:


apps/beeai-sdk/src/beeai_sdk/platform/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from functools import cache
+
+import httpx
+
+
+@cache
+def get_client() -> httpx.AsyncClient:
+    return httpx.AsyncClient(base_url="http://127.0.0.1:8333")
```
jezekra1: we should probably get `PLATFORM_URL` from env here or have a nice way to initialize the client in agents without having to pass custom client to every method

PR #1016, opened 2025-08-06
feat(ui): add artifact update parts processing

PetrBulanek:


apps/beeai-ui/src/api/a2a/client.ts
```diff
@@ -67,12 +60,12 @@ export const buildA2AClient = (providerId: string) => {
           .with({ kind: 'task' }, (task) => {
             taskId = task.id;
           })
-          .with({ kind: 'status-update' }, (event) => {
+          .with({ kind: 'status-update' }, { kind: 'artifact-update' }, (event) => {
```
tomkis: Cant we split this into to `with`s then you dont even need `isArtifactUpdateEvent` and can handle each case separately.

PR #1017, opened 2025-08-06
issue-767

tedhabeck:
Add conditional enable/disable of OIDC flows and endpoint protection controlled on platform startup by a .env var in beeai-cli `BEEAI__OIDC_ENABLED=False`  (Default is False).

Full details for turning on OIDC are in the CONTRIBUTING.md of this branch.

This has been tested with both w3id and an IBMiD SSO provision.


helm/crds/certificate.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: cert-manager.io/v1
```
jezekra1, line 1: Isn't there a way to add istio using subchart?

https://istio.io/latest/docs/setup/install/helm/

We cannot add this to the helm chart without a feature flag, some deployments don't allow CRD creation, istio must be optional, please add optional subchart or `istio.enabled` to values and make this configurable.

Consider:
Ideally we wouldn't have this in our helm chart at all if we're only using the ambient mode due to the range of permission related issues or conflicts this might cause. (what if the user already has istio in their system?)

I'd consider adding a guide how to set up internal TLS with istio ambient mode instead of baking it into our chart, but it's probably okay if it's a single toggle that's off by default

helm/crds/certificate.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: Certificate
+metadata:
+  name: beeai-platform-tls
+  namespace: istio-system
```
jezekra1, line 5: users may not have access to this namespace or the option to create new namespaces

helm/templates/ui/deployment.yaml
```diff
@@ -61,6 +61,29 @@ spec:
           env:
             - name: API_URL
               value: "http://beeai-platform-svc:{{.Values.service.port}}"
+          {{- if .Values.oidc.enabled }}
+            - name: NEXTAUTH_URL
+              value: {{ .Values.oidc.nextauth_url }}
+            - name: NEXTAUTH_REDIRECT_PROXY_URL
+              value: {{ .Values.oidc.nextauth_redirect_proxy_url }}
+            - name: NEXTAUTH_IBM_ISSUER
```
jezekra1, line 69: why is it called IBM? Is there a reason we don't support other oidc providers?

tedhabeck: Good point, we do support any OIDC provider, will rename, thanks!

apps/beeai-server/src/beeai_server/auth/utils.py
```diff
@@ -0,0 +1,91 @@
...
+if not configuration.oidc.disable_oidc:
+    audience = configuration.oidc.client_id
+    JWKS_URL = configuration.oidc.jwks_url
+    JWKS_FOLDER = "/tmp/jwks"
+    JWKS_FILE = os.path.join(JWKS_FOLDER, "pubkeys.json")
+
+    os.makedirs(JWKS_FOLDER, exist_ok=True)
+
+    try:
+        response = requests.get(JWKS_URL)
```
jezekra1, line 31: please don't do network requests on import - we have `bootstrap.py` for initialization

apps/beeai-server/src/beeai_server/__init__.py
```diff
@@ -29,6 +43,26 @@ def serve():
         logger.error("Native windows is not supported, use WSL")
         return
 
+    # Download the public jwk key set (jwks)
+    if JWKS_URL is not None:
```
jezekra1, line 47: please use bootstrap.py for bootstrapping

apps/beeai-server/src/beeai_server/api/routes/acp.py
```diff
@@ -1,3 +1,119 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
+import fastapi
```
jezekra1, line 4: remove this file - acp was deprecated

apps/beeai-server/src/beeai_server/auth/models.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+User object with admin attribute controlled by bluegroup membership.
+
+"""
+
+
+class AuthenticatedUser:
```
jezekra1: [nitpick]: use pydantic models or the existing `User` model

apps/beeai-server/src/beeai_server/auth/models.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self, uid: str, is_admin: bool, display_name: str, email: str) -> None:
+        self.username = uid
+        self.emailAddress = email
```
jezekra1, line 13: [nitpick]: snake_case

apps/beeai-server/src/beeai_server/__init__.py
```diff
@@ -20,6 +20,20 @@
 
 logger = logging.getLogger(__name__)
 
+configuration: Configuration = get_configuration()
```
jezekra1, line 23: these don't need to be public

helm/values.yaml
```diff
@@ -73,13 +75,27 @@ securityContext: { }
...
+
 # This is for setting up a service more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/
 service:
   # This sets the service type more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
   type: ClusterIP
   # This sets the ports more information can be found here: https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports
+  portName: http
   port: 8333
-  nodePort: 8333
+  # nodePort: 8333
```
jezekra1, line 97: why this change?

helm/values.yaml
```diff
@@ -113,27 +129,48 @@ livenessProbe:
     port: http
   initialDelaySeconds: 5
   failureThreshold: 5
-  periodSeconds: 5
+  periodSeconds: 30
```
jezekra1, line 132: can you explain why bump this?

helm/templates/pvc.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: v1
+metadata:
+  name: beeai-jwks
+  namespace: default
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
jezekra1, line 11: this is only a cache, do we need to store it across restarts?

also we probably don't need 1GB for this

helm/values.yaml
```diff
@@ -113,27 +120,48 @@ livenessProbe:
...
     path: /healthcheck
     port: http
   initialDelaySeconds: 5
   failureThreshold: 20
-  periodSeconds: 2
+  periodSeconds: 30
 
 # Additional volumes on the output Deployment definition.
-volumes: [ ]
+volumes:
```
jezekra1, line 142: Why do we need this? I thought istio in ambient mode shouldn't require any changes to deployments

helm/values.yaml
tedhabeck: Yes, a 5 second period would mean the check runs every 5 seconds, which is a waste of compute resource.  A 30 second interval should be more than adequate for the liveness / readiness probes.

jezekra1: That's just a ping request, that costs nothing.
If you set it to 30 seconds the users can wait for up to 30 seconds longer for the platform to come up (there is a wait for available status on all deployments which depends on the readiness probe). Or we'd have to nail the initial delay precisely, but that also depends on the user hardware.

Default kubernetes value is [10s](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes), but I'd say 5s is reasonable if it speeds up platform startup time just a little bit.

helm/crds/certificate.yaml
tedhabeck: - OK, I can install an set it up with helm. However, istio needs to be installed in a separate namespace, so making it a dependency of the beeai-platform chart won't work (A chart and it's dependencies are assumed to be deployed into a single namespace).
- Re: `it's probably okay if it's a single toggle that's off by default`, agreed.

apps/beeai-cli/src/beeai_cli/token_store.py
```diff
@@ -0,0 +1,33 @@
...
+from pathlib import Path
+
+import aiofiles
+
+TOKEN_FILE = Path.home() / ".beeai" / "token.json"
+
+
+async def save_token(token: str):
+    TOKEN_FILE.parent.mkdir(parents=True, exist_ok=True)
+    async with aiofiles.open(TOKEN_FILE, "w") as f:
```
jezekra1, line 15: we don't need aiofiles, there is already `anyio` available

apps/beeai-cli/src/beeai_cli/token_store.py
```diff
@@ -0,0 +1,33 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import os
+from pathlib import Path
+
+import aiofiles
+
+TOKEN_FILE = Path.home() / ".beeai" / "token.json"
```
jezekra1, line 10: These paths are stored and configurable in `configuration.py`:
https://github.com/i-am-bee/beeai/blob/main/apps/beeai-cli/src/beeai_cli/configuration.py#L29
```
    home: pathlib.Path = pathlib.Path.home() / ".beeai"
```

apps/beeai-cli/pyproject.toml
```diff
@@ -21,6 +21,7 @@ dependencies = [
     "tenacity~=9.1.2",
     "typer~=0.16.0",
     "pyyaml>=6.0.2",
+    "aiofiles"
```
jezekra1, line 24: remove this package

jezekra1:
Is this CLI flow accurate?

Have you considered using the [oauth device flow](https://docs.verify.ibm.com/ibm-security-verify-access/docs/tasks-deviceflow)? It is supported by all major oauth providers

image

apps/beeai-server/src/beeai_server/auth/utils.py
```diff
@@ -0,0 +1,91 @@
...
+from beeai_server.configuration import get_configuration
+
+logger = logging.getLogger(__name__)
+
+configuration: ConfigurationDependency = get_configuration()
+
+
+# Initialize JWKS only if OIDC is enabled
+if not configuration.oidc.disable_oidc:
+    audience = configuration.oidc.client_id
```
pilartomas, line 23: The platform should act as oAuth resource server, it is not a client and hence the audience should not be the `client_id`.

apps/beeai-cli/src/beeai_cli/api.py
```diff
@@ -50,29 +51,48 @@ def server_process_status(
     return ProcessStatus.not_running
 
 
+async def set_auth_header():
+    token = await load_token()
+    if not token:
+        raise RuntimeError("No token found. Please run `beeai login` first.")
+    if not token.get("id_token"):
+        raise RuntimeError("No id token found. Please run `beeai login` first.")
+    return f"Bearer {token['id_token']}"
```
pilartomas, line 60: This is not the appropriate token to send, it must be the `access_token`. The server is not the `id_token`'s audience. The server must use the `access_token` to grant access to resources.

If verified identity data is required by the resource server, the token should have `id_token` scope which allows the server to pick up the data on the auth server's `/userinfo` endpoint.

If unverified identity data is sufficient (e.g. user can change their name freely), the client (CLI/UI) should extract the data from the `id_token` and send them as regular request payload. This is fine since `sub/email` for authorization is still taken from the `access_token`.

apps/beeai-server/src/beeai_server/auth/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    JWKS_URL = configuration.oidc.jwks_url
+    JWKS_FOLDER = "/tmp/jwks"
+    JWKS_FILE = os.path.join(JWKS_FOLDER, "pubkeys.json")
```
pilartomas, line 26: The keys might rotate sometimes. General approach is to:

1. bootstrap the discovery from auth server's url (use well-known documents)
2. cache the jwks keys but not indefinitely - see https://github.com/auth0/node-jwks-rsa/tree/master

Point 1. can be usually skipped when working with JWTs, not sure if we need to support opaque tokens. However, it might be necessary to grab the user info with the access token anyway.

apps/beeai-server/src/beeai_server/auth/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    jwks = jwks or jwks_dict
+    aud = aud or audience
+    logger.debug("aud=%s", str(aud))
+    logger.debug("audience=%s", str(audience))
+    # Decode JWT using keys from JWKS
+    for pub_key in jwks.get("keys", []):
+        try:
+            obj_key = PyJWK(pub_key)
+            # explicitly only check exp and iat, nbf (not before time) is not included in w3id
+            claims = jwt.decode(token, obj_key, algorithms=["RS256"], options=None, verify=True, audience=aud)
```
pilartomas, line 81: It is also common to check the issuer.

pilartomas:
> Have you considered using the [oauth device flow](https://docs.verify.ibm.com/ibm-security-verify-access/docs/tasks-deviceflow)? It is supported by all major oauth providers

The common approach, as far as I know, is to use the usual authorization code flow if browser can be expected on the system. The final redirect uri is back to the endpoint exposes by the cli (e.g. `http://localhost:8000/callback?code=xyz`) which also closes the browser.

apps/beeai-server/src/beeai_server/auth/utils.py
jezekra1: I think we should bootstrap the JWKS_URL also from the well known files:
for example:

https://your-app.verify.ibm.com/oidc/endpoint/default/.well-known/openid-configuration

the `jwks_uri` is included there

pilartomas: Rather target oAuth well-knowns: https://datatracker.ietf.org/doc/html/rfc8414

apps/beeai-server/src/beeai_server/auth/utils.py
abhi201191: resolved in PR #1059

apps/beeai-server/src/beeai_server/__init__.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059

apps/beeai-server/src/beeai_server/api/routes/acp.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059

apps/beeai-server/src/beeai_server/auth/models.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059

apps/beeai-server/src/beeai_server/__init__.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059

helm/templates/pvc.yaml
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059
Not using PVC anymore.

apps/beeai-cli/src/beeai_cli/token_store.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059

apps/beeai-cli/src/beeai_cli/token_store.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059
moved this to configuration.py

apps/beeai-cli/pyproject.toml
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059
removed as not using anymore.

apps/beeai-server/src/beeai_server/auth/utils.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059
Using "beeai-server" as audience

apps/beeai-cli/src/beeai_cli/api.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059
Using access_token for authentication. For userinfo (email), getting the user info from IDPs /userinfo endpoint

apps/beeai-server/src/beeai_server/auth/utils.py
abhi201191: resolved in PR https://github.com/i-am-bee/beeai-platform/pull/1059
Also checking for issuer in jwt decode.

PR #1018, opened 2025-08-07
feat(ui): add message feedback

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/716

### DIFF
--- a/apps/beeai-ui/src/modules/feedback/hooks/useFeedback.ts
+++ b/apps/beeai-ui/src/modules/feedback/hooks/useFeedback.ts
@@ -1,0 +1,158 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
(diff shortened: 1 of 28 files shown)


apps/beeai-ui/src/modules/feedback/hooks/useFeedback.ts
```diff
@@ -0,0 +1,165 @@
...
+      }
+    },
+    whileElementsMounted: autoUpdate,
+    middleware: [offset(FEEDBACK_DIALOG_OFFSET), flip()],
+  });
+
+  const dismiss = useDismiss(context);
+  const role = useRole(context, { role: 'dialog' });
+
+  const { getReferenceProps, getFloatingProps } = useInteractions([dismiss, role]);
```
kapetr, line 51: Not part of this PR, but the floating-ui usage seems to be adding a bit too much clutter to the component code. We could consider creating an abstraction into a custom hook.

PetrBulanek: Good point! :-)

Issue #1002, opened 2025-08-04
LLM Proxy extension server support


tomkis:
**Is your feature request related to a problem? Please describe.**
Currently, users can globally configure only a single inference provider. When attempting to define additional providers, the new configuration overwrites the existing one. This limitation prevents users from managing multiple inference providers concurrently and selecting among them dynamically during agent runs. The need is to support multiple providers globally and allow selection per A2A task, with agent builders optionally recommending a default provider.

**Describe the solution you'd like**
Create a service that allows generating per-task tokens representing specific inference provider configurations. These tokens would be passed via the LLM Proxy Extension to the agent during execution. The proxy would then resolve the token to the correct provider, enabling dynamic routing. The agent builder may suggest a default provider token, but users can override this per task as needed.

**Additional context**
- We might need to include inference provider setup in the GUI, as it is currently only available through the CLI.

- this ticket does not address how the agent builder should recommend a provider. That will need to be handled separately.

- We need to define a strategy for selecting the most viable model for a given run, potentially considering task type, performance, cost, and latency.

**Technical Proposal**

<img width="3264" height="2448" alt="Image" src="https://github.com/user-attachments/assets/98d8c542-7c48-40ec-9c59-35b85e83069a" />

jezekra1:
The proposed solution will add two endpoints to the platfrom:

`POST /context` - create new context - return context token (= session = thread = conversation, naming adapted from a2a) 
`POST /context/{context_id}/token` - issue a new token (with different permission requests)

**Requiring new permissions**
The agent can at any point request more permissions using the `auth_required` message from a2a. This should trigger a UI popup or redirect where the user will approve the new requirements (mcp tool use, knowledge base search)

**Resource scope**
Resources - (currently files, vector stores) will be now associated with the context - agent can create and read files only associated with the context, not other files from other contexts

**Token vs. context lifetime**
Token expiration:
- Tokens are short-lived, meant for a single task. If the token expires during agent run (very slow agent, new token should be elicited by the agent using `auth_required` message)  

Context expiration: The context has an expiration of X days (configurable, 7 by default), after expiration:
- transient resources associated with the context (files, vector stores) will be removed
- persistent resources (agent messages - future feature) will be kept, unless entire context is manually removed

**Secure agent delegation**
> ℹ️ out of scope of this issue
[...]

PR #1021, opened 2025-08-07
feat(platform): add context scoped resources and tokens

jezekra1:
ref: #1002

### DIFF
--- a/apps/beeai-server/src/beeai_server/api/auth.py
+++ b/apps/beeai-server/src/beeai_server/api/auth.py
@@ -54,11 +54,14 @@ def issue_internal_jwt(
     now = utc_now()
     expires_at = now + timedelta(minutes=20)
     payload = {
-        "user_id": str(user_id),
         "context_id": str(context_id),
+        "sub": str(user_id),
         "exp": expires_at,
         "iat": now,
-        "permissions": {
+        "iss": "beeai-server",
+        "aud": "beeai-server",  # the token is for ourselves, noone else should consume it
+        "resource": [f"context:{context_id}"],
+        "scope": {
             "global": global_permissions.model_dump(mode="json"),
             "context": context_permissions.model_dump(mode="json"),
         },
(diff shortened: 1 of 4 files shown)


apps/beeai-server/src/beeai_server/api/auth.py
```diff
@@ -1,35 +1,80 @@
...
-        return await user_service.get_user(user_id=UUID(payload["user_id"]))
-    except jwt.ExpiredSignatureError as e:
-        raise Exception("Token expired") from e
-    except jwt.InvalidTokenError as e:
-        raise Exception("Invalid token") from e
+    expires_at = now + timedelta(minutes=20)
+    payload = {
+        "user_id": str(user_id),
+        "context_id": str(context_id),
+        "exp": expires_at,
```
pilartomas: Suggestion - common set of fields for JWTs is `sub, exp, iss, aud` and all of them are supposed to be validated (+ signature).

apps/beeai-server/src/beeai_server/api/auth.py
```diff
@@ -1,35 +1,80 @@
...
+        "iat": now,
+        "permissions": {
+            "global": global_permissions.model_dump(mode="json"),
+            "context": context_permissions.model_dump(mode="json"),
+        },
+    }
+    return jwt.encode(payload, secret_key, algorithm="HS256"), expires_at
```
pilartomas, line 69: If you expect the token holder to be able to use the information from the JWT, it is probably better to use a public/private keypair for signing. The holder needs to make sure the token is valid before using that information somehow.

apps/beeai-server/src/beeai_server/api/auth.py
```diff
@@ -1,35 +1,80 @@
...
+    expires_at = now + timedelta(minutes=20)
+    payload = {
+        "user_id": str(user_id),
+        "context_id": str(context_id),
+        "exp": expires_at,
+        "iat": now,
+        "permissions": {
+            "global": global_permissions.model_dump(mode="json"),
+            "context": context_permissions.model_dump(mode="json"),
+        },
```
pilartomas: In oAuth, there are `scopes` and newly `resource indicators`. These two concepts cover this permissions ground. I'd use the same naming conventions from the getgo.

apps/beeai-server/src/beeai_server/api/auth.py
jezekra1: right now we don't need this feature, we can change it in the future I think, this does not need a big migration :)

PR #1024, opened 2025-08-08
feat: MCP demands handling in GUI

tomkis:


apps/beeai-ui/src/api/a2a/extensions/component/citation.ts
```diff
@@ -5,7 +5,7 @@
 
 import { z } from 'zod';
 
-import type { A2AExtension } from './types';
+import type { A2AComponentExtension } from '../types';
```
PetrBulanek: Is there any particular reason for "Component" in the name? I think we could align it with the URI. How about renaming the folder to "ui" and the type to`A2AUiExtension`?

And I'm not quite sure about getMessageSchema. In this context, it's not really a message, is it?

apps/beeai-ui/src/api/a2a/extensions/component/trajectory.ts
```diff
@@ -5,7 +5,7 @@
 
 import { z } from 'zod';
 
-import type { A2AExtension } from './types';
+import type { A2AComponentExtension } from '../types';
```
PetrBulanek: Same as above.

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -55,7 +55,10 @@ function AgentRunProvider({ agent, children }: PropsWithChildren<Props>) {
 
   const errorHandler = useHandleError();
 
-  const a2aAgentClient = useMemo(() => buildA2AClient(agent.provider.id), [agent.provider.id]);
+  const a2aAgentClient = useMemo(
```
PetrBulanek, line 56: We can maybe pass the whole `agent` as param into this function. Or at least, you are missing `agent.capabilities.extensions` in deps array here.

apps/beeai-ui/src/api/a2a/extensions/service/mcp.ts
PetrBulanek, line 1: I would just rename the folder to "services" to match the extension URI.

apps/beeai-ui/src/api/a2a/extensions/utils.ts
```diff
@@ -19,3 +21,26 @@ export function getExtensionData<U extends string, D>(extension: A2AExtension<U,
...
+export function extractServiceExtensionDemands<U extends string, D, F>(extension: A2AServiceExtension<U, D, F>) {
+  return function (agentExtensions: AgentExtension[]) {
+    const foundExtension = agentExtensions.find(({ uri }) => uri === extension.getUri());
+    if (!foundExtension) {
+      return null;
+    }
+
+    const data = extension.getDemandsSchema().parse(foundExtension.params);
+    return data;
+  };
```
PetrBulanek: You can lift the `extension.getUri()` and `extension.getDemandsSchema()` outside of the inner function to make run only once, similarly as in the `extractComponentExtensionData`.

apps/beeai-ui/src/api/a2a/extensions/utils.ts
```diff
@@ -19,3 +21,26 @@ export function getExtensionData<U extends string, D>(extension: A2AExtension<U,
...
+
+export function fullfilServiceExtensionDemand<U extends string, D, F>(extension: A2AServiceExtension<U, D, F>) {
+  return (metadata: Record<string, unknown>, fullfillment: F) => {
+    const fullfilment = extension.getFullfilmentSchema().parse(fullfillment);
+
+    return {
+      ...metadata,
+      [extension.getUri()]: fullfilment,
+    };
+  };
```
PetrBulanek: Same here with `extension.getFullfilmentSchema()` and `extension.getUri()`. Also we usually use classic functions instead of the arrow functions where possible. ;-)

PetrBulanek (changes requested):

apps/beeai-ui/src/api/a2a/extensions/component/citation.ts
tomkis: The naming is there just so that it correleates with two types of extension we implement in the platform|:
- Components extension, that extends the message structure to introdouce componetns
- Service extension that extends the behaviour of agent

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
tomkis: I didn't want to increase the coupling here. I think the whole agent concept should be ideally encapsulated in A2A client and this would only provide ID.

The main reason is that we even duplicate typing now. Basically when you check `types.ts` where is definition of `Agent` it duplicates a lot of stuff and exposes it where it should have been hidden in the A2A client.

The main reason why we can't do it now is that agent discovery is not currently implement in the client but done separately, we need to refactor this.

apps/beeai-server/src/beeai_server/application.py
```diff
@@ -134,7 +134,8 @@ async def lifespan(_app: FastAPI, procrastinate_app: procrastinate.App, mcp_serv
             register_telemetry()
             async with procrastinate_app.open_async(), run_workers(app=procrastinate_app), mcp_service:
                 try:
-                    yield
+                    async with mcp_service:
```
pilartomas: duplication, remove

PetrBulanek (approved):

PR #1025, opened 2025-08-11
refactor(ui): use separate hooks for dialog functionality

PetrBulanek:


apps/beeai-ui/src/modules/runs/constants.ts
```diff
@@ -6,3 +6,12 @@
...
 export const SupportedUis: string[] = [SupportedUIType.Chat, SupportedUIType.HandsOff];
+
+export const PROMPT_EXAMPLES_DIALOG_OFFSET = {
+  mainAxis: 27, // Space between the input and the examples
+};
+
+export const RUN_SETTINGS_DIALOG_OFFSET = {
+  mainAxis: 56,
+  crossAxis: -12,
+};
```
kapetr: My preference would be to keep the constants in the hook file, as they’re only used there.

kapetr (approved):

apps/beeai-ui/src/modules/runs/constants.ts
PetrBulanek: That makes sense. :-)

kapetr (approved):

PR #1026, opened 2025-08-11
docs: first iteration of A2A integration  docs

tomkis:


docs/build-agents/agent-details.mdx
```diff
@@ -0,0 +1,71 @@
...
+from beeai_sdk.a2a.extensions import AgentDetail, AgentDetailContributor, AgentDetailTool
+
+server = Server()
+
+@server.agent(
+    detail=AgentDetail(
+        ui_type="chat",
+        user_greeting="Welcome! I'm here to help you with your tasks.",
+        license="Apache 2.0",
+        programming_language="Python",
```
jezekra1, line 36: not necessary, python by default

docs/build-agents/agent-details.mdx
```diff
@@ -0,0 +1,71 @@
...
+        framework="BeeAI",
+        tools=[
+            AgentDetailTool(
+                name="Weather",
+                description="Get the weather for a given location",
+            ),
+        ],
+        homepage_url="https://github.com/beeai-dev/beeai-agents",
+        source_code_url="https://github.com/beeai-dev/beeai-agents",
+        container_image_url="ghcr.io/beeai-dev/beeai-agents:latest",
```
jezekra1: let's not promote latest here, people should never use latest, I'd use something like

```suggestion
        container_image_url=f"ghcr.io/beeai-dev/beeai-agents:{os.getenv('RELEASE_VERSION', '0.0.0')}",
```

or it might be better to just remove it from the docs, it's not mandatory

docs/build-agents/files.mdx
```diff
@@ -0,0 +1,119 @@
...
+)
+from beeai_sdk.platform import File
+from beeai_sdk.util.file import load_file
+
+
+server = Server()
+
+
+@server.agent(
+    default_input_modes=["text/plain"]
```
jezekra1: there should also be `default_output_modes` when we're creating files right?

docs/build-agents/hello-world.mdx
```diff
@@ -0,0 +1,145 @@
...
+- **First argument:** an [A2A `Message`](https://a2a-protocol.org/latest/specification/#64-message-object).
+- **Second argument:** a `Context` object with run details (e.g., `task_id`, `context_id`).
+</Step>
+
+<Step title="Extract text from Message">
+Use `get_message_text()` to quickly extract the text content from a `Message`.
+</Step>
+
+<Step title="Make it an async generator">
+The agent function should be asynchronous and yield results as they’re ready.
```
jezekra1, line 119: This is not necessary, all modes are supported:
- async generator
- sync genrator
- sync function
- async function
you can interact with the queue by using `context.yield_sync` or `context.yield_async`

see tests for examples:
https://github.com/i-am-bee/beeai/blob/7e1f52f0e063c48afa2c256599a083de7c996e5a/apps/beeai-sdk/tests/e2e/test_yields.py#L79

Wouldn't add all examples to the hello world, but we can make some more "advanced" tutorial. 

> Note: working with the platform api (file/vector-store management) will be a bit cumbersome (we have async only platform api client for now

docs/build-agents/messages.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+#### Plain strings
+Simple strings are automatically converted to textual A2A `Message` objects:
+
+```python
+yield "Hello, world!"
+```
+
+#### Plain dict
+Dictionaries send an empty A2A `Message` and supply the dictionary content as message metadata:
```
jezekra1: This is not true anymore, dictionaries send `DataPart`. You can yield `Metadata` now:

https://github.com/i-am-bee/beeai/blob/7e1f52f0e063c48afa2c256599a083de7c996e5a/apps/beeai-sdk/tests/e2e/test_yields.py#L488

docs/build-agents/messages.mdx
```diff
@@ -0,0 +1,132 @@
...
+```
+
+#### Artifact
+Tangible outputs produced by the agent, such as documents, files, or other generated content.
+
+```python
+import uuid
+from a2a.types import Artifact, FilePart, FileWithUri
+
+yield Artifact(
```
jezekra1, line 81: use `AgentArtifact` convenience wrapper (no need to generate id)

docs/build-agents/messages.mdx
```diff
@@ -0,0 +1,132 @@
...
+from a2a.types import TaskStatus, TextPart, Message, TaskState
+
+yield TaskStatus(
+    message=Message(message_id=str(uuid.uuid4()), role="agent", parts=[TextPart(text="Please provide some input.")]),
+    state=TaskState.input_required,
+)
+```
+
+#### TaskStatusUpdateEvent
+Similar to `TaskStatus` but designed for streaming updates. Use this when you need to provide progressive status updates, such as when streaming content to the client.
```
jezekra1: There is actually no reason to use this (`TaskStatus` is already streaming), only two possible usecases:
- if you want to add `metadata` to TaskStatusUpdateEvent (quite rare)
- if you want to just forward updates from another agent

docs/build-agents/messages.mdx
```diff
@@ -0,0 +1,132 @@
...
+        message=Message(
+            message_id=str(uuid.uuid4()), role="agent", parts=[TextPart(text="Please provide some input.")]
+        ),
+        state=TaskState.input_required,
+    ),
+)
+```
+
+#### TaskArtifactUpdateEvent
+Similar to `Artifact` but designed for artifact generation updates. Use this when progressively generating artifacts and want to provide updates on the generation progress.
```
jezekra1: same as above

docs/build-agents/agent-details.mdx
tomkis: I wanted to mention here complete example.

docs/build-agents/agent-details.mdx
tomkis: I'll use there as example something like `ghcr.io/beeai-dev/beeai-agents:v0.0.1`

docs/build-agents/hello-world.mdx
tomkis: Good point, i guess we want to create "RunContext" section that would cover all above. Let's keep the hello world simple.

jezekra1 (approved):

PR #1031, opened 2025-08-11
feat(beeai-cli): configure Istio and TLS when OIDC is enabled

tedhabeck:
This PR provides  conditional install / configuration of istio only based upon the value of a beeai-cli environment variable during platform startup.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -61,8 +61,7 @@ serviceAccount:
 podAnnotations: { }
 # This is for setting Kubernetes Labels to a Pod.
 # For more information checkout: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
-podLabels:
-  istio.io/dataplane-mode: ambient
+podLabels: { }
 
 podSecurityContext: { }
 # fsGroup: 2000


apps/beeai-cli/src/beeai_cli/configuration.py
```diff
@@ -31,6 +31,7 @@ class Configuration(pydantic_settings.BaseSettings):
         f"https://github.com/i-am-bee/beeai-platform@v{version()}#path=agent-registry.yaml"
     )
     admin_password: SecretStr | None = None
+    oidc_enabled: bool = False
```
jezekra1: we don't use environment variables for this, it's configurable through standardized helm format
```
beeai platform start --set oidc.enabled=true
```

apps/beeai-cli/src/beeai_cli/commands/platform/base_driver.py
```diff
@@ -114,6 +115,11 @@ async def deploy(self, set_values_list: list[str], import_images: list[str] | No
                 f"Pulling image {image}",
             )
 
+        # conditional check to install istio of odic.enabled=
+        if any("oidc.enabled=true" in value.lower() for value in set_values_list):
+            print("oidc.enabled=true")
+            await install_security(self)
```
JanPokorny:
```suggestion
        if any("oidc.enabled=true" in value.lower() for value in set_values_list):
            await install_security(self)
```

apps/beeai-cli/src/beeai_cli/configuration.py
```diff
@@ -20,7 +20,7 @@ def version():
 @functools.cache
 class Configuration(pydantic_settings.BaseSettings):
     model_config = pydantic_settings.SettingsConfigDict(
-        env_file=None, env_prefix="BEEAI__", env_nested_delimiter="__", extra="allow"
+        env_file=".env", env_prefix="BEEAI__", env_nested_delimiter="__", extra="allow"
```
JanPokorny: It's a CLI, you can't have this since it would load random `.env` files from the user's system (we know the hard way)

```suggestion
        env_file=None, env_prefix="BEEAI__", env_nested_delimiter="__", extra="allow"
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import yaml
+
+
+async def install_security(driver):
```
JanPokorny: Add arg type

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+            "istio/ztunnel",
+            "-n",
+            "istio-system",
+            "--set",
+            "profile=ambient",
+            "--set",
+            "values.global.platform=k3s",
+            "--wait",
+        ],
+        "Installing istio-cni",
```
JanPokorny:
```suggestion
        "Installing istio-ztunnel",
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+            "-n",
+            "cert-manager",
+            "wait",
+            "--for=condition=Available",
+            "--timeout=5m",
+            "--all",
+            "deployment",
+        ],
+        "Waiting for certmanager deployments to be available",
+    )
```
JanPokorny: Install cert-manager using Helm instead, consistent with others + shorter (can `--wait`): https://cert-manager.io/docs/installation/helm/#installing-cert-manager-with-helm

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+        ).encode("utf-8"),
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            "curl -L https://istio.io/downloadIstio | sh -",
+        ],
+        "Installing istioctl...",
```
JanPokorny:
```suggestion
        "Installing istioctl",
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+
+    istio_dir = (
+        (
+            await driver.run_in_vm(
+                [
+                    "/bin/sh",
+                    "-c",
+                    "ls -lahF | grep istio | awk '{print $9}'",
+                ],
+                "Checking for istio install folder...",
```
JanPokorny:
```suggestion
                "Checking for istio install folder",
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+        .strip()
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            f"export PATH=/{istio_dir}/bin:$PATH && k3s kubectl apply -f /{istio_dir}samples/addons/prometheus.yaml",
+        ],
+        "Installing prometheus...",
```
JanPokorny:
```suggestion
        "Installing Prometheus",
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            f"export PATH=/{istio_dir}/bin:$PATH && k3s kubectl apply -f /{istio_dir}samples/addons/kiali.yaml",
+        ],
+        "Installing Kiali...",
```
JanPokorny:
```suggestion
        "Installing Kiali",
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            "k3s kubectl -n istio-system expose deployment kiali --protocol=TCP --port=20001 --target-port=20001 --type=NodePort --name=kiali-external",
+        ],
+        "Exposing Kiali service...",
```
JanPokorny:
```suggestion
        "Exposing Kiali service",
```

apps/beeai-cli/.env.template
```diff
@@ -0,0 +1 @@
+BEEAI__OIDC_ENABLED=False
```
JanPokorny: Remove this file, see above

helm/values.yaml
```diff
@@ -76,6 +77,29 @@ securityContext: { }
...
+      "type": "oidc",
+      "class": "IBM",
+      "client_id": "<oidc_client_id>",
+      "client_secret": "<oidc_client_secret>",
+      "issuer": "<oidc_issuer>",
+      "jwks_url": "<oidc_jwks_url>",
+      "nextauth_url": "http://localhost:8334",
+      "nextauth_redirect_proxy_url": "http://localhost:8334"
+    }
+  ]
```
JanPokorny: Where do you read this configuration?

JanPokorny: Also: it's not formatted properly, Helm uses `camelCase`, YAML syntax, with explanatory comments above the string, default value should be empty, example values should be commented out

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+        "Waiting for certmanager deployments to be available",
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            "k3s kubectl label namespace default istio.io/dataplane-mode=ambient",
+        ],
+        "Labeling the default namespace...",
```
JanPokorny:
```suggestion
        "Labeling the default namespace",
```

TLS_AND_OIDC_GUIDE.md
```diff
@@ -0,0 +1,38 @@
...
+
+The default namespace is labeled istio.io/dataplane-mode=ambient so all intra pod trafic is via ztunnel with the exception of the beeai-platform pod due to it's use of the hostNetwork (istio can not bring a hostNetwork enabled pod into the mesh).
+
+The Kiali console is installed when BEEAI_OIDC_ENABLED is set to True.  The port can be found by shelling into the VM and running the following kuberneties command:
+```bash
+limactl shell --workdir / beeai-platform
+habeck@lima-beeai-platform:/$ kubectl -n istio-system get svc | grep "kiali-external" | awk '{print $5}' | cut -d ':' -f2 | cut -d '/' -f1 
+30431
+```
+Using the output from the above command navigate to `http://localhost:30431/kiali/console`
```
JanPokorny: Inline TLS_AND_OIDC_GUIDE.md in CONTRIBUTIN.md, no need for extra file

CONTRIBUTING.md
```diff
@@ -74,6 +74,33 @@ eval "$(mise run beeai-platform:shell)"
...
+  client_secret: <your_oidc_client_secret>
+  issuer: <your_issuer>
+  jwks_url: <your_jwks_endpoint_here>
+  nextauth_trust_host: true
+  nextauth_url: "https://beeai-platform.api.testing:8336"
+  nextauth_redirect_proxy_url: "https://beeai-platform.api.testing:8336"
+  nextauth_secret: "<To generate a random string, you can use the Auth.js CLI: npx auth secret>
+```
+
+
```
JanPokorny: Update docs after fixing issues

apps/beeai-cli/src/beeai_cli/configuration.py
JanPokorny: Tip: if you don't want to specify env value every time, you can put it in `mise.local.toml`. But I see that you don't actually read the env in this PR anyway so you should update the docs

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            "k3s kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null",
+        ],
+        "installing gateway CRDS",
+    )
```
JanPokorny: Technically this should be bash since `sh` is not guaranteed to be bash (and `&>` is bash only)

```suggestion
    await driver.run_in_vm(
        ["/bin/bash", "-c", "k3s kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null"],
        "Downloading gateway CRDS",
    )

    await driver.run_in_vm(
        [
            "/bin/bash",
            "-c",
            "k3s kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null",
        ],
        "installing gateway CRDS",
    )
```

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+        "Downloading gateway CRDS",
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            "k3s kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null",
+        ],
+        "installing gateway CRDS",
```
JanPokorny: These two run the same command, mistake?

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
```diff
@@ -0,0 +1,440 @@
...
+        .stdout.decode()
+        .strip()
+    )
+
+    await driver.run_in_vm(
+        [
+            "/bin/sh",
+            "-c",
+            f"export PATH=/{istio_dir}/bin:$PATH && k3s kubectl apply -f /{istio_dir}samples/addons/prometheus.yaml",
+        ],
```
JanPokorny: Why do you need `istioctl` in PATH for this? It seems that you only apply the YAML. You could skip downloading the full binary and just apply the YAMLs from GitHub.

CONTRIBUTING.md
```diff
@@ -74,6 +74,33 @@ eval "$(mise run beeai-platform:shell)"
...
 ```
 
+### Enabling or disabling auth flows for beeai_cli
+
+- Copy the apps/beeai-cli/.env.template as .env
+- To turn on oauth on platform startup set `BEEAI__OIDC_ENABLED=True` in the .env file.
+- Add an entry to /etc/hosts on your local system:
+```
+# Added by BEEAI-PLATFORM
+127.0.0.1        beeai-platform.api.testing
```
JanPokorny: The domain name is inconsistent.

- The TLD should be `.localhost` since everything `*.localhost` goes to localhost (at least on macOS and Ubuntu), so you could skip editing hosts file entirely
- Internally we use `beeai-server` for the API server, not `api`... but better yet it's the best candidate to serve the 2nd level domain anyway, so skip this segment
- Our domain is `beeai.dev` so let's be consistent and use `beeai.localhost`

So my suggestion for the domain is:
- change `beeai-platform.api.testing` -> `beeai.localhost` everywhere
- remove the instructions to edit `/etc/hosts` since that's not needed anymore

TLS_AND_OIDC_GUIDE.md
JanPokorny: (Also update instructions, I see that you reference the env var)

helm/values.yaml
tedhabeck: For this PR it's not required. I'll remove everything but oidc.enabled
FYI: All the above is used in this PR: (The UI next-auth update) https://github.com/i-am-bee/beeai-platform/pull/1041

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
tedhabeck: `BaseDriver` imports install_security,  if `install_security` tries to import `BaseDriver` to use in typing it creates a circular dependency and won't build.

tedhabeck: Resolved via DM  @JanPokorny . Needed to conditionally import the type and quote the type.

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
tedhabeck: Good catch! yep, cut/paste error.   Fixed now.

apps/beeai-cli/src/beeai_cli/commands/platform/configure_securityl.py
tedhabeck: These are customized to work with istio, and not the full community charts.

I removed the unnecessary path entry.

JanPokorny: You can install from here, no need to install Istio for a few YAMLs https://raw.githubusercontent.com/istio/istio/refs/heads/master/samples/addons/prometheus.yaml

helm/values.yaml
```diff
@@ -116,14 +120,14 @@ livenessProbe:
     port: http
   initialDelaySeconds: 5
   failureThreshold: 5
-  periodSeconds: 5
+  periodSeconds: 30
```
jezekra1: please, let's not touch the healthchecks if it's not necessary (i.e if it's not crashing)

PR #1032, opened 2025-08-12
chore(ui): replace ui_type with interaction_mode

### LINKED ISSUE #999: Remove "ui types" (hands-off/chat) and replace with single/multi-turn specification
## Description

Currently, the BeeAI platform requires developers to specify agent UI types as either `chat` or `hands-off` via the `PlatformUIType` enum. This creates unnecessary complexity and couples the agent definition to specific UI implementations. Instead, we should simplify this to allow developers to specify whether their agent supports single-turn or multi-turn interactions, letting the platform handle the appropriate UI rendering.

## Current Implementation

The current system uses:
- `UiType` enum with `Chat = 'chat'` and `HandsOff = 'hands-off'` values
- `PlatformUIType` enum with `'chat' | 'hands-off'` options
- Agent metadata annotation: `PlatformUIAnnotation.ui_type`
- UI rendering logic that switches between Chat and HandsOff components based on `ui_type`

## Proposed Changes

Up for discussion - but thinking the dev is able to define:
InteractionMode='singleTurn' OR 'multiTurn'

## Benefits

1. Developers think in terms of interaction patterns (single vs multi-turn) rather than specific UI implementations
2. Agent definitions are decoupled from specific UI component names
3. New UI components can be added without changing agent definitions
4. The interaction mode clearly communicates the agent's intended usage pattern
5. Fewer magic strings and more semantic naming

kapetr:
Closes #999 

The only remaining references to the old naming are in the docs files (`gui-components.mdx` and `build-agents.mdx`), I didn't edit those, since they reference `acp_sdk` and need to be updated.

### DIFF
--- a/docs/how-to/build-agents.mdx
+++ b/docs/how-to/build-agents.mdx
@@ -6,10 +6,7 @@ description: "Bring agents from any framework into BeeAI"
 This guide shows you how to wrap existing agents from any framework to work with the BeeAI Platform. You'll learn to use the ACP SDK to create a standardized interface for your agent.
 
 <Tip>
-  If you're building new agents from scratch, check out the [BeeAI
-  Framework](https://github.com/i-am-bee/beeai-framework) with its [native
-  platform
-  integration](https://github.com/i-am-bee/beeai-framework/blob/main/python/docs/integrations.md#beeai-platform-integration).
+If you're building new agents from scratch, check out the [BeeAI Framework](https://github.com/i-am-bee/beeai-framework) with its [native platform integration](https://github.com/i-am-bee/beeai-framework/blob/main/python/docs/integrations.md#beeai-platform-integration).
 </Tip>
 
 ## Prerequisites
@@ -74,7 +71,6 @@ my-agent/
 ```
 
 **Important files:**
-
 - `agent.py` - Where you implement your agent's behavior
 - `pyproject.toml` - Add new Python dependencies here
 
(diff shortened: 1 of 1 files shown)


agents/ui/hello/src/client.py
PetrBulanek, line 1: Seems like there are only formatting changes (some are also in other files). Do we want to commit this? Do the guys in Python have something like we have Prettier in JS, or why is this happening? @jezekra1 @JanPokorny

docs/how-to/build-agents.mdx
```diff
@@ -6,7 +6,10 @@ description: "Bring agents from any framework into BeeAI"
 This guide shows you how to wrap existing agents from any framework to work with the BeeAI Platform. You'll learn to use the ACP SDK to create a standardized interface for your agent.
 
 <Tip>
-If you're building new agents from scratch, check out the [BeeAI Framework](https://github.com/i-am-bee/beeai-framework) with its [native platform integration](https://github.com/i-am-bee/beeai-framework/blob/main/python/docs/integrations.md#beeai-platform-integration).
+  If you're building new agents from scratch, check out the [BeeAI
+  Framework](https://github.com/i-am-bee/beeai-framework) with its [native
+  platform
+  integration](https://github.com/i-am-bee/beeai-framework/blob/main/python/docs/integrations.md#beeai-platform-integration).
```
PetrBulanek: A bit weird new lines formatting.

agents/ui/hello/src/client.py
kapetr: Yes - `ruff`. But I accidentally ran it on the whole project. Later on I realized there are separate `format` tasks for mise - I reverted the most of the unnecessary updates, but some of them slipped away I see.

docs/how-to/build-agents.mdx
kapetr: Prettier did that, I will revert this, since I didn't change anything in the end.

PR #1033, opened 2025-08-12
feat(beeai-cli): refactor and improve `beeai platform ...`

JanPokorny:
- split VM backends into classes, factor out common logic
- unify the installation of k3s
- use helm binary to deploy chart, now we can use `--wait` for deploying and `template` for pre-pulling
- pre-pull used images separately (without progress bars for now, it's easy to pass through the `ctr` output but then it's super noisy)
- update `limactl` to `1.2.1`, no longer needs the SSH workaround
- other minor improvements


helm/values.yaml
```diff
@@ -418,11 +418,11 @@ forge:
     DATABASE_URL: "sqlite:////data/forge.db"
   storage:
     storage: 256Mi
-    storageClass: null
+    storageClass: local-path
```
JanPokorny, line 421: On `k3s` it automatically receives `local-path` and then fails Helm upgrade since the field changed independently... not sure if there's a more correct fix

PR #1041, opened 2025-08-13
enh: issue-767 UI component update

tedhabeck:
Enable w3id authentication in the UI.


apps/beeai-ui/src/app/page.tsx
```diff
@@ -16,7 +18,15 @@ export const dynamic = 'force-dynamic';
 
 export default async function LandingPage() {
   let firstAgentName;
-
+  const session = await auth();
+  const isOidcDisabled = process.env.NEXT__DISABLE_OIDC === 'true';
+  if (!session?.user) {
+    await connection();
```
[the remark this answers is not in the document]
tedhabeck: I'm guessing you did not consider the version of next-auth in use.  
ref:
(alias) connection(): Promise<void>
import connection
This function allows you to indicate that you require an actual user Request before continuing.

During prerendering it will never resolve and during rendering it resolves immediately.

apps/beeai-ui/src/auth.ts
```diff
@@ -0,0 +1,155 @@
...
+let provider_list: {
+  id: string;
+  name: string;
+  type: 'oidc';
+  issuer: string;
+  client_id: string;
+  client_secret: string;
+  nextauth_redirect_proxy_url: string;
+}[] = [];
+if (!process.env.NEXT__DISABLE_OIDC) {
```
kapetr: It should be:
```
if (process.env.NEXT__DISABLE_OIDC !== 'true')
````

But I'd rather use positive naming and create a constant, where the env will be evaluated to boolean (ideally to `utils/constants.ts`), such as:
```
export const OIDC_ENABLED = process.env.OIDC_ENABLED === 'true';
```

File: apps/beeai-ui/src/auth.ts [hunk 3]
```diff
@@ -0,0 +1,155 @@
...
+  const rootPath = process.env.OIDC__PROVIDERS_PATH || './providers';
+  provider_list = JSON.parse(await readFile(`${rootPath}/providers.json`, 'utf8'));
+}
```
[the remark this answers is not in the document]
tedhabeck: Done

apps/beeai-ui/src/app/auth/providers/providerlist.ts
PetrBulanek, line 1: Do we want to support all these providers? In the previous bee-ui application, we actually only had IBM w3id.

apps/beeai-ui/src/app/auth/providers/providerlist.ts
```diff
@@ -0,0 +1,160 @@
...
+    'spotify',
+    'twich',
+    'twitter',
+    'vipps',
+    'w3id',
+    'workos',
+    'zoom',
+  ];
+  constructor() {
+    function getPrvoiderFromName(name: string) {
```
PetrBulanek: Typo `getPrvoiderFromName` => `getProviderFromName`

apps/beeai-ui/src/app/signin/page.tsx
```diff
@@ -0,0 +1,73 @@
...
+                  // Otherwise if a redirects happens Next.js can handle it
+                  // so you can just re-thrown the error and let Next.js handle it.
+                  // Docs:
+                  // https://nextjs.org/docs/app/api-reference/functions/redirect#server-component
+                  throw error;
+                }
+              }}
+            >
+              <div className={classes.beeaiLogin}>
+                <button type="submit" className="cds--btn cds--btn--md cds--layout--size-md cds--btn--primary">
```
PetrBulanek: We should rather use the Button component from the @carbon/react package.

apps/beeai-ui/src/app/signin/signin.module.scss
PetrBulanek, line 1: For most CSS properties, we use rem units (padding, margin, width, height, etc.). You can use the helper function `rem(16px)` or use variables from Carbon (e.g., `$spacing-05)`.

I'm also getting a few Stylelint errors here. Prettier should be able to fix all of that automatically.

image

apps/beeai-ui/src/index.ts
```diff
@@ -3,6 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+export * from './app/signin/page';
+export * from './auth';
```
PetrBulanek: This should not be necessary, if I am not mistaken. We only use this index file for the option of importing shared UI components into the beeai-web app.

PetrBulanek (dismissed):
When I start the build using `mise beeai-platform:start` and then go to the UI, which runs by default at `http://localhost:8334/`, I am redirected to the sign-in page with a login button, but when I open the agent menu from the header, I can easily navigate to another route. Is this the correct behavior? I would expect the sign-in page to have a different layout (the list of agents would not be visible until I log in) and any navigation elsewhere would be blocked with a redirect back to the sign-in page.

CONTRIBUTING.md
```diff
@@ -74,6 +74,95 @@ eval "$(mise run beeai-platform:shell)"
 deactivate
 ```
 
+### Enabling or disabling security
+
+## Disabling security
```
kapetr: The heading structure is incorrect - you’re nesting an h2 inside an h3. Please adjust it to follow the proper document hierarchy.

CONTRIBUTING.md
```diff
@@ -74,6 +74,95 @@ eval "$(mise run beeai-platform:shell)"
...
+
+Authentication, and authorization are disabled by default.
```
kapetr: This says auth should be disabled by default, but when I start the UI, it is enabled and the Login page pops up.

apps/beeai-ui/src/app/auth/providers/providerlist.ts
tedhabeck: fixed

apps/beeai-ui/src/app/auth/providers/providerlist.ts
tedhabeck: Per our discussion in today's call, I've reduced it to just w3id and IBMiD

apps/beeai-ui/src/app/signin/page.tsx
tedhabeck: I tried that but carbon/react Button does not compile in a server side component.  Hence the cheat to use it's style markers as a work-around.

e.g.:
Import trace for requested module:
./src/auth.ts
 ⨯ TypeError: createContext only works in Client Components. Add the "use client" directive at the top of the file to use it. Read more: https://nextjs.org/docs/messages/context-in-server-component
    at __webpack_require__ (.next/server/webpack-runtime.js:33:43)
    at __webpack_require__ (.next/server/webpack-runtime.js:33:43)
    at __webpack_require__ (.next/server/webpack-runtime.js:33:43)
    at eval (webpack-internal:///(rsc)/./src/app/signin/page.tsx:12:71)
    at __webpack_require__.a (.next/server/webpack-runtime.js:105:13)
    at eval (webpack-internal:///(rsc)/./src/app/signin/page.tsx:1:21)
    at <unknown> (rsc)/./src/app/signin/page.tsx (/Users/habeck/beeai-istio-helm/beeai-platform/apps/beeai-ui/.next/server/app/signin/page.js:139:1)
    at Function.__webpack_require__ (.next/server/webpack-runtime.js:33:43) {
  page: '/signin'
}
 GET /signin?callbackUrl=http%3A%2F%2Flocalhost%3A3000%2F 500 in 1514ms

CONTRIBUTING.md
tedhabeck: The odic section in the values.yaml should have had enabled: false.  Fixed.

apps/beeai-ui/src/app/signin/signin.module.scss
tedhabeck: I'm using visual studio code (latest) and have the stylelint plugin installed.  I do not see those CSS warnings.  Is there something I need to configure in settings?

apps/beeai-ui/src/auth.ts
tedhabeck: Done

apps/beeai-ui/src/app/signin/page.tsx
kapetr: The solution is to create a client component (e.g. `SignInForm`) with `'use client'` directive so you can safely use Carbon components like `Button`.

apps/beeai-ui/src/app/signin/signin.module.scss
kapetr: You are right. The stylelint rules were not working properly, the #1068 is fixing that.

apps/beeai-ui/src/app/signin/page.tsx
tedhabeck: The signin form has a "use server" directive in it already...

kapetr: You need to create the new component in a new file, look at the other routes - `RunView` used in `app/run/page.tsx `might be a good example.

tedhabeck: Cool, done.

CONTRIBUTING.md
```diff
@@ -108,6 +104,146 @@ The default namespace is labeled `istio.io/dataplane-mode=ambient`. This ensures
...
+   ▲ Next.js 15.3.4
+   - Local:        https://localhost:3000
+   - Network:      https://192.168.1.92:3000
+   - Environments: .env
+   - Experiments (use with caution):
+     ⨯ cssChunking
+
+ ✓ Starting...
+ ✓ Ready in 1705ms
+ ⚠ ./src/auth.ts
```
kapetr, line 189: Does it make sense to keep this console output here?

apps/beeai-ui/src/app/(signin)/layout.tsx
kapetr, line 1: There's quite a bit of duplicated code. We could move the common parts to the parent layout and only keep the differences between (main) and (signin) layouts here. We can refactor this later.

kapetr (approved):

PR #1054, opened 2025-08-18
feat: LLM fulfilment in GUI

tomkis:


apps/beeai-ui/src/modules/platform-context/contexts/PlatformContextProvider.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+      setContextId(context.id);
+    },
+    [setContextId],
+  );
+
+  const resetContext = useCallback(() => {
+    setContextId(null);
+
+    createContext().then(setContext);
```
PetrBulanek, line 32: Just a minor point – why not use async/await?

PetrBulanek (approved):

PR #1059, opened 2025-08-18
feat(oauth): Platform API authorization check + beeai-cli authentication

abhi201191:


apps/beeai-server/src/beeai_server/__init__.py
```diff
@@ -20,6 +20,20 @@
...
+
+if not configuration.auth.disable_auth:
+    JWKS_URL = configuration.auth.jwks_url
+
+configuration: Configuration = get_configuration()
+
+JWKS_URL = None
+
+if not configuration.auth.disable_auth:
+    JWKS_URL = configuration.auth.jwks_url
```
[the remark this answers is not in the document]
abhi201191: fixed

apps/beeai-server/src/beeai_server/auth/utils.py
```diff
@@ -0,0 +1,91 @@
...
+
+    try:
+        response = requests.get(JWKS_URL)
+        response.raise_for_status()
+        with open(JWKS_FILE, "w") as f:
+            json.dump(response.json(), f, indent=2)
+        print(f"JWKS downloaded and saved to: {JWKS_FILE}")
+    except Exception as e:
+        logger.error("Failed to fetch JWKS: %s", e)
+        raise
```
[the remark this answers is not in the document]
abhi201191: fixed

apps/beeai-cli/src/beeai_cli/commands/auth.py
```diff
@@ -0,0 +1,61 @@
...
+from beeai_cli.token_store import clear_token, save_token
+
+app = AsyncTyper()
+
+config = Configuration()
+BASE_URL = str(config.host).rstrip("/")
+API_BASE_URL = f"{BASE_URL}/api/v1/"
+LOGIN_URL = f"{API_BASE_URL}login"
+CALLBACK_URL = f"{API_BASE_URL}display-passcode"
+TOKEN_PATH = os.path.expanduser("~/.beeai_token")
```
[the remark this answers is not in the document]
abhi201191: fixed

apps/beeai-server/src/beeai_server/__init__.py
```diff
@@ -29,6 +43,26 @@
...
+            "/usr/bin/openssl",
+            "/usr/bin/openssl",
+            "rsa",
+            "--in",
+            "/etc/config/ingestion.key",
+            "--pubout",
+            "-out",
+            "/jwks/ingestion.pem",
+        )
+        logger.info("openssl pubout rc: %s", str(rc))
```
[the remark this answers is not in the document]
abhi201191: removed code block

apps/beeai-server/src/beeai_server/api/routes/auth.py
```diff
@@ -0,0 +1,43 @@
...
+
+from beeai_server.api.dependencies import AuthServiceDependency, ConfigurationDependency
+
+logger = logging.getLogger(__name__)
+
+SLASH_LOGIN = "/login"
+
+router = APIRouter()
+
+pending_tokens: dict[str, dict] = {}
```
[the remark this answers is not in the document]
abhi201191: removed pending_tokens. Not using anymore.

apps/beeai-cli/src/beeai_cli/configuration.py
```diff
@@ -34,13 +34,45 @@
...
+                self.auth_token = SecretStr(token)
+                return self.auth_token
+        return None
+
+    def set_auth_token(self, token: str):
+        """Persist and cache auth token (after login)."""
+        self.home.mkdir(parents=True, exist_ok=True)
+        self.token_file.write_text(token)
+        self.auth_token = SecretStr(token)
+        print(f"here----------------{self.auth_token}")
```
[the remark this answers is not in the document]
abhi201191: resolved

apps/beeai-cli/src/beeai_cli/api.py
```diff
@@ -105,6 +134,9 @@ async def api_stream(
 
 
 @asynccontextmanager
-async def a2a_client(agent_card: AgentCard) -> AsyncIterator[Client]:
+async def a2a_client(agent_card: AgentCard, use_auth: bool = True) -> AsyncIterator[Client]:
+    headers = {}
+    if config.oidc_enabled and use_auth:
+        headers["Authorization"] = await set_auth_header()
     async with httpx.AsyncClient() as httpx_client:
         yield ClientFactory(ClientConfig(httpx_client=httpx_client)).create(card=agent_card)
```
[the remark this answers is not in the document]
abhi201191: fixed

apps/beeai-server/src/beeai_server/domain/repositories/token.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
```
jezekra1: forgotten file

helm/values.yaml
```diff
@@ -78,13 +79,12 @@ initContainers:
 
 oidc:
   enabled: false
-  discovery_url: "https://preprod.login.w3.ibm.com/oidc/endpoint/default/.well-known/openid-configuration"
-  admin_emails: "habeck@us.ibm.com"
+  admin_emails: "abhishek@in.ibm.com"
```
jezekra1: This should be an array:
```
admin_emails:
    - example.admin@ibm.com
```

also let's put a placeholder there:
`example.admin@ibm.com`

apps/beeai-server/pyproject.toml
```diff
@@ -41,6 +41,8 @@ dependencies = [
     "ibm-watsonx-ai>=1.3.28",
     "psycopg[binary]>=3.2.9",
     "openai>=1.97.0",
+    "authlib==1.5.2",
```
jezekra1: unused dependencies

apps/beeai-cli/src/beeai_cli/commands/auth.py
```diff
@@ -0,0 +1,199 @@
...
+        resp = await client.get(url)
+        resp.raise_for_status()
+        metadata = resp.json()
+
+    payload = {"metadata": metadata, "expiry": time.time() + config.resource_metadata_ttl}
+    metadata_file.write_text(json.dumps(payload, indent=2))
+    return metadata
+
+
+def generate_pkce_pair():
```
jezekra1, line 46: Doing your own crypto is usually a bad practice, please use a stadardized way:
```
from authlib.oauth2.rfc7636 import create_s256_code_challenge
```

apps/beeai-server/src/beeai_server/application.py
```diff
@@ -36,6 +40,8 @@
 from beeai_server.telemetry import INSTRUMENTATION_NAME, shutdown_telemetry
 
 logger = logging.getLogger(__name__)
+SESSION_KEY = secrets.token_hex(16)
+middleware = [Middleware(SessionMiddleware, secret_key=SESSION_KEY, https_only=True, session_cookie="session")]
```
jezekra1: Why do we need sessions in addition to access token?
This is not a distributed implementation (won't work with multiple replicas)

jezekra1 (changes requested):
partial review

apps/beeai-server/src/beeai_server/application.py
abhi201191: You are right, forgot to remove the session from server. We don't need it anymore.

apps/beeai-cli/src/beeai_cli/commands/auth.py
abhi201191: Makes sense.

apps/beeai-server/src/beeai_server/domain/repositories/token.py
abhi201191: Removed

apps/beeai-server/pyproject.toml
abhi201191: Removed

apps/beeai-cli/src/beeai_cli/commands/auth.py
abhi201191: Fixed

apps/beeai-server/src/beeai_server/application.py
abhi201191: Removed

helm/values.yaml
abhi201191: Done.

apps/beeai-cli/src/beeai_cli/commands/auth.py
```diff
@@ -0,0 +1,197 @@
...
+    code_verifier = secrets.token_urlsafe(64)
```
jezekra1: let's use `from authlib.common.security import generate_token`

abhi201191: done

abhi201191:
The auth section in values.yaml has been updated. It now supports both basic and oidc configurations.

```
auth:
  enabled: false # Warning, disable only for local deployments
  jwtSecretKey: ""
  basic: 
    enabled: false
  oidc:
    enabled: false
```
If authentication is disabled globally, the basic and oidc configurations are ignored. When authentication is enabled, at least one of basic or oidc must be enabled.

apps/beeai-cli/src/beeai_cli/configuration.py
```diff
@@ -34,13 +36,35 @@ class Configuration(pydantic_settings.BaseSettings):
         f"https://github.com/i-am-bee/beeai-platform@v{version()}#path=agent-registry.yaml"
     )
     admin_password: SecretStr | None = None
+    oidc_enabled: bool = False
+    resource_metadata_ttl: int = 86400
+    client_id: str = "0004ec72-bb41-49d0-804d-430167e5a148"  # pre-registered with AS
```
jezekra1: can we make this something nicer like "beeai-public-client"? This is configurable in ISV

apps/beeai-server/src/beeai_server/bootstrap.py
```diff
@@ -68,5 +69,8 @@ def _set_di[T](service: type[T], instance: T):
 
     _set_di(ITextExtractionBackend, DoclingTextExtractionBackend(di[Configuration].text_extraction))
 
+    # Download JWKS
+    _set_di(JWKS_DICT, setup_jwks(di[Configuration]))
```
jezekra1: This will set a dependency injection for `dict`, please create a custom class: 

Either by inheriting dict `class JwksDict(dict)` or just use a `data: dict | None` field in a normal class

apps/beeai-server/src/beeai_server/api/auth.py
```diff
@@ -81,3 +90,111 @@ def verify_internal_jwt(token: str, configuration: Configuration) -> ParsedToken
...
+            logger.warning(f"Skipping provider with missing issuer: {provider.name}")
+            continue
+        try:
+            response = requests.get(f"{issuer}/jwks")
+            response.raise_for_status()
+            jwks_dict = response.json()
+            jwks_dict_by_issuer[issuer] = jwks_dict
+            logger.debug(f"Fetched JWKS for issuer {issuer} from {issuer}/jwks")
+        except Exception as e:
+            logger.error(f"Failed to fetch JWKS from {issuer}/jwks : {e}")
```
jezekra1, line 113: The app is useless if auth is enabled and the issuer is not configured - no user can log in.
I'd just throw error and fail to boot. Also will help us recover from any transient networking issues (the server will restart automatically instead of failing to fetch JWKS and staying in a broken state forever)

apps/beeai-server/src/beeai_server/api/dependencies.py
```diff
@@ -35,19 +44,69 @@
...
+    """
+    try:
+        token = extract_oauth_token(bearer_auth, cookie_auth)
+    except Exception as e:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail=f"Invalid Authorization header: {e}",
+        ) from e
+
+    claims, issuer = await decode_oauth_jwt_or_introspect(
```
jezekra1, line 70: `decode_oauth_jwt_or_introspect` can return `None` from `introspect_token` -> this will fail to unpack

apps/beeai-server/src/beeai_server/api/dependencies.py
```diff
@@ -35,19 +44,69 @@
...
+        context_permissions=ROLE_PERMISSIONS[user.role],
+    )
 
 
 async def authorized_user(
     user_service: UserServiceDependency,
     configuration: ConfigurationDependency,
     basic_auth: Annotated[HTTPBasicCredentials | None, Depends(HTTPBasic(auto_error=False))],
     bearer_auth: Annotated[HTTPAuthorizationCredentials | None, Depends(HTTPBearer(auto_error=False))],
+    cookie_auth: Annotated[str | None, Security(api_key_cookie)],
```
jezekra1, line 112: what is the `cookie_auth` is it still a leftover from a session middleware or do we need this somewhere?

apps/beeai-cli/src/beeai_cli/configuration.py
abhi201191: Yes, we can. Do we want to have this application created in our ISV tenant or where? Accordingly will create this.

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +68,106 @@ class AgentRegistryConfiguration(BaseModel):
...
+class OidcConfiguration(BaseModel):
+    disable_oidc: bool = False
+    admin_emails: list[str] = Field(default_factory=list)
+    providers: list[OidcProvider] = Field(default_factory=list)
+    scope: list[str] = ["openid", "email", "profile"]
+
+    @model_validator(mode="before")
+    @classmethod
+    def parse_providers(cls, values: dict) -> dict:
+        providers_string = os.environ.get("OIDC__PROVIDERS")
```
jezekra1: pydantic will load a json array automatically (should be in `values["providers"]`) no need to parse the string again

also don't use `environ` here, this config supports `.env` which will not be propagated to `os.environ` afaik

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +68,106 @@ class AgentRegistryConfiguration(BaseModel):
...
+        if providers_string:
+            try:
+                providers_data = json.loads(providers_string)
+                for p in providers_data:
+                    p["client_secret"] = Secret(p["client_secret"])
```
jezekra1: no need to wrap in secret - pydantic will do this automatically

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +68,106 @@ class AgentRegistryConfiguration(BaseModel):
...
+    disable_oidc: bool = False
+    admin_emails: list[str] = Field(default_factory=list)
+    providers: list[OidcProvider] = Field(default_factory=list)
+    scope: list[str] = ["openid", "email", "profile"]
+
+    @model_validator(mode="before")
+    @classmethod
+    def parse_providers(cls, values: dict) -> dict:
+        providers_string = os.environ.get("OIDC__PROVIDERS")
+        if providers_string:
```
jezekra1: this whole block is probably ineffective - this is empty if you are using .env
```
        providers_string = os.environ.get("OIDC__PROVIDERS")
```
so values are returned as usual and parsed using pydantic and everything works correctly, you can remove the validator altogether

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +68,106 @@ class AgentRegistryConfiguration(BaseModel):
...
+    @model_validator(mode="after")
+    def validate_auth(self):
+        if self.disable_oidc:
+            logger.critical("Oauth Authentication is disabled! This is suitable only for local development.")
+            return self
+        if not self.providers:
+            raise ValueError("At least one OIDC provider must be configured if OIDC is enabled")
+        for provider in self.providers:
+            required = ["issuer", "client_id", "client_secret"]
+            for field in required:
```
jezekra1: The fields are already validated by pydantic, they are not missing

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +68,106 @@ class AgentRegistryConfiguration(BaseModel):
...
     disable_auth: bool = False
+    oidc: OidcConfiguration = Field(default_factory=OidcConfiguration)
+    basic: BasicAuthConfiguration = Field(default_factory=BasicAuthConfiguration)
+
+    @model_validator(mode="before")
+    @classmethod
+    def pre_validate_auth(cls, values: dict) -> dict:
+        disable_auth = values.get("disable_auth")
+        if isinstance(disable_auth, str):
+            disable_auth = disable_auth.lower() in ("true", "1", "yes")
```
jezekra1: this is not a reliable way to check for truthfullness, we use pydantic settings and it's methods

please use a validation mode "after" - that gives you access to self after parsing, you can still modify the values

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +68,106 @@ class AgentRegistryConfiguration(BaseModel):
...
+            values["oidc"] = {"disable_oidc": True}
+
+        else:
+            basic = values.get("basic")
+            oidc = values.get("oidc")
+
+            basic_enabled = basic and str(basic.get("disable_basic", "false")).lower() not in ("true", "1", "yes")
+            oidc_enabled = oidc and str(oidc.get("disable_oidc", "false")).lower() not in ("true", "1", "yes")
+
+            # If only one is enabled, disable the other
```
jezekra1: 1. why cannot be both enabled?
2. it's a bit weird to modify users configuration this way - it's better to fail than replace the configuration value, makes it harder to debug

apps/beeai-server/src/beeai_server/api/dependencies.py
abhi201191: done

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: 1. Line #146 and #147 handles the case when both are enabled.
2. I think this is needed in the following scenario:
When both auth and oidc are enabled but no values are provided for basic. In such cases, when AuthConfiguration is initialized, and basic also initialized with their default values (`disable_basic = False`). This leads to a ValueError since `admin_password` is not set. Therefore, it is necessary to explicitly set the appropriate values.

abhi201191: It is mandatory to specify all environment values. For example, if only OIDC is required, then basic_enabled = true must still be explicitly set.

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: Done

apps/beeai-server/src/beeai_server/api/auth.py
abhi201191: Make sense. Done

apps/beeai-cli/src/beeai_cli/configuration.py
abhi201191: In our ISV tenant, I dont have option to set the client id
image

apps/beeai-server/src/beeai_server/bootstrap.py
abhi201191: fixed

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: removed the whole block, as not required.

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: removed the whole block, as not required.

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: removed the whole block, as not required.

apps/beeai-server/src/beeai_server/api/dependencies.py
abhi201191: From UI, it is being used to send token to server APIs.

apps/beeai-cli/src/beeai_cli/configuration.py
jezekra1: I see, the field looked editable but it's not actually. That's a little inconvenient, we'll need to change this with our final ISV application location.

apps/beeai-server/pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "ibm-watsonx-ai>=1.3.28",
     "psycopg[binary]>=3.2.9",
     "openai>=1.97.0",
+    "itsdangerous==2.2.0",
```
jezekra1: this is not imported anywhere, is it leftover?

abhi201191: Yes, fixed

apps/beeai-server/src/beeai_server/api/dependencies.py
```diff
@@ -35,19 +43,69 @@
...
+        userinfo = await fetch_user_info(token, f"{provider.issuer}/userinfo")
+        email = userinfo.get("email")
+
+    if not email:
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="Email not available in token or userinfo")
+
+    is_admin = email in configuration.auth.oidc.admin_emails
+
+    try:
+        user = await user_service.get_user_by_email(email=email)
```
pilartomas, line 95: Establishing the user identity from the email claim is generally not a good idea. Especially if it is unknown what specific issuer will be used.

Generally, it is best to use `sub` and leave the identity linking (usually email-based) on the auth server via federation. Furthermore, the user should be derived from `(issuer, sub)` pair to ensure there will be no impersonation is case the issuer will be changed in the future.

However, in case we really need to rely on the `email` claim, we must at least check `email_verified` claim.

abhi201191: Good point. I was also considering using `sub` as the unique identifier.
One thing I need clarity on is how we should determine whether a user has the `admin` role if we switch to `sub`.

Alternatively, we could stick with using `email` for admin checks, provided we also validate the `email_verified` claim.

@jezekra1 FYI

abhi201191: fixed: also validating `email_verified` claim

apps/beeai-server/src/beeai_server/configuration.py
jezekra1: as discussed on call - let's make this behavior explicit - both can be enabled
if one is enabled, do not turn off the other, that might be unintentional, just fail if there are some variables missing
the users will manage this in values.yaml explicitly

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -66,22 +66,87 @@ class AgentRegistryConfiguration(BaseModel):
...
+            if basic_enabled and not oidc:
+                values["oidc"] = {"disable_oidc": True}
+            elif oidc_enabled and not basic:
+                values["basic"] = {"disable_basic": True}
+            elif not basic and not oidc:
+                # If neither is provided, initialize both with defaults
+                values["basic"] = {}
```
jezekra1: this is invalid configuration - auth is enabled in this part of code, so the users must provide at least one configuration, if both are empty we want to fail rather than using defaults

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: Removed the code block updating the values if not set. Now, it will fail if missing.

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: Yes, this validation is being done in `after`. If both are disabled and auth is enabled, code will throw ValueError
```
if self.basic.disable_basic and self.oidc.disable_oidc:
            raise ValueError("If auth is enabled, either basic or oidc must be enabled")
```

apps/beeai-server/src/beeai_server/configuration.py
abhi201191: Removed the code block updating the values if not set

PR #1068, opened 2025-08-20
chore(ui): fix stylelint checks

kapetr:


apps/beeai-ui/src/components/ExpandButton/ExpandButton.module.scss
```diff
@@ -24,6 +24,8 @@
   }
   &:focus-visible {
     @include focus-outline('outline');
+
+    /* stylelint-disable-next-line no-duplicate-selectors */
```
penge, line 28: Curious why not having something like?

```scss
&:focus-visible {
  @include focus-outline('outline');

  outline-offset: 0;
  border-radius: $spacing-02;
}
```

kapetr: To be compliant with upcoming sass behaviour and eliminate these console warnings:
> Sass's behavior for declarations that appear after nested rules will be changing to match the behavior specified by CSS in an upcoming version. To keep the existing behavior, move the declaration above the nested rule. To opt into the new behavior, wrap the declaration in & {}

I'm also not happy with the solution, but I haven't found a better one.

penge: Thanks for the clarification!

penge (approved):

PR #1089, opened 2025-08-26
chore: next-auth session fix, and token verification fix

tedhabeck:
Correctly verify and decode incoming tokens, enabling correct session values in the ui.


apps/beeai-ui/src/components/Avatar/Avatar.module.scss
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.avatar {
+  position: absolute;
+  right: 80px;
```
kapetr: Screenshot 2025-09-01 at 12 18 05

apps/beeai-ui/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useSession } from 'next-auth/react';
+
+import classes from './Avatar.module.scss';
+import UserAvatar from './UserAvatar';
+
+export default function Avatar() {
+  const { data: session } = useSession();
+  function getUserAvatar() {
+    return <UserAvatar user={session?.user} />;
+  }
```
kapetr: I'd remove this function and load the user from session inside `UserAvatar` component.

apps/beeai-ui/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <div className={classes.avatar}>
+      <OverflowMenu flipped renderIcon={getUserAvatar}>
```
kapetr: Then you could just go with `renderIcon={UserAvatar}`

apps/beeai-ui/src/components/AppHeader/AppHeader.tsx
```diff
@@ -45,6 +46,7 @@ export function AppHeader({ className }: Props) {
               </div>
             </>
           )}
+          <Avatar />
```
kapetr, line 48: I’d suggest hiding the avatar with the menu entirely, when auth is not enabled.

apps/beeai-ui/src/components/Avatar/UserAvatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export default function UserAvatar(props: UserAvatarProps) {
+  let userInitials = '?';
+  if (props.user?.name) {
+    const namesArray = props.user.name.split(' ');
+    const lastIndex = namesArray.length - 1;
+    userInitials = namesArray[0][0] + ' ' + namesArray[lastIndex][0];
+  }
+  return (
+    <div className="user-icon overflow-user-menu">
```
kapetr: There are no styles defined for these css classes, and we are using (s)css modules for styling, please see other components.

apps/beeai-ui/src/components/Avatar/UserAvatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <g id="Group">
+            <circle id="Oval" stroke="#979797" cx="20" cy="20" r="19.5"></circle>
+            <text id="TH" fontFamily="Helvetica" fontSize="12" fontWeight="bold" fill="#FFFFFF">
+              <tspan x="10" y="24">
+                {userInitials}
+              </tspan>
+            </text>
+          </g>
+        </g>
+      </svg>
```
kapetr: Is this ai generated? This can be done with pure html + css, please remove the svg, or let me know, and I'll help with that.

apps/beeai-ui/src/providers.tsx
```diff
@@ -21,18 +22,20 @@ interface Props {
...
-            <RouteTransitionProvider>
-              <ModalProvider>
-                <AppProvider featureFlags={featureFlags}>{children}</AppProvider>
-              </ModalProvider>
-            </RouteTransitionProvider>
-          </ThemeProvider>
-        </ProgressBarProvider>
-      </QueryProvider>
-    </ToastProvider>
+    <SessionProvider basePath={'/auth'}>
```
kapetr: We could define a constant for the path, as it's also used in auth.ts

apps/beeai-ui/src/components/Avatar/UserAvatar.tsx
tedhabeck: I used sketch to to generate an svg and then tweaked it to render the user initials.

apps/beeai-ui/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { OverflowMenu, OverflowMenuItem } from '@carbon/react';
+import { useSession } from 'next-auth/react';
+
+import classes from './Avatar.module.scss';
+import UserAvatar from './UserAvatar';
+
+export default function Avatar() {
+  const { data: session } = useSession();
+  const oidcEnabled = process.env.OIDC_ENABLED === 'true';
```
kapetr: We already have a boolean constant for this.

apps/beeai-ui/src/components/Avatar/UserAvatar.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { useSession } from 'next-auth/react';
+
+export default function UserAvatar() {
+  const { data: session } = useSession();
+  let userInitials = '';
+  if (session?.user?.name) {
+    const namesArray = session.user.name.split(' ');
+    const lastIndex = namesArray.length - 1;
+    userInitials = namesArray[0][0] + ' ' + namesArray[lastIndex][0];
```
kapetr: This is a bit unsafe, let's use this function for extracting initials (copied from another project):
```
const getUserInitials = (name: string) => {
  if (!name) return '';

  // Names can have unicode characters in them, use unicode aware regex
  const matches = [...name.matchAll(/(\p{L}{1})\p{L}+/gu)];
  const initials = (matches.shift()?.[1] ?? '') + (matches.pop()?.[1] ?? '');
  return initials.toUpperCase();
};
```

kapetr (approved):

PR #1101, opened 2025-08-28
feat(sdk): add rest transport to the agent

pilartomas:
Signed-off-by: Tomas Pilar <thomas7pilar@gmail.com>

### DIFF
--- a/apps/beeai-sdk/tests/e2e/test_runs.py
+++ b/apps/beeai-sdk/tests/e2e/test_runs.py
@@ -147,27 +147,6 @@ async def test_run_cancel_awaiter(awaiter: tuple[Server, Client]) -> None:
     assert task_response.status.state == TaskState.canceled
 
 
-async def test_run_cancel_slow_echo(slow_echo: tuple[Server, Client]) -> None:
-    server, client = slow_echo
-    message = create_text_message_object()
-
-    # Start a task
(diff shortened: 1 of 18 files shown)


apps/beeai-sdk/tests/e2e/test_runs.py
```diff
@@ -147,27 +147,6 @@ async def test_run_cancel_awaiter(awaiter: tuple[Server, Client]) -> None:
     assert task_response.status.state == TaskState.canceled
 
 
-async def test_run_cancel_slow_echo(slow_echo: tuple[Server, Client]) -> None:
```
jezekra1, line 150: why is the test invalid?

pilartomas: It waits for task completion and tries to cancel it after.

jezekra1: I see, that was probably my bad refactor with the `get_final_task_from_stream`. We have another test that suffers from the same issue -  `test_run_cancel_awaiter` , can we fix at least one of them so we have cancellation covered?

pilartomas: The `test_run_cancel_awaiter` does not have the same problem, at least in the current implementation. It works.

Issue #1056, opened 2025-08-18
Dynamic model selection


tomkis:
**Is your feature request related to a problem? Please describe.**  
Currently, the LLM proxy accepts any model (`*`) specified during token issuance, which limits fine-grained control over which model should be used in a given conversation. Additionally, there's no mechanism for clients to view or select from a curated list of available models based on their suitability for a task, making model selection opaque and less effective.

**Describe the solution you'd like**  
- Expose an endpoint that returns all available LLM models with their `id` and `name`.
- The client will download an agent card containing a `recommended_model` field.
- Use `recommended_model` to query the new endpoint and retrieve a list of available models, sorted by how well they align with the recommendation.
- Update the LLM proxy to:
  - Read the selected model ID from the token issued.
  - Match and enforce this model selection throughout the conversation context, rather than defaulting to `*`.

PR #1103, opened 2025-08-28
Rework providers and variables

jezekra1:
Ref (model providers): #1056 
Ref (local env variables): #1034
 

**Model providers**
- support for adding multiple llm/embedding providers (API only)
- rename `/api/v1/llm` to `/api/v1/openai`
- add `/api/v1/openai/models` endpoint with list of all available models
- prefix all models with their provider (`openai:gpt-4o`, `other:my-custom-model`)
- add match endpoint `/api/v1/model_providers/match` to aid model selection for LLM extensions
- rename cli command `beeai env setup` -> `beeai model setup`

**(System) configurations**
- add `/api/v1/configurations/system` endpoint for changing system default models (in the future add user overrides)

**Agent providers**
- remove `/api/v1/variables` endpoint -> agent variables are now scoped per providers `/api/v1/providers/{provider_id}/variables`
- change CLI - `beeai env add a=b` -> `beeai [agent] env add <agent-name> a=b`

TBD (in separate PR): rework CLI provider management:
- add `beeai model add/remove/list` commands to add additional models
- add "change default models" CLI option


apps/beeai-cli/src/beeai_cli/__init__.py
```diff
@@ -21,7 +21,7 @@
 logging.basicConfig(level=logging.INFO if Configuration().debug else logging.FATAL)
 
 app = AsyncTyper(no_args_is_help=True)
-app.add_typer(beeai_cli.commands.env.app, name="env", no_args_is_help=True, help="Manage environment variables.")
+app.add_typer(beeai_cli.commands.model.app, name="model", no_args_is_help=True, help="Manage environment variables.")
```
pilartomas: Update description?

apps/beeai-server/tests/conftest.py
```diff
@@ -24,8 +25,10 @@ class TestConfiguration(BaseSettings):
     model_config = SettingsConfigDict(extra="ignore")
     kubeconfig: Path = Path.home() / ".beeai/lima/beeai-local-test/copied-from-guest/kubeconfig.yaml"
     llm_api_base: str = "http://localhost:11434/v1"
-    llm_model: str = "llama3.1"
+    llm_provider_type: ModelProviderType = ModelProviderType.other
+    llm_model: str = "other:llama3.1:8b"
```
pilartomas: Why is the provider type tracked in two variables?

jezekra1: good point, made it a `computed_field`

apps/beeai-cli/src/beeai_cli/commands/agent.py
```diff
@@ -895,3 +908,57 @@ async def agent_detail(
...
+        str, typer.Argument(..., help="Short ID, agent name or part of the provider location")
+    ],
+    env: typing.Annotated[list[str], typer.Argument(help="Environment variables to pass to agent")],
+) -> None:
+    """Store environment variables"""
+    env_vars = dict(parse_env_var(var) for var in env)
+    async with configuration.use_platform_client():
+        provider = select_provider(search_path, await Provider.list())
+        await provider.update_variables(variables=env_vars)
+    # await Variables.save(env_vars)
```
JanPokorny: commented out code

apps/beeai-sdk/src/beeai_sdk/platform/model_provider.py
```diff
@@ -0,0 +1,130 @@
...
+    moonshot = "moonshot"
+    nvidia = "nvidia"
+    ollama = "ollama"
+    openai = "openai"
+    openrouter = "openrouter"
+    perplexity = "perplexity"
+    together = "together"
+    voyage = "voyage"
+    rits = "rits"
+    other = "other"
```
JanPokorny: Enum with lowercase values seems weird, Python convention is UPPERCASE: https://docs.python.org/3/howto/enum.html

apps/beeai-server/src/beeai_server/domain/models/model_provider.py
```diff
@@ -0,0 +1,183 @@
...
+    moonshot = "moonshot"
+    nvidia = "nvidia"
+    ollama = "ollama"
+    openai = "openai"
+    openrouter = "openrouter"
+    perplexity = "perplexity"
+    together = "together"
+    voyage = "voyage"
+    rits = "rits"
+    other = "other"
```
JanPokorny: Ditto with lowercase enum

apps/beeai-server/src/beeai_server/domain/models/model_provider.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+class ModelCapability(StrEnum):
+    llm = "llm"
+    embedding = "embedding"
```
JanPokorny: Ditto with lowercase enum

apps/beeai-server/src/beeai_server/domain/repositories/env.py
```diff
@@ -1,13 +1,38 @@
...
+from enum import StrEnum
 from typing import Protocol, runtime_checkable
+from uuid import UUID
 
 NOT_SET = object()
 
 
+class EnvStoreEntity(StrEnum):
+    provider = "provider"
+    model_provider = "model_provider"
```
JanPokorny: Ditto with lowercase enum

apps/beeai-ui/src/modules/platform-context/api/mutations/useMatchProviders.ts
```diff
@@ -0,0 +1,56 @@
...
+    },
+  });
+
+  const { isSuccess, data } = query;
+
+  useEffect(() => {
+    if (isSuccess && data) {
+      onSuccess(data);
+    }
+  }, [isSuccess, data, onSuccess]);
```
PetrBulanek, line 53: Is this `useEffect` needed? I guess, we could leverage the `onSuccess` callback in `useQuery`, or am I wrong?

I am also not sure if i should be inside api/mutations as it is actually a query. I know that it calls api.POST internally, but I don't really like it that way. Maybe it needs a little refactoring? I would also move queryKey inside separate keys.ts file to align it with other queries. What do you think? @kapetr

PR #1121, opened 2025-09-02
fix lost feedback and variables on provider upgrade

jezekra1:
this fixes the issue that variables and feedback configured for managed providers is lost due to re-create on update

New docker image tag (0.3.1) was treated as an entirely different provider (than 0.3.0) during registry synchronization.
Old provider was removed cascading the deletion to `variables` and `user_feedback` and new provider was created with different ID.

This PR relaxes the id to be calculated only from the image base - without the tag and introduces an "upgrade" procedure.


apps/beeai-server/src/beeai_server/domain/repositories/provider.py
```diff
@@ -14,6 +14,9 @@ async def list(self, *, auto_remove_filter: bool | None = None) -> AsyncIterator
         yield ...  # type: ignore
 
     async def create(self, *, provider: Provider) -> None: ...
+    async def update(self, *, provider: Provider) -> None:
+        pass
```
pilartomas:
```suggestion
    async def update(self, *, provider: Provider) -> None: ...
```

pilartomas (approved):

PR #1128, opened 2025-09-04
fix(ui): align agent run url with server

kapetr:


apps/beeai-ui/src/modules/runs/api/queries/useBuildA2AClient.ts
```diff
@@ -0,0 +1,32 @@
...
+  providerId?: string;
+};
+
+export function useBuildA2AClient<UIGenericPart = never>({
+  providerId,
+  extensions,
+  onStatusUpdate,
+}: Props<UIGenericPart>) {
+  const { data: agentClient } = useQuery({
+    queryKey: ['agent', 'client', providerId],
```
PetrBulanek: I suggest unifying it with other useQuery calls, i.e., moving queryKey to a separate file, such as `apps/beeai-ui/src/modules/providers/api/keys.ts`.

PetrBulanek (approved):

PR #1146, opened 2025-09-08
chore: ACP cleanup

tomkis:


Tomas2D (approved):

apps/beeai-ui/src/api/errors.ts
```diff
@@ -47,11 +47,11 @@ export class HttpError extends CustomError {
   }
 }
 
-export class AcpError extends CustomError {
-  error: AcpErrorResponse;
-  code: AcpErrorResponse['error']['code'];
+export class APIError extends CustomError {
```
PetrBulanek: Now we have APIError and ApiError, which I find a bit confusing. Maybe we could rename it to AgentProtocolError or something similar, at least for now?

PetrBulanek (approved):

apps/beeai-ui/src/api/errors.ts
jezekra1: I think we can remove these classes altogether, ACP error is no longer coming from the server

Issue #633, opened 2025-05-13
User configurable agent settings at runtime


michael-desmond:
**Is your feature request related to a problem? Please describe.**

It is somewhat common for agents to support custom settings beyond the standard LLM configuration (sometimes agents support multiple LLM's but that may be out of scope for this). 

A good example is [gpt-researcher](https://docs.gptr.dev/docs/gpt-researcher/gptr/deep_research) that supports deep research, and tone settings that affect the nature of the final research report. Its also common to see temperature settings, feature flags (web search on/off) etc. 

Ideally the agent developer could define available settings in the agent detail metadata, and these settings would then be made available in the agent interface in a sensible manner. The user could then change settings when interacting with the agent.

**Describe the solution you'd like**

1. Ability to define settings in agent detail metadata i.e.

```
{
  "type": "object",
  "properties": {
    "tone": {
      "title": "Tone",
      "description": "Select the tone of the output report",
      "type": "string",
      "oneOf": [
        { "const": "Objective", "description": "Objective (impartial and unbiased presentation of facts and findings)" },
        { "const": "Formal", "description": "Formal (adheres to academic standards with sophisticated language and structure)" },
        { "const": "Analytical", "description": "Analytical (critical evaluation and detailed examination of data and theories)"}
      ]
    }
  }
}
```

2. Published settings are configurable via agent UI in a sensible manner. 
3. Settings are passed to the running agent. 
 
**Additional context**

[ACP discussion](https://github.com/i-am-bee/acp/discussions/66)

## Designs

[Figma - ready for dev](https://www.figma.com/design/oaJEgk4ZcH3mI1juPGITsx/Agent-Testbed-dev-ready--external-?node-id=4-5007&t=RIEmrTy3ZzdeawWx-1)

michael-desmond:
cc @gabe-l-hart

geneknit:
I would think the developer can decide which are visible to user. Very few would want to expose temperature to an end user (unless their end user also is AI savy), but tone feels like something they would want their user to be able to modify at runtime and therefore expose via a GUI.

tomkis:
Thanks @michael-desmond we are still deciding how to tackle the configuration as part of the ACP. We truly believe there's a great value for agent configuration even outside of the platform.

geneknit:
@tomkis we should look at this use case as feeding requirements through the platform (from user interface through ACP).

[...]

PR #1147, opened 2025-09-08
feat(ui): prototype of agent settings

tomkis:
Refs #633

### DIFF
--- a/mise.lock
+++ b/mise.lock
@@ -1,4 +1,4 @@
-[tools."asdf:CrouchingMuppet/asdf-lima"]
+[[tools."asdf:CrouchingMuppet/asdf-lima"]]
 version = "1.2.1"
 backend = "asdf:CrouchingMuppet/asdf-lima"
 
@@ -34,7 +34,7 @@ backend = "core:node"
 [tools.node.checksums]
 "node-v22.19.0-darwin-arm64.tar.gz" = "sha256:c59006db713c770d6ec63ae16cb3edc11f49ee093b5c415d667bb4f436c6526d"
 
-[tools."pipx:toml-cli"]
+[[tools."pipx:toml-cli"]]
 version = "0.8.1"
 backend = "pipx:toml-cli"
 


apps/beeai-sdk/src/beeai_sdk/a2a/extensions/ui/settings.py
```diff
@@ -0,0 +1,72 @@
...
+from __future__ import annotations
+
+from typing import Literal
+
+from pydantic import BaseModel
+
+from beeai_sdk.a2a.extensions.base import BaseExtensionClient, BaseExtensionServer, BaseExtensionSpec
+
+
+class CheckboxField(BaseModel):
```
PetrBulanek, line 14: I'm missing the "type" here. Is that intentional? CheckboxField cannot stand alone? Always as part of CheckboxGroupField?

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
```diff
@@ -0,0 +1,83 @@
...
+import { z } from 'zod';
+
+import type { A2AServiceExtension } from '../types';
+
+const URI = 'https://a2a-extensions.beeai.dev/ui/settings/v1';
+
+const checkboxField = z.object({
+  id: z.string().nonempty(),
+  label: z.string().nonempty(),
+  default_value: z.boolean(),
```
PetrBulanek, line 15: Shouldn't it be `nullish()`?

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const optionItem = z.object({
+  label: z.string().nonempty(),
+  value: z.string().nonempty(),
+});
+
+const singleSelectField = z.object({
+  type: z.literal('single_select'),
+  id: z.string().nonempty(),
+  options: z.array(optionItem),
```
PetrBulanek: `nonempty()`?

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
```diff
@@ -0,0 +1,83 @@
...
+  default_value: z.string().nonempty(),
```
PetrBulanek, line 33: `nullish()` and if so, ui picks the first value?

apps/beeai-ui/src/modules/runs/settings/RunSettings.tsx
```diff
@@ -0,0 +1,72 @@
...
+  if (!settingsRender?.fields.length) {
+    return null;
+  }
+
+  return (
+    <div className={classes.root}>
+      <IconButton
+        kind="ghost"
+        size="sm"
+        label="Customize Tools"
```
PetrBulanek: We should probably change this label.

apps/beeai-ui/src/modules/runs/settings/SingleSelectSettingsField.module.scss
PetrBulanek, line 1: I would add a 1px gap between items, because now when one is selected and I hover over the one next to it, they visually merge together.

apps/beeai-sdk/src/beeai_sdk/a2a/extensions/ui/settings.py
PetrBulanek, line 1: I would add the option of some kind of heading for individual groups. I think I even saw it in the design.

image

PetrBulanek: That was probably in a previous version, but it's not in Figma now, so you can ignore the previous comment.

kapetr: Yes, it should follow the latest designs. But I agree :)

apps/beeai-sdk/src/beeai_sdk/a2a/extensions/ui/settings.py
kapetr: I guess that's the idea, @tomkis what do you think?

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
kapetr: `default_value`? No, it's always `boolean`

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
kapetr: I guess an empty array doesnt make sense here. Will update.

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
kapetr: I'd keep it required - the developer should specify the default.

apps/beeai-ui/src/modules/runs/settings/SingleSelectSettingsField.module.scss
kapetr: I updated the style to ensure only one item is highlighted at a time, disabling the selected highlight on hover.

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
PetrBulanek: I thought that you just don't have to specify it, and then it actually fallbacks to "false". But it probably doesn't matter. :-)

PetrBulanek (approved):

apps/beeai-ui/src/api/a2a/extensions/ui/settings.ts
kapetr: I understand, but this follows sdk definition, where it has a default value `false`, so allowing undefined here would be redundant in my opinion.

PR #1150, opened 2025-09-08
feat(server, sdk): add context history

jezekra1:


apps/beeai-server/src/beeai_server/api/routes/contexts.py
```diff
@@ -86,3 +96,29 @@ async def generate_context_token(
...
+async def list_context_history(
+    context_id: UUID,
+    context_service: ContextServiceDependency,
+    user: Annotated[AuthorizedUser, Depends(RequiresContextPermissionsPath(context_data={"read"}))],
+) -> PaginatedResult[ContextHistoryItem]:
+    # Cannot list history for a different context
+    if user.context_id and user.context_id != context_id:
+        raise fastapi.HTTPException(status.HTTP_403_FORBIDDEN, "Insufficient permissions")
+    history = [item async for item in context_service.list_history(context_id=context_id, user=user.user)]
+    return PaginatedResult(items=history, total_count=len(history))
```
pilartomas: Pagination TODO?

apps/beeai-sdk/src/beeai_sdk/server/store/context_store.py
```diff
@@ -0,0 +1,53 @@
...
+    if not metadata:
+        return metadata
+    return {k: v for k, v in metadata.items() if k in ALLOWED_METADATA_EXTENSION_URIS}
+
+
+async def record_event(event: Event, context_store: ContextStoreInstance):
+    # TODO: we filter metadata because they may contain sensitive information and auth tokens
+    match event:
+        case Message() as msg:
+            await context_store.store(msg.model_copy(update={"metadata": filter_metadata(msg.metadata)}))
```
pilartomas, line 53: The filter will be problematic for general purpose use.

I'd personally not filter anything at the moment. In principle, I'd spect all passing secrets to be short-lived as a best practice.

apps/beeai-sdk/src/beeai_sdk/server/store/platform_context_store.py
```diff
@@ -0,0 +1,48 @@
...
+
+class PlatformContextStoreInstance(ContextStoreInstance):
+    def __init__(self, context_id: str, platform_extension: PlatformApiExtensionServer):
+        self._context_id = context_id
+        self._platform_extension = platform_extension
+
+    async def load_history(self) -> AsyncIterator[Message | Artifact]:
+        async with self._platform_extension.use_client():
+            for history_item in await Context.list_history(self._context_id):
+                yield history_item
```
pilartomas: I didn't track it all the way but I am quite certain pagination interface is in place. Therefore, there is no need for this "bottle-neck" in code. Would be good to get this right to avoid issues once the actual pagination has been implemented.

apps/beeai-server/src/beeai_server/domain/models/common.py
```diff
@@ -32,3 +33,19 @@ def validate_metadata(metadata: dict[str, str]) -> dict[str, str]:
     ),
     AfterValidator(validate_metadata),
 ]
+
+
+class PaginatedResult[T: BaseModel](BaseModel):
+    items: list[T]
+    total_count: int
+    has_more: bool = False
+
```
pilartomas, line 42: Cursor-based pagination generally returns the cursor explicitly to make it clear what to fetch next.

jezekra1: that's the `last_id` computed field below, should I name it differently?

apps/beeai-sdk/src/beeai_sdk/server/store/platform_context_store.py
jezekra1: hm... I guess it makes sense to implement pagination properly even for this PR and then fetch all pages in `load_context`, will help avoid loading too many objects in memory on the server.

Will update this on Monday

apps/beeai-server/src/beeai_server/api/routes/contexts.py
jezekra1: Makes sense to paginate the history as well, it can actually be really long, I'll add it in this PR.

For most other endpoints, pagination is TODO

apps/beeai-sdk/src/beeai_sdk/server/store/context_store.py
jezekra1: it'll store a lot of unnecessary data though, I can reverse the filter and keep everything but filter out at least LLM/embedding and platform api extension (they send only tokens and urls)

apps/beeai-server/src/beeai_server/domain/models/common.py
pilartomas: Common practice is a cursor field with opaque value (usually ID of the next item).

There is a lack of standardization in REST world but we should aim for such an implementation that common high-level fetch clients (e.g. React Query) will work out of the box. @PetrBulanek @kapetr thoughts?

PetrBulanek: Both patterns (page or cursor) work with React Query we use. However, [useInfiniteQuery](https://tanstack.com/query/latest/docs/framework/react/reference/useInfiniteQuery) is designed around the cursor approach (getNextPageParam just reads the token), so that tends to feel more natural and requires less client-side logic.

jezekra1: Not sure I understand :D so renaming `last_id` -> `cursor` will do the trick?

pilartomas: That would be part of it but also the semantics will be a bit different. The goal is for the structure to be easily mappable to common paginators like the one mentioned above, since there is no standard.

pilartomas: Did a more research on this and I'd say I'm overthinking this. At best, I can identify is a common pattern but nothing what would approach some sort of standard.

jezekra1: renamed `last_id` to `next_page_token` and `after` to `page_token`, inspired by a2a spec

apps/beeai-sdk/src/beeai_sdk/server/store/platform_context_store.py
jezekra1: pagination for history added

apps/beeai-sdk/src/beeai_sdk/server/store/context_store.py
jezekra1: reversed the logic as decribed

pilartomas (approved):

PR #1159, opened 2025-09-08
Cookie auth fix

tedhabeck:
Add middleware for openapi-fetch, updated server auth.py to provide token on session api so it could be used by `openapi-fetch`, updated authorized_user routine to process cookie_auth and fixed an exception in server `decode_oauth_jwt_or_introspect` API.


apps/beeai-ui/src/api/index.ts
```diff
@@ -3,12 +3,35 @@
...
+      const authRes = await auth();
+      if (authRes?.id_token) {
+        idToken = authRes.id_token;
+      }
+    }
+    // add Authorization header to every request
+    if (idToken) {
+      request.headers.set('Authorization', `Bearer ${idToken}`);
+    }
+    return request;
```
pilartomas: The server expects the access token not the id token. Generally, there will be audience mismatch and missing scopes in the ID token so the validation will fail.

pilartomas (changes requested):

apps/beeai-server/src/beeai_server/api/auth.py
```diff
@@ -166,7 +166,7 @@ async def introspect_token(token: str, configuration: Configuration) -> tuple[di
 async def decode_oauth_jwt_or_introspect(
     configuration: Configuration, token: str, jwks_dict: JwksDict | None = None, aud: str | None = None
 ) -> tuple[dict | None, str | None]:
-    if jwks_dict:
+    if jwks_dict and token:
```
jezekra1, line 169: [nitpick] if `token` is `None` it would be better to not even call the decode function (and the parameter type is wrong)

apps/beeai-ui/src/api/index.ts
```diff
@@ -3,12 +3,35 @@
...
+    // fetch token
+    if (OIDC_ENABLED) {
+      const authRes = await auth();
+      if (authRes?.id_token) {
+        idToken = authRes.id_token;
+      }
+    }
+    // add Authorization header to every request
+    if (idToken) {
+      request.headers.set('Authorization', `Bearer ${idToken}`);
```
jezekra1: if we are going to use headers, drop the cookie auth please

kapetr: Then we'll probably need to add the authorization header to `A2AClient` (in `src/api/a2a/client.ts`) as well. That should be possible by overriding fetch function in the `fromCardUrl` options.

apps/beeai-ui/src/api/index.ts
tedhabeck: A few things here.  First and foremost, both the cookie and the token are id_tokens.  If we toss the ID token, we have NO NextJS session at all (i.e. the useSession() hook will break completely) so asking for this change is not going to fly.

Also, validation is not failing, so I'm not sure where that's coming from.   The audience is specified by the odic provider (in our case IBM Security Verify).   The server is validating the token fine.

Can you elaborate on the scopes issue?  I'm not seeing this issue at all.

apps/beeai-ui/src/api/index.ts
tedhabeck: > if we are going to use headers, drop the cookie auth please

The reason for supplying the token on the api is that openapi-fetch does not automatically send the cookie if it is a server-side request.   For client side requests, the cookie will be sent.  So both are required.

kapetr: We could add the Authorization header in the API proxy (`app/api/[...path]/route.ts`) to keep things consistent. In that case, the cookie could probably be dropped, or am I wrong?

apps/beeai-ui/src/api/index.ts
pilartomas: To my understanding, the current implementation is:

- Cookie - NextJS JWT (or JWE?) that is set up here to carry id and access token -> https://github.com/i-am-bee/beeai-platform/blob/fc4ac0152d6ee7fca270f2188bbb703f3ab8140d/apps/beeai-ui/src/auth.ts#L211-L212
- ID Token - used by NextJS to get user info
- Access Token - should be send to the platform server because that's the bearer that grants access (to platform server but also user info)

I am not suggesting tossing the ID Token away (web might still need it) just forwarding the proper token to the platform server.

> Also, validation is not failing, so I'm not sure where that's coming from.

Probably because audiences of the tokens match. It wouldn't be the case if web was running e.g. at `web.example.com` and platform server at `api.example.com`. In that case the platform server should reject the ID token for invalid audience.

Generally, the **ID Token** is NOT a bearer and the audience is strictly the client that requests it (NextJS server in our case). When the resource server needs the same verified data, it uses the **Access Token** with `id_token` scope against the `/userinfo` endpoint on the auth server. This has been already implemented by @abhi201191 - https://github.com/i-am-bee/beeai-platform/blob/fc4ac0152d6ee7fca270f2188bbb703f3ab8140d/apps/beeai-server/src/beeai_server/api/dependencies.py#L83.

In practice this isn't always possible, depends on the issuer (e.g. Google is problematic). But in our case, we should follow oAuth strictly and the ISV is compliant as far as I know.

Also I think the **Refresh Token** won't work with the identity token, it relates to the access token only.

apps/beeai-ui/src/utils/constants.ts
```diff
@@ -33,6 +33,6 @@ export const COMPOSE_LINK = `${DOCUMENTATION_LINK}/how-to/compose-agents`;
 
 export const LF_PROJECTS_LINK = 'https://lfprojects.org/';
 
-export const OIDC_ENABLED = process.env.OIDC_ENABLED === 'true';
+export const OIDC_ENABLED = process.env.NEXT_PUBLIC_OIDC_ENABLED === 'true';
```
kapetr, line 36: Please update `.env.example`

apps/beeai-ui/src/auth.ts
```diff
@@ -166,22 +192,9 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
...
-      // const jwt = { access_token: params.token || '' };
-      let jwt: JWT | null = null;
-      jwt = await internalDecode(params);
-      return jwt;
-    },
-  },
   cookies: {
     sessionToken: {
-      name: `beeai-platform`,
+      name: `beeai-platform-1`,
```
kapetr, line 197: Debug?

apps/beeai-ui/src/app/(main)/page.tsx
```diff
@@ -19,11 +19,12 @@ export const dynamic = 'force-dynamic';
...
+  // only force login if oidc is enabled.
   if (OIDC_ENABLED) {
-    const session = await auth();
+    session = await getCurrentSession();
     if (!session?.user) {
       await connection();
-      // only force login if oidc is enabled.
       redirect(routes.login());
     }
   }
```
kapetr, line 30: I think this logic might be redundant, since NextAuth should already handle the redirect. Could we remove it?

apps/beeai-ui/src/app/api/[...path]/route.ts
```diff
@@ -29,6 +29,11 @@ async function handler(request: NextRequest, context: RouteContext) {
   }
   targetUrl += search;
 
+  if (OIDC_ENABLED) {
+    // ensure the beeai-platform cookie is set with an access token.
+    await getCurrentSession(request);
```
Comment (kapetr) [inline apps/beeai-ui/src/app/api/[...path]/route.ts:34]: Why not do this?
```typescript
if (OIDC_ENABLED) {
    const session = await auth();
    if (session?.access_token) {
      headers.set('Authorization', `Bearer ${session?.access_token}`);
    }
  }
```

We could then remove `getCurrentSession` and just send the access token in the headers.

kapetr (changes requested):

apps/beeai-ui/src/app/(main)/page.tsx
pilartomas: Agreed, it would be best not to send the access token to the browser unencrypted, which I assume is a side-effect of that getter. (NextAuth keeps its jwt cookie encrypted iirc)

kapetr (approved):
Approving so this can make it into the upcoming release. The requested changes will be addressed in #1149

PR #1170, opened 2025-09-10
feat(web): add auto-generated ToC for blog posts

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/1126


apps/beeai-web/src/hooks/useSections.ts
```diff
@@ -0,0 +1,125 @@
...
+      resizeObserver.unobserve(scrollableContainer);
+    };
+  }, [scrollableContainer, debouncedHandleResize]);
+
+  useEffect(() => {
+    if (!scrollableContainer) {
+      return;
+    }
+
+    scrollableContainer.addEventListener('scroll', handleScroll);
```
kapetr: `handleScroll` isn’t an expensive function, but it might still be good practice to throttle it and not let every scroll event through.

PR #1175, opened 2025-09-11
feat(ui): add auto_resize for TextField form input

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/1153


apps/beeai-ui/src/modules/form/components/FormLabel.tsx
kapetr, line 1: Isn’t this already what FormLabel from Carbon does?

PetrBulanek: Essentially, yes, but the Carbon component also has a class `cds--label--no-margin`, which cannot be removed using any prop, so I thought it would be better to create my own component.

PR #1176, opened 2025-09-11
feat: implementation of Embedding extension in GUI

tomkis:


apps/beeai-ui/src/modules/platform-context/api/index.ts
```diff
@@ -13,10 +13,10 @@ export async function createContext() {
   return ensureData(response);
 }
 
-export async function matchProviders(suggestedModels: string[]) {
+export async function matchProviders(suggestedModels: string[], capability: 'llm' | 'embedding') {
```
PetrBulanek: I think we can create general type `ModelCapability` and import it here.

PR #1189, opened 2025-09-15
Enhancing CLI auth flow

abhi201191:


apps/beeai-cli/src/beeai_cli/commands/auth.py
```diff
@@ -128,10 +128,19 @@ async def exchange_token(oidc: dict, code: str, code_verifier: str, config) -> d
 
 @app.command("login")
 async def cli_login(resource_url: str | None = None):
+    # if config.auth_manager.config.get("resources"):
+    #     default_url = config.auth_manager.config.get("active_resource")
+    # else:
+    #     default_url = "beeai.res.ibm.com"
```
JanPokorny: Commented out code

apps/beeai-cli/src/beeai_cli/commands/auth.py
```diff
@@ -128,10 +128,19 @@ async def exchange_token(oidc: dict, code: str, code_verifier: str, config) -> d
...
+    if not default_url:
+        default_url = "beeai.res.ibm.com"
     if not resource_url:
-        entered = input(f"Enter the server url (default: {config.host}):").strip()
-        resource_url = entered or str(config.host)
-
+        entered = input(f"Enter the server address (default: {default_url}):").strip()
+        resource_url = entered or str(default_url)
+    resource_url = normalize_url(resource_url)
+    print(resource_url)
```
JanPokorny: Debug print

mise.lock
```diff
@@ -1,4 +1,4 @@
-[[tools."asdf:CrouchingMuppet/asdf-lima"]]
```
JanPokorny: You have an old version of `mise`. Please do not update the lockfile with an old version.

apps/beeai-cli/src/beeai_cli/__init__.py
```diff
@@ -26,6 +27,9 @@
 app.add_typer(beeai_cli.commands.mcp.app, name="mcp", no_args_is_help=True, help="Manage MCP servers and toolkits.")
 app.add_typer(beeai_cli.commands.build.app, name="", no_args_is_help=True, help="Build agent images.")
 app.add_typer(beeai_cli.commands.auth.app, name="", no_args_is_help=True, help="Beeai login.")
+app.add_typer(
```
JanPokorny, line 28: Having the login logic split between `auth` and `resource_server` command modules is messy. Ideally we should make this a single CLI namespace, like `beeai server login`. Also, it seems that in code, the names "resource" and "resource server" and "server" are used interchangeably, and since all seem to refer to the same thing, we should unify on something clear like `server` or `host`.

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
```diff
@@ -0,0 +1,87 @@
...
+                marker = (
+                    " [active]"
+                    if (
+                        resource_name == config.auth_manager.config.get("active_resource")
+                        and issuer == config.auth_manager.config.get("active_token")
+                    )
+                    else ""
+                )
+                console.print(f"{j}. {issuer}{marker}")
+            choice = input("\nSelect an authorization server: ").strip()
```
JanPokorny: Use `InquirerPy` for interactive prompts. See other modules for examples.

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
```diff
@@ -0,0 +1,87 @@
...
+app = AsyncTyper()
+
+config = Configuration()
+
+
+@app.command("show")
+def show_server():
+    active = config.auth_manager.get_active_resource()
+    if not active:
+        console.print("No active server!!!\n")
```
JanPokorny: Use `console.error`, `console.success` etc. for messages, see other modules for examples. (Applies to most `console.print` calls here.)

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
```diff
@@ -0,0 +1,87 @@
...
+    for res in resources:
+        marker = " (active)" if res == config.auth_manager.get_active_resource() else ""
+        console.print(f"{res}{marker}\n")
+
+
+@app.command("switch")
+def switch_server():
+    resources = config.auth_manager.config.get("resources", {})
+    if not resources:
+        console.print("No server logged in. Run `beeai login` first.\n")
```
JanPokorny: Use consistent formatting with other console messages, e.g. `[green]...[/green]` for commands.

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
```diff
@@ -0,0 +1,87 @@
...
+
+    resource_names = list(resources.keys())
+
+    console.print("Available servers:")
+    for i, res in enumerate(resources, start=1):
+        marker = " (active)" if res == config.auth_manager.get_active_resource() else ""
+        console.print(f"{i}. {res}{marker}")
+
+    choice = input("Select a server number: ")
```
JanPokorny: Use `InquirerPy` for interactive prompts. See other modules for examples.

apps/beeai-cli/src/beeai_cli/api.py
```diff
@@ -19,9 +19,14 @@
...
 from beeai_cli.configuration import Configuration
+from beeai_cli.utils import normalize_url
 
 config = Configuration()
-BASE_URL = str(config.host).rstrip("/")
+BASE_URL = (
+    normalize_url(config.auth_manager.get_active_resource())
+    if config.auth_manager.get_active_resource()
+    else config.default_host
+)
```
JanPokorny:
```suggestion
BASE_URL = (
    normalize_url(res)
    if (res := config.auth_manager.get_active_resource())
    else config.default_host
)
```

apps/beeai-cli/src/beeai_cli/utils.py
```diff
@@ -260,3 +260,11 @@ def verbosity(verbose: bool, show_success_status: bool = True):
...
     parsed = urlparse(resource_url)
     return parsed.netloc or parsed.path
+
+
+def normalize_url(url: str) -> str:
+    if not url.startswith(("http://", "https://")):
+        url = "https://" + url
+    parsed = urlparse(url)
+    # ensure no trailing slash
+    return f"{parsed.scheme}://{parsed.netloc}"
```
JanPokorny, line 271: Why have this as a helper instead of doing input validation? Also, why are we dropping the path and only using the host part of the URL?

apps/beeai-cli/src/beeai_cli/__init__.py
JanPokorny: Alternatively we could call it `context` like Docker

apps/beeai-cli/src/beeai_cli/commands/auth.py
abhi201191: Removed

mise.lock
abhi201191: Done

apps/beeai-cli/src/beeai_cli/__init__.py
abhi201191: Merged into single module `server login`, `server change` etc

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
abhi201191: Done

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
abhi201191: This is not exactly an error. Using print only with proper formatting

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
abhi201191: Done

apps/beeai-cli/src/beeai_cli/commands/resource_server.py
abhi201191: Done

apps/beeai-cli/src/beeai_cli/api.py
abhi201191: Changed

apps/beeai-cli/src/beeai_cli/utils.py
abhi201191: This is not exactly a validation. 
For example, in beeai server login, user can provide the whole url or just the server name (`agents.res.ibm.com`), this helper function will normalize it.

JanPokorny: @abhi201191 We can simplify by just requiring a valid URL and validating it with Pydantic.

PR #1193, opened 2025-09-15
feat(ui): cancel task with input-required on new message

kapetr:


apps/beeai-ui/src/modules/runs/chat/ChatMessagesView.tsx
```diff
@@ -48,7 +48,7 @@ export function ChatMessagesView() {
                   <li key={message.id} className={classes.message}>
                     {isUser && <ChatUserMessage message={message} />}
 
-                    {isAgent && <ChatAgentMessage message={message} />}
+                    {isAgent && <ChatAgentMessage message={message} onShow={isLast ? scrollToBottom : undefined} />}
```
kapetr, line 51: The previous solution (below) didn’t scroll down when something was awaited before the agent message was created, which can happen during the submit. The await was not used eventually, but I kept the fix since it feels more robust.

PR #1197, opened 2025-09-16
chore: update docs after OIDC_ENABLED env rename

kapetr:


kapetr:
Env rename was reverted back to the original name.

PR #1204, opened 2025-09-16
feat: update beeai-framework

Tomas2D:


mise.lock
```diff
@@ -2,77 +2,50 @@
 version = "1.2.1"
 backend = "asdf:CrouchingMuppet/asdf-lima"
 
-[tools.fd]
+[[tools.fd]]
```
jezekra1: this is older mise lock format, please update mise or exclude this from the PR

jezekra1 (changes requested):

mise.lock
jezekra1: turns out this is the new format, someone probably commited older version before already

jezekra1 (approved):

PR #1227, opened 2025-09-20
fix(server): update loop type annotation in Server class

iChintanSoni:


jezekra1:
Thanks, I fixed it in with a more broad type compatible with both version #1232

PR #1230, opened 2025-09-22
chore(server): oauth discovery, authlib and caching

pilartomas:
- Implement issuer metadata discovery
- Implement JWKS discovery
- Use Authlib
- Add caching to discovery and remote validation
- Remove cookie auth, only support bearer


apps/beeai-server/src/beeai_server/api/dependencies.py
```diff
@@ -63,35 +60,33 @@ async def authenticate_oauth_user(
...
+    try:
+        claims = (
+            claims
+            if claims and "email" in claims and "email_verified" in claims
+            else await get_user_info(token, provider=provider)
         )
+        email = claims.get("email")
+        email_verified = claims.get("email_verified")
+        if not isinstance(email, str) or not isinstance(email_verified, (str, bool)) or not bool(email_verified):
+            raise RuntimeError("Email not verified")
```
jezekra1, line 87: This will get converted to status code 500, I'd raise `HttpException` with status 401 here.

pilartomas: I think it is 401, check line 89.

PR #1235, opened 2025-09-23
feat(ui): add session history

PetrBulanek:
Refs: https://github.com/i-am-bee/beeai-platform/issues/623

- Added session history support
- Form & auth UI extensions not implemented yet (no agents to test)
- Single-turn agent history not implemented yet (no agents to test)
- Pagination for messages not implemented yet (expecting changes on API to return whole agent/user messages in the future)


apps/beeai-ui/src/components/SidePanel/NavItem.module.scss
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.button:global(.cds--btn) {
+  @include reset-focus();
+  justify-content: flex-start;
+  align-items: center;
+  gap: rem(8px);
+  inline-size: 100%;
+  padding: $spacing-03 - rem(1px);
```
kapetr, line 12: Just curious, why the -1px?

apps/beeai-ui/src/components/SidePanel/NavItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <li>
+      <Button
+        kind="ghost"
+        size="sm"
+        className={clsx(classes.button, { [classes.isActive]: isActive })}
+        onClick={onClick}
+      >
+        {label}
+
+        {isExternal && <ArrowUpRight />}
```
kapetr: Not a subject of this PR, but this doesn't feel right - having external link on a button without actually providing href (the url is actually opened via `window.open` in `onClick`, defined in `CustomNav`)

apps/beeai-ui/src/components/layouts/SideNav.module.scss
```diff
@@ -3,11 +3,35 @@
...
+  min-block-size: 0;
+}
+
+.nav {
+  flex-shrink: 0;
+}
+
+.navBody {
+  max-block-size: rem(186px);
+}
```
kapetr, line 29: Is this correct, do we want to limit the height to this number? I'd consider a different name to make it clearer it's a style applied to the first group in the navigation panel.

apps/beeai-ui/src/hooks/useFetchNextPage.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+    },
+    ...inViewProps,
+  });
+
+  useEffect(() => {
+    if (inViewReturn.entry?.isIntersecting && !isFetching && hasNextPage) {
+      fetchNextPage();
+    }
+  }, [hasNextPage, inViewReturn.entry?.isIntersecting, isFetching, fetchNextPage]);
```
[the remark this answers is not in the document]
kapetr: If I remember correctly, this dealt with a case onChange callback didn't cover, we should probably provide explanation comment, why it is necessary.

apps/beeai-ui/src/modules/feedback/api/utils.ts
```diff
@@ -31,6 +31,6 @@ export function createSendFeedbackPayload({
     message: getMessageRawContent(message),
     provider_id: agent.provider.id,
     context_id: contextId,
-    task_id: message.id,
+    task_id: message.taskId!,
```
kapetr: 👍

kapetr (changes requested):
I see the main issue as the management of the context ID. Right now, there are multiple sources of truth for the value (url parameter, `history-context`, `platform-context`), which makes it almost impossible to avoid bad states with this setup. In my opinion, the primary source of truth should be the URL parameter, so I’d remove the `contextId` state from `PlatformContextProvider` and move it up to the `HistoryProvider` - some renaming of these contexts might be needed. It’s still not trivial from here, but I think it would be a better starting point.

apps/beeai-ui/src/components/SidePanel/NavItem.module.scss
PetrBulanek: The padding should be 8px, but there is already a 1px border, so I extract it.

apps/beeai-ui/src/components/SidePanel/NavItem.tsx
PetrBulanek: Yeah, I agree. I would definitely prefer to use the native approach. It would make sense not to have Button at all, but a classic Link with href.

apps/beeai-ui/src/components/layouts/SideNav.module.scss
PetrBulanek: In reality, this sets the maximum height to the visibility of five items, which seemed reasonable to me. Perhaps it should be explicitly visible there by some calculation and not a magic number. I'll also try to come up with a better name for the class.

apps/beeai-ui/src/modules/feedback/api/utils.ts
PetrBulanek: I'm not entirely sure, it's a copy-paste from bee-ui. :-)

apps/beeai-ui/src/hooks/useFetchNextPage.ts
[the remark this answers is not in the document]
PetrBulanek: I'm not entirely sure, it's a copy-paste from bee-ui. :-)

kapetr (approved):

PR #1238, opened 2025-09-23
feat(ui): finish runtime secrets ui

kapetr:


apps/beeai-ui/src/modules/runs/secrets/SecretTag.module.scss
PetrBulanek, line 1: It doesn't look good in dark mode.

image

apps/beeai-ui/src/modules/agents/components/AgentSecrets.tsx
```diff
@@ -0,0 +1,42 @@
...
+                size="md"
+                onClick={() =>
+                  openModal((props) => <SecretsAddModal {...props} secret={secret} updateSecret={updateSecret} />)
+                }
+              />
+            </li>
+          ))}
+        </ul>
+      ) : (
+        <p className={classes.empty}>This agent does not have any secrets</p>
```
PetrBulanek: I think this should be styled the same as "no tools" message. We could make it a reusable component.

image
image

apps/beeai-ui/src/modules/agents/components/AgentSecrets.module.scss
```diff
@@ -0,0 +1,22 @@
...
+
+  li {
+    display: flex;
+    flex-direction: column;
+    padding-block: $spacing-04 $spacing-05;
+    gap: $spacing-03;
+    align-items: flex-start;
+    p {
+      @include type-style(label-01);
+      font-size: rem(12px);
```
PetrBulanek: I think, label-01 has the same font-size you are setting here.

apps/beeai-ui/src/modules/form/components/MessageSecretsForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { register, handleSubmit } = useForm({ mode: 'onChange' });
+
+  if (!secretPart) {
+    return null;
+  }
+
+  const onSubmit = async (values: FormValues) => {
+    storeSecrets(values);
+
+    const secretsFulfillment = Object.entries(secretPart.secret.secret_demands).reduce<AgentRequestSecrets>(
```
PetrBulanek: Just nitpicking - `Object.entries(secretPart.secret.secret_demands)` is called twice, so I'd make it a variable.

apps/beeai-ui/src/modules/form/components/MessageSecretsForm.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const isLastMessage = messages.at(-1)?.id === message.id;
+
+  return (
+    <form onSubmit={handleSubmit(onSubmit)}>
+      <fieldset disabled={!isLastMessage} className={classes.root}>
+        {Object.entries(secretPart.secret.secret_demands).map(([key, { name, description }]) => {
+          return (
+            <div key={key} className={classes.demand}>
+              <p>{description}</p>
+              <PasswordInput id={`${id}:${key}`} labelText={name} {...register(key, { required: true })} autoFocus />
```
PetrBulanek: Autofocus should probably only be on the first input.

apps/beeai-ui/src/modules/runs/contexts/agent-secrets/index.ts
```diff
@@ -0,0 +1,12 @@
...
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { use } from 'react';
+
+import { AgentSecretsContext } from './agent-secrets-context';
+
+export function useAgentSecrets() {
+  return use(AgentSecretsContext);
```
PetrBulanek, line 11: No `if (!context)` check. Is this just temporary because of ComposeProvider, or is it intentional?

apps/beeai-ui/src/modules/agents/components/AgentSecrets.tsx
```diff
@@ -0,0 +1,42 @@
...
+              <p>{secret.description}</p>
+
+              <SecretTag
+                secret={secret}
+                size="md"
+                onClick={() =>
+                  openModal((props) => <SecretsAddModal {...props} secret={secret} updateSecret={updateSecret} />)
+                }
+              />
+            </li>
```
PetrBulanek, line 23: This is quite the same as `SecretCard` component. Perhaps we could somehow reuse it?

apps/beeai-ui/src/modules/runs/secrets/EditSecretForm.module.scss
```diff
@@ -0,0 +1,23 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.root {
+  display: flex;
+  flex-direction: column;
+  max-inline-size: rem(544px);
```
PetrBulanek: I don't think this is needed, the width is limited by the modal itself. Unless I'm overlooking something.

apps/beeai-ui/src/modules/runs/secrets/EditSecretForm.tsx
```diff
@@ -0,0 +1,67 @@
...
+        Submit
+      </Button>
+    </form>
+  );
+}
+
+export enum EditSecretFormVariant {
+  Modal = 'modal',
+  Inline = 'inline',
+}
```
PetrBulanek: Probably some leftover. I guess, the max-inline-size mentioned above was for the inline variant, right?

apps/beeai-ui/src/modules/runs/contexts/agent-secrets/AgentSecretsProvider.tsx
```diff
@@ -0,0 +1,157 @@
...
+      return secretsSchema.parse(JSON.parse(value));
+    } catch (error) {
+      console.warn('Failed to parse agent secrets from localStorage', error);
+      return {};
+    }
+  },
+};
+
+export function AgentSecretsProvider({ agent, agentClient, children }: PropsWithChildren<Props>) {
+  const [agentSecrets, setAgentSecrets] = useLocalStorage<Secrets>('agent-secrets', {}, secretsLocalStorageOptions);
```
PetrBulanek: I think we should prefix the localStorage key in a similar way we do for example for setting theme (`@i-am-bee/beeai/THEME`).

apps/beeai-ui/src/modules/runs/contexts/agent-secrets/AgentSecretsProvider.tsx
```diff
@@ -0,0 +1,157 @@
...
+            ...prevAgentValue,
+            secrets: { ...prevAgentValue?.secrets, ...secrets },
+          },
+        };
+      });
+    },
+    [agent.provider.id, setAgentSecrets],
+  );
+
+  const revokeSecret = useCallback(
```
PetrBulanek: I don't see this implemented anywhere. There's no way to revoke the secret now, right? Is this being considered for a future iteration?

apps/beeai-ui/src/modules/runs/contexts/agent-secrets/index.ts
kapetr: Yes, I'd considered that temporary, until the usage in ComposeProvider is figured out. Now the context is filled with empty defaults. so a little bit different approach then we use, but works too.

apps/beeai-ui/src/modules/runs/contexts/agent-secrets/AgentSecretsProvider.tsx
kapetr: That's from initial UI components dev. It might have been in the designs somewhere? I'd remove it, I added the possbility to edit existing keys.

PR #1249, opened 2025-09-29
feat(platform): allow building images in external clusters

jezekra1:


helm/templates/config/provider_templates.yaml
```diff
@@ -93,10 +102,14 @@ data:
     spec:
       backoffLimit: 0
       activeDeadlineSeconds: {{`{{ job_timeout_seconds }}`}}
-      ttlSecondsAfterFinished: 600
+      ttlSecondsAfterFinished: 60
```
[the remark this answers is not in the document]
jezekra1: that's on purpose for security reasons

apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -266,6 +284,33 @@ def _oci_registry_defaultdict(self):
...
+    def _set_default_provider_build_values(self):
+        self.provider_build.k8s_namespace = self.provider_build.k8s_namespace or self.k8s_namespace
+        self.provider_build.k8s_kubeconfig = self.provider_build.k8s_kubeconfig or self.k8s_kubeconfig
+        self.provider_build.manifest_template_dir = (
+            self.provider_build.manifest_template_dir or self.provider.manifest_template_dir
+        )
+        return self
+
+    @model_validator(mode="after")
+    def _parse_kubeconfig(self):
```
JanPokorny: Having the same prop mean "path to kubeconfig file" and "dict of kubeconfig data" at different points is weird, wouldn't a cached property work better here?

apps/beeai-server/src/beeai_server/bootstrap.py
```diff
@@ -29,15 +30,14 @@ def setup_database_engine(config: Configuration) -> AsyncEngine:
...
+async def setup_kubernetes_client(namespace: str | None = None, kubeconfig: pathlib.Path | str | dict | None = None):
     if namespace is None:
-        ns_path = Path("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+        ns_path = anyio.Path("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
         if await ns_path.exists():
             namespace = (await ns_path.read_text()).strip()
 
     async def api_factory():
-        return await kr8s.asyncio.Api(bypass_factory=True, namespace=namespace, kubeconfig=str(config.k8s_kubeconfig))
+        return await kr8s.asyncio.Api(bypass_factory=True, namespace=namespace, kubeconfig=kubeconfig)
```
JanPokorny, line 40: Is this correct with `dict` kubeconfig? The kr8s docs don't mention this possibility

apps/beeai-server/src/beeai_server/configuration.py
JanPokorny: Also, I am not sure if loading kubeconfig as dict is necessary when kr8s can load it directly from a file path

JanPokorny (approved):

apps/beeai-server/src/beeai_server/configuration.py
jezekra1: I made it the same hybrid type that kr8s accepts, so that I can load the secret as env rather than a file, because I thought there were more security implications with files, but I'll refactor it to accept just Path, I think there are no risks with mounting secrets as files after all.

PR #1256, opened 2025-09-29
feat(ui): refactor platform context control 

kapetr:


apps/beeai-ui/src/modules/platform-context/api/types.ts
```diff
@@ -11,6 +11,7 @@ type CreateContextRequest = ApiRequest<'/api/v1/contexts'>;
 export type CreateContextParams = CreateContextRequest & {
   metadata: Required<Pick<ContextMetadata, 'agent_name' | 'provider_id'>>;
 };
+export type CreateContextResponse = ApiResponse<'/api/v1/contexts/{context_id}'>;
```
PetrBulanek: This is actually `ReadContextResponse` as it internally gets the response for GET. In fact, it's the same object in the end, so it probably doesn't matter, and using the POST is a bit verbose imho:

```export type CreateContextResponse = ApiResponse<'/api/v1/contexts', 'post', 'application/json', 201>;```

File: apps/beeai-ui/src/modules/platform-context/contexts/PlatformContextProvider.tsx [hunk 2]
```diff
@@ -2,197 +2,68 @@
...
 }
 
-export function PlatformContextProvider<UIGenericPart>({
-  agentClient,
-  agent,
-  contextId: contextIdProp,
-  children,
-}: PropsWithChildren<Props<UIGenericPart>>) {
-  const { featureFlags } = useApp();
+export function PlatformContextProvider({ contextId: contextIdProp, history, children }: PropsWithChildren<Props>) {
```
PetrBulanek: I think we could expand the contextId type to `contextId?: string | null;`, make null the default param value `contextId: contextIdProp = null` and get rid of `contextIdProp ?? null` in the both useState and useEffect calls.

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -263,16 +269,12 @@ function AgentRunProvider({
         pendingRun.current = undefined;
         pendingSubscription.current = undefined;
 
-        if (getMessages().length === 2) {
-          queryClient.invalidateQueries({ queryKey: contextKeys.lists() });
-        }
-
+        queryClient.invalidateQueries({ queryKey: contextKeys.lists() });
```
PetrBulanek, line 287: Is it necessary to call this invalidation after each run? The conditionally scoped call during the first iteration could be enough, or am I missing anything?

apps/beeai-ui/src/modules/platform-context/contexts/PlatformContextProvider.tsx
kapetr: I removed `contextId` from the props in the final solution.

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -169,8 +171,8 @@ function AgentRunProvider({
     setInput(undefined);
     pendingRun.current = undefined;
 
-    resetContext(agent);
```
PetrBulanek: This seems like a mistake. We still need to call the resetContext here, don't we?

PetrBulanek (changes requested):

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
kapetr: We don't, it's reset via url parameter.

PR #1268, opened 2025-10-01
feat: support Unpack when extracting dependencies

Tomas2D:


jezekra1 (changes requested):
Please add a test involving a dependency with `Unpack`

PR #1298, opened 2025-10-07
feat(ui): show message form submission

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/1152

### DIFF
--- a/apps/beeai-ui/src/modules/form/components/fields/MultiSelect.module.scss
+++ b/apps/beeai-ui/src/modules/form/components/fields/MultiSelect.module.scss
@@ -21,6 +21,7 @@
     &:disabled {
       cursor: not-allowed;
       color: $text-disabled;
+      border-color: $field;
       background-color: $field;
     }
     &:global(.cds--tag--selected) {


apps/beeai-ui/src/modules/form/components/fields/FielFieldValue.module.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.files {
+  > li {
+    max-inline-size: calc(50% - #{math.div($spacing-04, 2)});
```
kapetr: It'd be good to create a variable, that would explain the calculated value.

apps/beeai-ui/src/modules/form/components/fields/MultiSelectField.tsx
```diff
@@ -47,7 +47,7 @@ export function MultiSelectField({ field }: Props) {
                   <OperationalTag
                     text={label}
                     size="lg"
-                    className={clsx(classes.option, { [classes.isSelected]: isSelected })}
+                    className={clsx(classes.option, { [(classes.isSelected, 'cds--tag--selected')]: isSelected })}
```
kapetr: The highlighting of selected item doesn't work because of this change. I think we should change this to `SelectableTag` and use `selected` prop.

PR #1299, opened 2025-10-07
feat(ui): history paging

kapetr:


apps/beeai-ui/src/modules/platform-context/api/queries/useListContextHistory.ts
```diff
@@ -45,7 +46,7 @@ export function useListContextHistory(params: Params) {
 
       return items;
     },
-    enabled: Boolean(contextId),
+    enabled: Boolean(contextId && initialData),
```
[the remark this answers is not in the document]
kapetr: I shifted this logic to the higher component and added comment why it is necessary.

apps/beeai-ui/src/modules/platform-context/api/queries/useListContextHistory.ts
```diff
@@ -33,9 +33,10 @@ export function useListContextHistory(params: Params) {
         },
       });
     },
-    initialPageParam: undefined,
-    getNextPageParam: (lastPage) =>
-      lastPage?.has_more && lastPage.next_page_token ? lastPage.next_page_token : undefined,
+    initialPageParam: initialData?.next_page_token ?? undefined,
```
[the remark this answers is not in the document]
kapetr: This is by design, we don't need to refetch the messages, they are immutable.

apps/beeai-ui/src/modules/runs/chat/ChatMessagesView.tsx
```diff
@@ -42,7 +45,7 @@ export function ChatMessagesView() {
               {messages.map((message, idx) => {
                 const isUser = isUserMessage(message);
                 const isAgent = isAgentMessage(message);
-                const isLast = idx == messages.length - 1;
+                const isLast = idx == 0;
```
PetrBulanek: I think we should add a comment here so it doesn't look like a mistake.

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -120,7 +104,7 @@ function AgentRunProvider({
   const updateCurrentAgentMessage = useCallback(
     (updater: (message: UIAgentMessage) => void) => {
       setMessages((messages) => {
-        const lastMessage = messages.at(-1);
+        const lastMessage = messages.at(0);
```
PetrBulanek: Same as above.

apps/beeai-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -181,7 +165,7 @@ function AgentRunProvider({
   }, []);
 
   const cancelPendingTask = useCallback(() => {
-    const lastMessage = getMessages().at(-1);
+    const lastMessage = getMessages().at(0);
```
PetrBulanek: Same as above.

PR #1289, opened 2025-10-06
feat(ui): add building provider from GitHub repo

PetrBulanek:
Refs: https://github.com/i-am-bee/beeai-platform/pull/1307

PR #1307, opened 2025-10-08
feat(server): agent build actions, agent origin tracking

jezekra1:
Ref: #1289 

Provider:
`id` (change) -> by default calculated from location but it is not strictly required anymore
`location` -> full docker image id or network URL (including tag), must be unique
`origin` (added) -> original source where the agent comes from (not unique):
- docker image base (`{registry}/{repository}`) - without tag
- github repo base (`git+{registry}/{repository}#{path}`) - without tag/commit version
Added a filter on provider list by origin -> lists all versions of the agent from the same repository.

So frontend when rendering github modal:
1. request POST `/provider_builds/preview` -> parses github url, checks github access, returns origin, destination
2. request GET `/providers?origin={origin}` -> lists existing agents from that repo
- 2b: if an agent from the same repo is found, UI renders single select:
    * update existing agent
    * add new agent
- 2c: if exactly the same agent in the same version is found (destination matches location of one of the existing agents) -> error message is displayed that the agent already exists
3. request POST `/provider_builds` with one of the following actions:
- `on_complete: {"type": "add_provider", ...} | {"type": "update_provider", "provider_id": ...}`

### DIFF
--- a/apps/beeai-cli/src/beeai_cli/commands/build.py
+++ b/apps/beeai-cli/src/beeai_cli/commands/build.py
@@ -149,9 +149,9 @@ async def server_side_build_experimental(
         build = await build.get()
         if build.status == BuildState.COMPLETED:
             if add:
-                message = "Agent added successfully. List agents using [bold]beeai list[/bold]"
+                message = "Agent added successfully. List agents using [green]beeai list[/green]"
             else:
-                message = f"Agent built successfully, add it to the platform using: [bold]beeai add {build.destination}[/bold]"
+                message = f"Agent built successfully, add it to the platform using: [green]beeai add {build.destination}[/green]"
             console.success(message)
         else:
             console.error("Agent build failed, see logs above for details.")
--- a/apps/beeai-server/src/beeai_server/domain/constants.py
+++ b/apps/beeai-server/src/beeai_server/domain/constants.py
@@ -18,6 +18,6 @@ Undefined: TypeAlias = Literal[_Undefined.undefined]  # noqa: UP040
 DEFAULT_AUTO_STOP_TIMEOUT: Final[timedelta] = timedelta(minutes=5)
 
 # A2A platform constants
-REQUIRED_ENV_EXTENSION_URI: Final[str] = "required_env"
+AGENT_DETAIL_EXTENSION_URI: Final[str] = "https://a2a-extensions.beeai.dev/ui/agent-detail/v1"
 
 MODEL_API_KEY_SECRET_NAME = "MODEL_API_KEY"
--- a/apps/beeai-sdk/src/beeai_sdk/a2a/extensions/ui/agent_detail.py
+++ b/apps/beeai-sdk/src/beeai_sdk/a2a/extensions/ui/agent_detail.py
@@ -22,6 +22,12 @@ class AgentDetailContributor(pydantic.BaseModel):
     url: str | None = None
 
 
+class EnvVar(pydantic.BaseModel):
+    name: str
+    description: str | None = None
+    required: bool = False
+
+
 class AgentDetail(pydantic.BaseModel, extra="allow"):
     interaction_mode: str | None = pydantic.Field("multi-turn", examples=["multi-turn", "single-turn"])
     user_greeting: str | None = None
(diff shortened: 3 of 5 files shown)


apps/beeai-server/src/beeai_server/domain/constants.py
```diff
@@ -1,10 +1,22 @@
...
 
 DOCKER_MANIFEST_LABEL_NAME: Final[str] = "beeai.dev.agent.json"
 
+
+class _Undefined(Enum):
+    undefined = "undefined"
+
+
+undefined = _Undefined.undefined
+Undefined: TypeAlias = Literal[_Undefined.undefined]  # noqa: UP040
```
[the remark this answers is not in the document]
jezekra1: wtf, these are different projects

apps/beeai-sdk/src/beeai_sdk/platform/provider.py
```diff
@@ -45,17 +47,71 @@ async def create(
...
+        payload = filter_dict(
+            {
+                "location": location,
+                "agent_card": agent_card.model_dump(mode="json") if agent_card else None,
+                "variables": variables,
+                "auto_stop_timeout_sec": auto_stop_timeout_sec,  # None means disable, but it is not allowed here
+            }
+        )
+        payload["origin"] = origin  # For origin, "None" means reset
+        payload = filter_dict(payload, undefined)
```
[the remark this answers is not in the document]
jezekra1: I don't find this more readable

apps/beeai-cli/src/beeai_cli/commands/build.py
```diff
@@ -136,18 +137,21 @@ async def server_side_build_experimental(
...
-        build = await ProviderBuild.create(location=github_url)
+        build = await ProviderBuild.create(location=github_url, on_complete=AddProvider() if add else None)
         async for message in build.stream_logs():
             print_log(message, ansi_mode=True)
         build = await build.get()
         if build.status == BuildState.COMPLETED:
-            console.success(
-                f"Agent built successfully, add it to the platform using: [bold]beeai add {build.destination}[/bold]"
-            )
+            if add:
```
JanPokorny, line 151: Just a small thing, other messages use green for commands to run, you could unify here

apps/beeai-server/src/beeai_server/api/schema/common.py
```diff
@@ -7,10 +7,10 @@
 
 
 class PaginationQuery(BaseModel):
-    limit: int = Field(default=40, ge=1, le=100)
+    limit: int = Field(default_factory=lambda: 40, ge=1, le=100)
```
JanPokorny, line 10: Huh, why is this necessary? Integers and strings are immutable

apps/beeai-server/src/beeai_server/domain/repositories/env.py
```diff
@@ -21,7 +21,7 @@ async def get(
         parent_entity: EnvStoreEntity,
         parent_entity_id: UUID,
         key: str,
-        default: str | None = NOT_SET,  # pyright: ignore [reportArgumentType]
+        default: str | None | Undefined = undefined,
```
JanPokorny, line 24: You could also use ellipsis as the sentinel value here, I've seen it used like that in Python

apps/beeai-server/src/beeai_server/api/schema/common.py
jezekra1: This is a hack, we might want to use it on more models, but I applied it only to the most common.

The openapi typescript generator treats fields with "default" in the openapi spec as required fields, unless you set this to false
```
  --properties-required-by-default
                             Treat schema objects as if `required` is set to all properties by default
```
but if you set this to false, then the response model fields become optional. This affected the UI in several places, including models from a2a sdk which we don't control, so I ended up not using the setting.

This is a fundamental issue in pydantic, contrary to zod there is no "input" and "output" schema (pre and post validation).

Finally, `default_factory` is not propagated to openapi spec as opposed to default, so this becomes correctly typed as optional in typescript :D 

I wanted to avoid duplicating all schemas to Input/output DTO, so we have this I admit a little hacky approach currently, together with `EntityModel` which fixes the problem for `id` fields of response models.

Example snippet:
```
      "UpdateProvider": {
        "properties": {
          "type": {
            "type": "string",
            "const": "update_provider",
            "title": "Type",
            "default": "update_provider"
          },
          "provider_id": {
            "type": "string",
            "format": "uuid",
            "title": "Provider Id"
          }
        },
        "type": "object",
        "required": [
          "provider_id"
        ],
        "title": "UpdateProvider",
        "description": "Will update provider specified by ID"
      },
```
What would be the ideal behavior for type generators:
- in request body, use `required` to determine, whether the field is required or not
- in response payload, use the `default`, even if the field is not mentioned in "required"

the switch `--properties-required-by-default` will either cause all to be required or all optional

PR #1315, opened 2025-10-10
fix: respect proxy headers

pilartomas:
I have tested both direct and through-ui scenarios, however, I have not tested a scenario with ingress.

To my knowledge, there is no standard ingress configuration for proxy header, it is specific to the ingress class.


apps/beeai-server/src/beeai_server/configuration.py
```diff
@@ -274,6 +274,7 @@ class Configuration(BaseSettings):
 
     platform_service_url: str = "beeai-platform-svc:8333"
     port: int = 8333
+    forwarded_allow_ips: str = ""
```
pilartomas: Is this a good default?

Issue #1158, opened 2025-09-08
UI – Introduce “Share Agent” buttons


jenna-winkler:
## Description

Currently, users must manually copy the URL from the browser address bar to share an agent. Adding Share Agent buttons will make sharing faster and more intuitive.

## User Story

As a user, I want a quick way to share the agent I’m interacting with so that I can easily send the agent’s URL to others and encourage usage.

## Functional Specs

- Introduce Share Agent button
- Clicking the button copies the current agent’s URL to the clipboard
- Users receive a subtle confirmation that the URL has been copied.

## Design

Figma: https://www.figma.com/design/oaJEgk4ZcH3mI1juPGITsx/Agent-Testbed-dev-ready--external-?node-id=477-4896&t=dNTD4hh9L8iscxui-1

PR #1323, opened 2025-10-14
feat(ui): add agent share button

kapetr:
Refs #1158


apps/beeai-ui/src/components/Tooltip/Tooltip.tsx
kapetr, line 1: These updates aren’t needed in the end (toast notification is used instead of tooltip), but they might come in handy in a future scenario:
- Allow using Tooltip as a controlled component.
- isEnabled prop could be useful for disabling tooltip without having to remove it from the render tree in specific cases.

PR #1328, opened 2025-10-15
feat(ui): model not configured page 

### LINKED ISSUE #1248: UI - Error state when model is not configured
## Description

When a model has not been configured via `beeai model setup`, launching the UI (either through `beeai ui` or by navigating directly to http://localhost:8334/) displays a generic error screen that doesn't clearly indicate the root cause or next steps.

Of note: This applies strictly to the developer local experience (pre-deployment, test phase).

## Current Behavior

Users see the following error screen when no model is configured:

Image

## Expected Behavior

The error state should:
- Clearly indicate that no model is configured
- Provide actionable next steps (e.g., "Run `beeai model setup` in your terminal to configure a model")
- Match the overall design language of the application
- Be user-friendly for first-time users who may not be familiar with the setup process

## Design

Figma: https://www.figma.com/design/RqJ7bfUc5qtlWli6BVoeWZ/Agent-Testbed-Jam?node-id=1533-21783&t=JEXm1e39RGHbz3MQ-1

kapetr:
Closes #1248

### DIFF
--- a/apps/beeai-ui/src/modules/runs/components/NoModelSelectedErrorPage.module.scss
+++ b/apps/beeai-ui/src/modules/runs/components/NoModelSelectedErrorPage.module.scss
@@ -4,17 +4,20 @@
  */
 
 .root {
-  display: grid;
-  grid-template-rows: 1fr max-content;
-  justify-content: center;
   inline-size: 100%;
   block-size: 100%;
   padding-inline: $spacing-08;
   overflow-y: auto;
 }
 
+.container {
+  display: grid;
+  grid-template-rows: 1fr max-content;
+  justify-content: center;
+  block-size: 100%;
+}
+
 .content {
-  max-inline-size: rem(800px);
   display: flex;
   flex-direction: column;
   align-items: center;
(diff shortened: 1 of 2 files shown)


apps/beeai-ui/src/modules/runs/components/NoModelSelectedErrorPage.module.scss
```diff
@@ -0,0 +1,54 @@
...
+  grid-template-rows: 1fr max-content;
+  justify-content: center;
+  inline-size: 100%;
+  block-size: 100%;
+  padding-inline: $spacing-08;
+  overflow-y: auto;
+}
+
+.content {
+  max-inline-size: rem(800px);
```
PetrBulanek: I think we should rather use generic Container component with `md` size. Or is there a specific reason why not?

kapetr: I needed specific size to have it centered and the content doesn't break lines. But it turns out in the end it fits exactly into md container and we even have 5px to spare 😄

PR #1346, opened 2025-10-16
docs: observability

jenna-winkler:


jenna-winkler:
langfuse integration needs to be tested before merging this PR

PR #1350, opened 2025-10-19
Harden provider build sandbox

### LINKED ISSUE #481: Secure agent containers
- enforce non-root users for all containers
- use gvisor or explore sandboxing solutions
- explore docker build sandboxing options
  - build is used only for 3rd party github-based providers, we can probably just issue a warning and postpone
  - maybe docker build is secure enough by default
  - solution options are docker in docker builders - e.g. kaniko, docker buildx with container driver

mohin-io:
Fixes #481

## Summary
- add a sandbox configuration to provider builds to centralize pod user, group, and runtime defaults
- harden the provider build job template by running all containers as non-root with dropped capabilities and an optional runtimeClass
- plumb the sandbox values into the Kubernetes template renderer so downstream overrides stay configuration-driven

## Design Decisions
- modeled sandbox knobs in Pydantic configuration to mirror existing provider_build settings and keep overrides declarative
- enforced runtime policies directly in the job template to minimize drift and guarantee every container uses the same guard rails
- exposed an optional runtime class hook (e.g., gVisor) without changing the default for clusters that have not provisioned it yet

## Alignment with Project Patterns
- reused the existing configuration object graph and template rendering flow instead of introducing ad-hoc environment variables
- followed the established YAML templating style for conditional blocks and Carbon-style security contexts to match other manifests
- kept capability drops and resource definitions consistent with prior containers to preserve least-privilege expectations

## Tests
- uv run ruff check --quiet
- uv run ruff format --quiet --check
- uv run pytest -m unit *(fails: kr8s client cannot authenticate without a Kubernetes test environment)*

### DIFF
--- a/agents/form/src/form/agent.py
+++ b/agents/form/src/form/agent.py
@@ -14,10 +14,7 @@ from a2a.types import (
     Message,
 )
 
-from a2a.utils.message import get_message_text
-
 from beeai_sdk.server import Server
-from beeai_sdk.server.context import RunContext
 
 import beeai_sdk.a2a.extensions
 from beeai_sdk.a2a.extensions.ui.form import (
--- a/agents/rag/src/rag/helpers/vectore_store.py
+++ b/agents/rag/src/rag/helpers/vectore_store.py
@@ -67,7 +67,6 @@ async def chunk_and_embed(embedding_function: EmbeddingFunction, file: File, vec
     """
 
     vector_store = await VectorStore.get(vector_store_id)
-    model_id = vector_store.model_id
 
     async with file.load_text_content() as loaded_file:
         text = loaded_file.text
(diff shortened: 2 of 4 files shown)


apps/beeai-server/src/beeai_server/infrastructure/kubernetes/default_templates/build-provider-job.yaml
```diff
@@ -68,7 +80,13 @@ spec:
             - name: image-tar
               mountPath: /tmp
           securityContext:
+            allowPrivilegeEscalation: false
```
jezekra1, line 83: Both builder implementations require priviledges, `kaniko` must be run as root and buildx has several options, but none of them allow these security constraints.

PR #1358, opened 2025-10-21
chore: rework beeai sdk

tomkis:


apps/beeai-sdk-ts/src/client/a2a/extensions/handle-agent-card.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (settingsDemands) {
+      fullfilledMetadata = fullfillSettingsDemand(fullfilledMetadata, await fullfillments.settings(settingsDemands));
+    }
+
+    if (secretDemands) {
+      fullfilledMetadata = fullfillSecretDemand(fullfilledMetadata, await fullfillments.secrets(secretDemands));
+    }
+
+    if (formDemands) {
+      const formFullfilment = await fullfillments.form(formDemands);
```
tomkis: Really suboptimal. Form acts like demand/fullfilment but in fact it's not because demand is always `Record<string,Demand>` which is not the case for Forms, we should unify this.

apps/beeai-ui/src/modules/variables/api/mutations/useUpdateVariables.ts
```diff
@@ -17,7 +17,7 @@ export function useUpdateVariables({ onSuccess }: Props = {}) {
     mutationFn: updateVariables,
     onSuccess,
     meta: {
-      invalidates: [variableKeys.lists()],
+      invalidates: [variableKeys.lists(), variableKeys.all()],
```
PetrBulanek: Is this intentional? `.all()` also invalidates `.lists()`, and I think `.lists()` alone is sufficient. :-)

apps/beeai-ui/src/modules/runs/settings/utils.ts
```diff
@@ -3,10 +3,10 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { AgentSettings, SettingsCheckboxFieldValue, SettingsRender } from 'beeai-sdk';
+import type { AgentSettings, SettingsCheckboxFieldValue, SettingsDemands } from 'beeai-sdk';
 import { match } from 'ts-pattern';
 
-export function getSettingsRenderDefaultValues(settingsRender: SettingsRender) {
+export function getSettingsRenderDefaultValues(settingsRender: SettingsDemands) {
   const defaults = settingsRender?.fields.reduce<AgentSettings>((valuesAcc, field) => {
```
PetrBulanek: Wouldn't it make sense to rename that function and parameter as well?

And the optional chaining is not needed here. ;-)

apps/beeai-sdk-ts/src/client/a2a/extensions/handle-agent-card.ts
```diff
@@ -0,0 +1,119 @@
...
+import type { LLMDemands, LLMFulfillments } from './services/llm';
+import { llmExtension } from './services/llm';
+import type { MCPDemands, MCPFulfillments } from './services/mcp';
+import { mcpExtension } from './services/mcp';
+import type { OAuthDemands, OAuthFulfillments } from './services/oauth-provider';
+import { oauthProviderExtension } from './services/oauth-provider';
+import { platformApiExtension } from './services/platform';
+import type { SecretDemands, SecretFulfillments } from './services/secrets';
+import { secretsExtension } from './services/secrets';
+import type { FormDemands, FormFullfillments } from './ui/form';
```
PetrBulanek: Throughout the entire repository, there are variants of the word "fulfillment" with incorrect spelling (there is "fullfillment" as well as "fullfilment"). Try to search `fullfil`. I would correct this globally.

apps/beeai-sdk-ts/src/client/a2a/extensions/ui/form.ts
PetrBulanek, line 1: Don't we want to unify the names of the variables of those schemas as well (renderSchema => formDemandsSchema etc.)?

apps/beeai-sdk-ts/src/client/a2a/extensions/ui/settings.ts
```diff
@@ -70,13 +70,13 @@ export type SettingsCheckboxGroupFieldValue = z.infer<typeof checkboxGroupFieldV
 export type SettingsSingleSelectFieldValue = z.infer<typeof singleSelectFieldValue>;
 export type SettingsFieldValue = z.infer<typeof settingsFieldValue>;
 
-export type SettingsRender = z.infer<typeof settingsRenderSchema>;
-export type AgentRunSettings = z.infer<typeof agentRunSettingsSchema>;
+export type SettingsDemands = z.infer<typeof settingsRenderSchema>;
+export type SettingsFullfillments = z.infer<typeof agentRunSettingsSchema>;
```
PetrBulanek: Same as above.

apps/beeai-ui/src/api/a2a/utils.ts
```diff
@@ -163,7 +163,7 @@ export function createTextPart(text: string): UITextPart {
   return textPart;
 }
 
-export function createFormPart(form: FormRender): UIFormPart | null {
+export function createFormPart(form: FormDemands): UIFormPart | null {
```
PetrBulanek: The types are wrong, the function never returns null.

Then we can also remove the `if (formPart)` check where that function is used.

```
const formPart = createFormPart(form);

if (formPart) {
  parts.push(formPart);
}

File: apps/beeai-ui/src/modules/runs/components/RunInput.tsx [hunk 7]
```diff
@@ -113,7 +113,7 @@ export function RunInput({ promptExamples, onMessageSent }: Props) {
 
             onMessageSent?.();
 
-            await chat(input);
+            await chat(input, {});
```
PetrBulanek: I think we can make the second argument optional and add a default value of `{}`.

apps/beeai-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
```diff
@@ -35,6 +40,16 @@ export function AgentDemandsProvider<UIGenericPart>({
 
   const { mutateAsync: createContextToken } = useCreateContextToken();
 
+  useEffect(() => {
+    if (agentClient?.demands.settingsDemands) {
+      setSelectedSettings(getSettingsRenderDefaultValues(agentClient.demands.settingsDemands));
+    }
+  }, [agentClient?.demands.settingsDemands]);
```
PetrBulanek: Can those `settingsDemands` on agentClient change during rerendering? If not, I would get rid of that useEffect and set it as the default value right in the `useState` call at the top.

apps/beeai-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
```diff
@@ -153,24 +168,39 @@ export function AgentDemandsProvider<UIGenericPart>({
...
+      return buildFullfilments({
+        contextToken,
+        selectedLLMProviders,
+        selectedEmbeddingProviders,
+        selectedMCPServers,
+        providedSecrets,
+        featureFlags,
+        selectedSettings,
+        formFullfillments: fulfillmentsContext.formFullfillments ?? null,
+        oauthRedirectUri: fulfillmentsContext.oauthRedirectUri ?? null,
```
PetrBulanek: Don't we want to accept undefined as well?

apps/beeai-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
```diff
@@ -184,6 +214,9 @@ export function AgentDemandsProvider<UIGenericPart>({
         selectEmbeddingProvider,
         selectMCPServer,
         selectedMCPServers,
+        selectedSettings,
+        settingsDemands: agentClient?.demands.settingsDemands ?? null,
```
PetrBulanek, line 214: Same as above.

apps/beeai-ui/src/modules/variables/api/mutations/useUpdateVariables.ts
tomkis: We no longer need this actually. Will remove.

apps/beeai-sdk-ts/src/client/a2a/extensions/ui/form.ts
tomkis: I did not change this here on purpose as it's still kinda unclear what is UI and what is dependency extension. We should refactor first.

apps/beeai-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
tomkis: I'd prefer to keep the type as null to explicitely mention that these are not supported.

apps/beeai-ui/src/api/a2a/part-processors.ts
```diff
@@ -38,11 +38,7 @@ export function processMessageMetadata(message: Message): UIMessagePart[] {
     parts.push(...sourceParts);
   }
   if (form) {
-    const formPart = createFormPart(form);
-
-    if (formPart) {
-      parts.push(formPart);
-    }
+    return [createFormPart(form)];
```
PetrBulanek: We should change this to:
```
parts.push(createFormPart(form));

File: apps/beeai-ui/src/modules/compose/contexts/ComposeProvider.tsx [hunk 12]
```diff
@@ -41,10 +41,13 @@ export function ComposeProvider({ children }: PropsWithChildren) {
...
   const { agentClient } = useBuildA2AClient({
     providerId: sequentialAgent?.provider.id,
-    extensions: getAgentExtensions(sequentialAgent),
     onStatusUpdate: handleTaskStatusUpdate,
   });
 
+  if (!agentClient) {
+    return <></>;
+  }
+
```
PetrBulanek, line 50: We should add some skeleton component or loader.

PR #1359, opened 2025-10-21
feat(ui): add single select component to form extension

PetrBulanek:
Closes: https://github.com/i-am-bee/beeai-platform/issues/1338


PetrBulanek:
@JanPokorny Can you please take a look at this too, see if it makes sense, and can we merge it? :-)

PR #1360, opened 2025-10-21
feat(server): increase default auto-stop timeout to 20 mins

matoushavlena:


apps/beeai-server/src/beeai_server/domain/constants.py
```diff
@@ -15,7 +15,7 @@ class _Undefined(Enum):
 undefined = _Undefined.undefined
 Undefined: TypeAlias = Literal[_Undefined.undefined]  # noqa: UP040
 
-DEFAULT_AUTO_STOP_TIMEOUT: Final[timedelta] = timedelta(minutes=5)
+DEFAULT_AUTO_STOP_TIMEOUT: Final[timedelta] = timedelta(minutes=20)
```
jezekra1, line 18: Please update `CreateProviderRequest` and `PatchProviderRequest` to include this number in the allowed range (currently maximum is 10 minutes)

PR #1363, opened 2025-10-21
fix(server): dynamic openapi server url

KaiWedekind:
### 🧪 What this PR does

This PR fixes the configuration of the OpenAPI server URL for the server component of the beeai platform.
It ensures that the OpenAPI specification exposes the correct servers URL when deployed.

### 📌 Why this change is needed

Previously, the servers entry in the OpenAPI spec resolved incorrectly in deployed environments, causing the OpenAPI documentation to point to a localhost base URL instead of the actual deployment endpoint.

### ✅ What has been changed

- The OpenAPI schema /api/v1/openapi.json now dynamically sets the servers field at runtime.
- The Swagger UI (/api/v1/docs) is updated to load from the dynamic OpenAPI schema.
- No functional API changes; only documentation and schema exposure are affected.

### 💡 Summary

This change ensures that the deployed OpenAPI documentation reflects the accurate API base URL, improving developer experience and alignment between the documented and actual API endpoints.

### DIFF
--- a/apps/agentstack-server/src/agentstack_server/application.py
+++ b/apps/agentstack-server/src/agentstack_server/application.py
@@ -50,7 +50,7 @@ logger = logging.getLogger(__name__)
 
 def get_version():
     try:
-        __version__ = version("beeai-server")
+        __version__ = version("agentstack-server")
     except PackageNotFoundError:
         __version__ = "0.1.0"
 


apps/beeai-server/src/beeai_server/application.py
```diff
@@ -169,7 +170,7 @@ async def lifespan(_app: FastAPI, procrastinate_app: procrastinate.App, mcp_serv
         default_response_class=ORJSONResponse,  # better performance then default + handle NaN floats
         docs_url="/api/v1/docs",
         openapi_url="/api/v1/openapi.json",
-        servers=[{"url": f"http://localhost:{configuration.port}"}],
+        servers=[{"url": os.getenv("HOST", f"http://localhost:{configuration.port}")}],
```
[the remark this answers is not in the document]
KaiWedekind: configuration.platform_service_url is set to "beeai-platform-svc:8333".

"beeai-platform-svc:8333" is an internal service address for Kubernetes.

This URL does not include a schema (http:// or https://), and is not valid for external OpenAPI docs — clients would fail if they tried to use it directly.

apps/agentstack-server/src/agentstack_server/application.py
```diff
@@ -109,6 +121,28 @@ def mount_routes(app: FastAPI):
...
+    @app.get("/api/v1/openapi.json", include_in_schema=False)
+    async def custom_openapi(request: Request):
+        openapi_schema = get_openapi(
+            title="BeeAI Platform server",
+            version=get_version(),
+            routes=app.routes,
+        )
+
+        base_url = str(request.base_url).rstrip("/")
+        openapi_schema["servers"] = [{"url": base_url}]
```
pilartomas: I believe `rstrip` is done for us already. Not needed here

apps/agentstack-server/src/agentstack_server/application.py
```diff
@@ -109,6 +121,28 @@ def mount_routes(app: FastAPI):
...
+
+        return JSONResponse(openapi_schema)
+
+    @app.get("/api/v1/docs", include_in_schema=False)
+    async def custom_docs(request: Request):
+        base_url = str(request.base_url).rstrip("/")
+        openapi_url = f"{base_url}/api/v1/openapi.json"
```
pilartomas: Here, you should use

```python
openapi_url = request.url_for(custom_openapi.__name__)
```

You'll need to update`Request` type import to do this, use the `fastapi` one. Should be compatible with the current usage in this module.

```python
from fastapi import Request
```

apps/agentstack-server/src/agentstack_server/application.py
```diff
@@ -111,6 +122,28 @@ def mount_routes(app: FastAPI):
     app.include_router(server_router, prefix="/api/v1", tags=["provider"])
     app.include_router(well_known_router, prefix="/.well-known", tags=["well-known"])
 
+    @app.get("/api/v1/openapi.json", include_in_schema=False)
+    async def custom_openapi(request: Request):
+        openapi_schema = get_openapi(
+            title="BeeAI Platform server",
```
pilartomas:
```suggestion
            title="Agentstack server",
```

apps/agentstack-server/src/agentstack_server/application.py
```diff
@@ -46,6 +48,15 @@
 logger = logging.getLogger(__name__)
 
 
+def get_version():
+    try:
+        __version__ = version("beeai-server")
```
pilartomas:
```suggestion
        __version__ = version("agentstack-server")
```

pilartomas:
LGTM, thanks 🥇

Issue #961, opened 2025-07-23
Trajectory UI/UX enhancements


jenna-winkler:
## Description

We need to improve how trajectory is displayed and interacted with in the UI. This issue covers making trajectory streaming, better timing for when controls appear, more flexible content display, and preventing it from taking over too much screen space.

## Current State

- "How did I get this answer?" button appears immediately after sending message to agent and is collapsed 
- Uses fixed thought/action structure
- Trajectory can potentially overtake the main content area when expanded

## Proposed Enhancements

- **Show trajectory while streaming** - Trajectory should be visible by default while waiting for the agent to respond. This gives users real-time insight into what the agent is thinking. 
   - TBD: Whether trajectory appears in chat (full screen), condensed to a box, or in a side panel.
- **Hide "How did I get this answer?" until complete** - The toggle button should only appear after the final answer is generated, and trajectory should automatically collapse at that point.
- **Remove rigid thought/action structure** - Replace the current fixed structure with flexible string content that can handle different types of trajectory data (reasoning, tool calls, observations, etc.)
- **Contain trajectory display** - Prevent trajectory from taking over the screen - likely through a side panel or other contained approach.

jenna-winkler:
I believe this has been implemented so closing - thank you @Nutcracker2012 & @kapetr !! (please reopen if I'm missing something and there was a reason we were keeping this open)

PR #1373, opened 2025-10-22
feat(ui): new trajectory ux

kapetr:
Refs #961 

Currently, trajectories don’t have a readable description field that matches the design, and we can’t make assumptions about their content format since each agent can implement it differently. That’s why the code block for printing trajectory content is kept for now.
Screenshot 2025-10-27 at 12 34 37

### DIFF
--- a/apps/beeai-ui/src/modules/trajectories/components/TrajectoryList.module.scss
+++ b/apps/beeai-ui/src/modules/trajectories/components/TrajectoryList.module.scss
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-$list-gap: 28px;
+$list-gap: #{rem(28px)};
 
 .root {
   position: relative;
@@ -29,7 +29,7 @@ $list-gap: 28px;
   padding-block-end: $spacing-05;
   display: flex;
   flex-direction: column;
-  row-gap: rem($list-gap);
+  row-gap: $list-gap;
   position: relative;
 
   .autoScroll & {
@@ -38,6 +38,6 @@ $list-gap: 28px;
   }
 
   .scrollGuard {
-    margin-block-start: rem(-$list-gap);
+    margin-block-start: -$list-gap;
   }
 }


apps/beeai-ui/src/modules/runs/chat/ChatAgentMessage.tsx
```diff
@@ -50,11 +50,15 @@ function Message({ message, onShow }: Props) {
...
     <div {...props} className={classes.root}>
-      {isPending ? (
+      {isPending && (
         <div className={classes.spinner}>
           <Spinner center />
         </div>
-      ) : (
+      )}
+
+      <MessageTrajectories message={message} autoScroll={isPending} toggleable={!isPending || hasContent} />
```
PetrBulanek: Isn't this `toggleable={!isPending || hasContent}` actually the same as `toggleable={!isPending}`? I'm not an expert on Boolean logic, but it seems to me that `|| hasContent` is already included there. :-)

apps/beeai-ui/src/modules/trajectories/components/TrajectoryItem.tsx
```diff
@@ -3,86 +3,50 @@
...
+    <motion.div {...fadeProps()} className={clsx(classes.root)}>
+      {title && <h3 className={classes.name}>{title}</h3>}
+
+      <div className={classes.body}>
+        {match(parsed)
+          .with({ type: 'string' }, ({ value }) => <LineClampText lines={5}>{value}</LineClampText>)
+          .otherwise(({ value }) => {
+            // if (hasDescription) {
+            //   return <LineClampText lines={5}>{parsed.input.thought}</LineClampText>;
+            // }
```
PetrBulanek: Probably some forgotten unused code that could be deleted.

apps/beeai-ui/src/modules/trajectories/components/TrajectoryList.module.scss
```diff
@@ -0,0 +1,41 @@
...
+  row-gap: rem(28px);
+  position: relative;
+
+  .autoScroll & {
+    max-block-size: rem(240px);
+    overflow-y: auto;
+  }
+
+  .scrollGuard {
+    margin-block-start: rem(-28px);
```
PetrBulanek: Is this the same as the row-gap value on .list? I'd rather use a sass variable for this.

apps/beeai-ui/src/modules/runs/chat/ChatAgentMessage.tsx
kapetr: Yes it is :)

PetrBulanek (approved):

PR #1395, opened 2025-10-24
chore: update a2a sdk

jezekra1:


agents/chat/src/chat/agent.py
```diff
@@ -296,6 +297,8 @@ async def chat(
     new_messages = [to_framework_message(item, extracted_files) for item in history]
 
     async for event, meta in agent.run(new_messages):
+        if isinstance(event, RequirementAgentFinalAnswerEvent) and use_streaming:
```
jezekra1, line 300: i accidentally removed this

PR #1414, opened 2025-10-29
chore: rename web and ui

PetrBulanek:


apps/agentstack-ui/src/utils/constants.ts
```diff
@@ -30,3 +28,9 @@ export const LF_PROJECTS_LINK = 'https://lfprojects.org/';
 export const TRUST_PROXY_HEADERS = (process.env.TRUST_PROXY_HEADERS ?? 'false').toLowerCase() === 'true';
 
 export const NEXTAUTH_URL = process.env.NEXTAUTH_URL ? new URL(process.env.NEXTAUTH_URL) : undefined;
+
+export const THEME_STORAGE_KEY = '@i-am-bee/agentstack/THEME';
+
+export const AGENT_SECRETS_SETTINGS_STORAGE_KEY = '@i-am-bee/agentstack/AGENT-SECRETS-SETTINGS';
```
kapetr, line 34: I'd move this to a separate file and create a constant for the common part, something like:
```
const LOCALE_STORAGE_PATH = '@i-am-bee/agentstack/'
```
What do you think?

tomkis (approved):

apps/agentstack-ui/src/utils/constants.ts
PetrBulanek: Yeah, I actually wanted to do something similar. Maybe even use some enum or something. But then I didn't want to add too many changes unrelated to the renaming. In the next PR, I want to remove the components for sequential workflow, so I would do it there, if you're okay with that. :-)

PR #1421, opened 2025-10-29
feat(server): connector injection

pilartomas:


apps/agentstack-server/src/agentstack_server/service_layer/services/a2a.py
```diff
@@ -169,6 +193,27 @@ async def _check_and_record_request(
...
+            mcp_ext = metadata.get(MCPServiceExtensionURI)
+            if mcp_ext:
+                with suppress(ValidationError):
+                    ext_meta = MCPServiceExtensionMetadata.model_validate(mcp_ext)
+                    for fulfillment in ext_meta.mcp_fulfillments.values():
+                        if fulfillment.transport.type == "streamable_http":
+                            connector = self._injectable_connectors.get(str(fulfillment.transport.url))
+
+                            if connector:
+                                assert connector.auth and connector.auth.token
```
jezekra1, line 208: is it possible that the connector auth token is missing here? If so let's use a better error message

jezekra1 (approved):

apps/agentstack-server/src/agentstack_server/service_layer/services/a2a.py
pilartomas: The actual check is in the constructor, such connector wouldn't be an injectable connector and therefore it won't be eligible for substitution.

PR #1434, opened 2025-10-30
feat(ui): hide offline agents

PetrBulanek:
Closes: https://github.com/i-am-bee/agentstack/issues/1562

### DIFF
--- a/apps/agentstack-ui/src/modules/agents/api/queries/useListAgents.ts
+++ b/apps/agentstack-ui/src/modules/agents/api/queries/useListAgents.ts
@@ -27,7 +27,7 @@ export function useListAgents({ includeUnsupportedUi, includeOffline, orderBy }:
       }
 
       if (!includeOffline) {
-        items = items.filter(({ state }) => state !== 'offline');
+        items = items.filter(({ state }) => state !== 'offline' && state !== 'error');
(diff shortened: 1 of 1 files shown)


apps/agentstack-ui/src/modules/agents/api/queries/useListAgents.ts
```diff
@@ -25,9 +26,13 @@ export function useListAgents({ onlyUiSupported, orderBy }: Props = {}) {
         items = items.sort(sortProvidersByCreatedAt);
       }
 
+      if (!includeOffline) {
+        items = items.filter(({ state }) => state !== 'offline');
```
kapetr: Shouldn't this condition include also `'error'` state according to this [comment](https://github.com/i-am-bee/agentstack/issues/1562#issuecomment-3532888074)?

PetrBulanek: Yes, it should. Will update that briefly. Thanks! :-)

PetrBulanek: Updated. :-)

kapetr (approved):

PR #1452, opened 2025-11-03
feat(server): add connector presets

### LINKED ISSUE #1439: [Feature]: Global ENV-configured connectors
**Is your feature request related to a problem? Please describe.
Users who install `Agent Stack` locally must currently register OAuth apps per-user or rely on each frontend to ship connectors. This creates friction and duplication across frontends and users.

Provide a platform capability to register a set of global, platform-scoped connectors that can be configured via environment variables (similar to existing auth configuration). Goals and requirements:
- Platform ships with a curated set of out-of-the-box connectors with pre-registered OAuth apps owned by `Agent Stack`.
- Allow platform administrators to supply `client_id` and `client_secret` via environment variables (or equivalent platform config) so the connector becomes available out-of-the-box.
- Connectors are global/platform-scoped (visible to all users) rather than user-scoped.
- The platform is responsible for advertising these connectors in the connector catalog so UIs (Agent Stack UI, etc.) do not need to implement custom logic to add them.
- Maintain existing user-scoped connectors behavior in parallel.

- Make each frontend pre-register or inject connectors into the UI — duplicates effort across every client.
- Require every user to register their own OAuth app — poor UX and high friction.
- Rely solely on `DCR` where available — works when supported but not universal across deployments.

**Additional context**
Motivation: reduce friction for local installs and frontends, centralize connector provisioning and ownership with the platform.

🤖 Generated with [GitHub Issue Creator](https://github.com/i-am-bee/github-issue-creator)

pilartomas:
Signed-off-by: Tomas Pilar <thomas7pilar@gmail.com>

closes #1439


pilartomas:
@matoushavlena what remains is to add a catalog of presents into `values.yaml`

PR #1481, opened 2025-11-06
chore: form extension rework

### LINKED ISSUE #1472: Form Extension redesign
The current approach to form extensions uses a single extension to handle both initialization and runtime input, which leads to significant confusion. Specifically, the runtime form can override the initial form, causing unexpected behavior. This ambiguity makes it difficult for agent builders to reason about and predict form behavior, increasing the chance of configuration errors and misinterpretation.

Refactor the concept of form extensions by splitting them into two clearly defined types:
1. **Request Runtime Form** – Handled as A2A `input-required`
2. **Request Initial Form** – Leverages dependency injection mechanism for the initial setup/configuration phase.

Each type should have its own dedicated extension URL to prevent interference and to provide a clearer contract for agent builders.

Ensure that the documentation is updated accordingly to explain the distinction, usage patterns, and any migration considerations.

tomkis:
fixes #1472

### DIFF
--- a/agents/form/src/form/agent.py
+++ b/agents/form/src/form/agent.py
@@ -85,7 +85,7 @@ class FormData(BaseModel):
     interests: list[str] | None
 
 @server.agent(
-    name="Single-turn Form Agent 2",
+    name="Single-turn Form Agent",
     documentation_url=f"https://github.com/i-am-bee/agentstack/blob/{os.getenv('RELEASE_VERSION', 'main')}/agents/form",
     version="1.0.0",
     default_input_modes=["text", "text/plain"],
--- a/apps/agentstack-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
+++ b/apps/agentstack-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
@@ -110,12 +110,7 @@ export function AgentDemandsProvider<UIGenericPart>({
   );
 
   const provideFormValues = useCallback((values: RunFormValues) => {
-    formFulfillmentsRef.current = {
-      form_fulfillments: {
-        ...formFulfillmentsRef.current.form_fulfillments,
-        initial_form: { values },
-      },
-    };
+    formFulfillmentsRef.current = { form_fulfillments: { initial_form: { values } } };
   }, []);
 
   const [selectedMCPServers, setSelectedMCPServers] = useState<Record<string, string>>({});
--- a/apps/agentstack-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
+++ b/apps/agentstack-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
@@ -89,7 +89,7 @@ function AgentRunProvider({ agent, agentClient, children }: PropsWithChildren<Ag
   const pendingRun = useRef<ChatRun>(undefined);
 
   const { contextId, getContextId, updateContextWithAgentMetadata } = usePlatformContext();
-  const { getFulfillments, formDemands } = useAgentDemands();
+  const { getFulfillments, provideFormValues, formDemands } = useAgentDemands();
   const { files, clearFiles } = useFileUpload();
 
   const updateCurrentAgentMessage = useCallback(
@@ -319,6 +319,8 @@ function AgentRunProvider({ agent, agentClient, children }: PropsWithChildren<Ag
     (form: UIMessageForm) => {
       checkPendingRun();
 
+      provideFormValues(form.response);
+
       const message: UIUserMessage = {
         id: uuid(),
         role: Role.User,
@@ -328,7 +330,7 @@ function AgentRunProvider({ agent, agentClient, children }: PropsWithChildren<Ag
 
       return run(message, {});
     },
-    [checkPendingRun, run],
+    [checkPendingRun, provideFormValues, run],
   );
 
   const { startAuth } = useStartOAuth({
--- a/agents/form/uv.lock
+++ b/agents/form/uv.lock
@@ -63,7 +63,7 @@ requires-dist = [
 
 [package.metadata.requires-dev]
 dev = [
-    { name = "beeai-framework", extras = ["duckduckgo", "wikipedia"], specifier = ">=0.1.58" },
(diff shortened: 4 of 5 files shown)


apps/agentstack-cli/src/agentstack_cli/commands/agent.py
```diff
@@ -778,9 +783,9 @@ async def run_agent(
 
         initial_form_render = next(
```
tomkis, line 809: We should probably refactor this and use the extension client and parse as service extension cc @jezekra1

tomkis:
@jenna-winkler this is a breaking change for the 0.4.X users, we'll bump up the version to 0.5.0. 

We should provide migration guide that can be prepared  from the diff in documentation in this PR.

apps/agentstack-sdk-ts/src/client/a2a/extensions/common/form.ts
```diff
@@ -113,17 +109,15 @@ const fieldSchema = z.discriminatedUnion('type', [
...
   title: z.string().nullish(),
   description: z.string().nullish(),
   columns: z.int().min(1).max(4).nullish(),
   submit_label: z.string().nullish(),
   fields: z.array(fieldSchema).nonempty(),
 });
 
-const responseSchema = z.object({
-  id: z.string().nonempty(),
+export const responseSchema = z.object({
```
PetrBulanek: I'd rename it to formResponseSchema to align with formRenderSchema.

apps/agentstack-sdk-ts/src/client/a2a/extensions/handle-input-required.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type { FormResponseValue } from './common/form';
+import { requestFormExtension } from './ui/request-form';
+
+export type RunFormValues = Record<string, FormResponseValue>;
```
PetrBulanek: Do we need this?

apps/agentstack-sdk-ts/src/client/a2a/extensions/services/form.ts
```diff
@@ -0,0 +1,31 @@
...
+    .partial(),
+});
+export type FormDemands = z.infer<typeof formDemandSchema>;
+
+const formFulfillmentSchema = z.object({
+  form_fulfillments: z.record(z.string(), responseSchema),
+});
+export type FormFulfillments = z.infer<typeof formFulfillmentSchema>;
+
+export const formExtension: A2AServiceExtension<typeof URI, z.infer<typeof formDemandSchema>, FormFulfillments> = {
```
PetrBulanek: You can use `FormDemands` instead of `z.infer<typeof formDemandSchema>`.

apps/agentstack-sdk-ts/src/index.ts
```diff
@@ -18,7 +21,6 @@ export * from './client/a2a/extensions/services/secrets';
 export * from './client/a2a/extensions/types';
 export * from './client/a2a/extensions/ui/agent-detail';
 export * from './client/a2a/extensions/ui/citation';
-export * from './client/a2a/extensions/ui/form';
```
PetrBulanek, line 21: We should also export `ui/request-form` here.

apps/agentstack-sdk-ts/src/client/a2a/extensions/handle-agent-card.ts
PetrBulanek, line 1: I think the return value for the `resolveMetadata` function should be typed. Similarly to `handleInputRequired`, where it already is.

`const metadata: Record<string, unknown> = {};`

apps/agentstack-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
```diff
@@ -200,6 +200,9 @@ function AgentRunProvider({ agent, agentClient, children }: PropsWithChildren<Ag
           message,
           contextId,
           fulfillments,
+          responses: {
+            form: message.form?.response ?? undefined,
```
PetrBulanek: The nullish coalescing is unnecessary here.

apps/agentstack-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
```diff
@@ -106,6 +109,15 @@ export function AgentDemandsProvider<UIGenericPart>({
...
 
+  const provideFormValues = useCallback((values: RunFormValues) => {
+    formFulfillmentsRef.current = {
+      form_fulfillments: {
+        ...formFulfillmentsRef.current.form_fulfillments,
+        initial_form: { values },
+      },
+    };
+  }, []);
+
```
PetrBulanek, line 115: Is it necessary to use ref? Wouldn't it be better if the `provideFormValues` function just set only values as state?

Besides, this spread `...formFulfillmentsRef.current.form_fulfillments` seems unnecessary to me, because there will never be any other key than `initial_form`, so it will always be overwritten. Or am I missing something?

apps/agentstack-sdk-py/src/agentstack_sdk/a2a/extensions/ui/form.py
```diff
@@ -91,7 +81,6 @@ class CheckboxField(BaseField):
 
 
 class FormRender(BaseModel):
-    id: str
```
PetrBulanek, line 94: The `id` is still used in `agents/form/src/form/agent.py`, we should remove it.

PetrBulanek (changes requested):

agents/form/src/form/agent.py
```diff
@@ -65,18 +68,24 @@
...
+    location: str | None
+    date_from: str | None
+    date_to: str | None
+    notes: list[FileInfo] | None
+    flexible: bool | None
+    interests: list[str] | None
 
 @server.agent(
-    name="Single-turn Form Agent",
+    name="Single-turn Form Agent 2",
```
kapetr, line 88: Debug leftover?

kapetr (approved):

agents/form/src/form/agent.py
kapetr: Already fixed

PR #1512, opened 2025-11-10
feat(ui): update homepage and sidebar

PetrBulanek:


apps/agentstack-ui/src/api/schema.d.ts
```diff
@@ -1687,6 +1687,8 @@ export interface components {
       metadata?: {
         [key: string]: string;
       } | null;
+      /** Provider Id */
+      provider_id?: string | null;
```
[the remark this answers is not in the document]
PetrBulanek: This is a generated file.

apps/agentstack-ui/src/components/Sidebar/AgentNav.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { routes } from '#utils/router.ts';
+
+import { NavGroup } from './NavGroup';
+import { NavList } from './NavList';
+import NewSession from './NewSession.svg';
+
+interface Props {
+  providerId: string;
+}
+
```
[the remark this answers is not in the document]
PetrBulanek: Not relevant comment.

apps/agentstack-ui/src/components/Sidebar/RecentlyUsedAgentsNav.tsx
```diff
@@ -0,0 +1,35 @@
...
+      agents?.map(({ name, provider: { id } }) => ({
+        label: name,
+        href: routes.agentRun({ providerId: id }),
+      })),
+    [agents],
+  );
+
+  return (
+    <NavGroup heading="Recently used agents" className={className}>
+      <NavList items={items} isLoading={isLoading} skeletonCount={5} noItemsMessage="No history" />
```
kapetr, line 32: Is the `noItemsMessage` correct?

apps/agentstack-ui/src/modules/agents/components/cards/AgentAuthor.tsx
```diff
@@ -0,0 +1,31 @@
...
+// import { getNameInitials } from '#utils/helpers.ts';
+import classes from './AgentAuthor.module.scss';
+
+interface Props {
+  author: AgentDetailContributor;
+}
+
+export function AgentAuthor({ author }: Props) {
+  const { name } = author;
+  // const initials = getNameInitials(name);
```
kapetr: Are we about to use initials in some cases?

apps/agentstack-ui/src/modules/agents/components/cards/AgentCard.tsx
```diff
@@ -38,6 +54,26 @@ AgentCard.Skeleton = function AgentCardSkeleton() {
...
+      <div className={classes.footer}>
+        <AgentAuthor.Skeleton />
+
+        <SkeletonText className={classes.timeAgo} />
+      </div>
     </article>
   );
 };
+
+function getDistance(date: string) {
```
kapetr, line 67: Later we could probably parametrize and reuse `ElapsedTime` component from `ToastProvider`

kapetr (approved):

apps/agentstack-ui/src/components/Sidebar/RecentlyUsedAgentsNav.tsx
PetrBulanek: It is actually in the design.

image

apps/agentstack-ui/src/modules/agents/components/cards/AgentAuthor.tsx
PetrBulanek: It should be added later, but I only found out after I did it, so I just commented it out.

However, in the end, only the icon should be added, so I may remove the initials and leave only the styles.

image

apps/agentstack-ui/src/modules/agents/components/cards/AgentCard.tsx
PetrBulanek: I looked at it originally, but the usage seemed a bit different to me, and I thought it wasn't necessary to abstract it. Maybe later, if we need something similar elsewhere. :-)

Issue #1490, opened 2025-11-07
[Feature]: Automatic Agent Card refresh for unmanaged agents after deployment


matoushavlena:
**Is your feature request related to a problem? Please describe.**  
When an unmanaged agent is redeployed, the Platform **does not refresh the Agent Card**.  
The Platform continues showing the **old agent version**, even though the agent endpoint exposes the updated version.

Example:
- Agent card: **1.172.0**
- Platform providers list: **1.169.1**

This causes UI confusion and makes debugging difficult.

Problem:
- Platform performs a **one-time copy** of the Agent Card at registration.
- Cronjob updates work **only for managed agents**.
- Unmanaged agents stay stale indefinitely.
- Current workaround: **delete the provider**, let cronjob recreate it.

**Describe the solution you'd like**  
The Platform should **automatically refresh the Agent Card** for unmanaged agents after a new agent version is deployed.  

A periodic cronjob sync or another automated update mechanism should ensure the Provider record always reflects the current agent version.

**Describe alternatives you've considered**  
- Manually deleting and recreating the provider (works but is not sustainable).  
- Relying on existing cronjob logic (currently only updates managed agents).  
- Leaving it as a one-time copy (not viable since agents frequently change).

**Additional context**  
Short-term workaround is deletion of the provider, but long-term Platform changes are needed.

PR #1528, opened 2025-11-11
feat(server): refresh unmanaged agent cards

aleskalfas:
#1490

### DIFF
--- a/apps/agentstack-server/src/agentstack_server/jobs/crons/provider.py
+++ b/apps/agentstack-server/src/agentstack_server/jobs/crons/provider.py
@@ -20,7 +20,7 @@ from agentstack_server.jobs.queues import Queues
 from agentstack_server.service_layer.services.providers import ProviderService
 from agentstack_server.service_layer.services.users import UserService
 from agentstack_server.service_layer.unit_of_work import IUnitOfWorkFactory
-from agentstack_server.utils.a2a import detect_card_changes, get_extension
+from agentstack_server.utils.a2a import get_extension
 from agentstack_server.utils.utils import extract_messages
 
 logger = logging.getLogger(__name__)
@@ -158,7 +158,7 @@ async def refresh_unmanaged_provider_state(
             state = UnmanagedState.OFFLINE
         finally:
             # Unified update: patch both agent card (if changed) and state (if changed) in a single call
-            card_changes = detect_card_changes(provider.agent_card, resp_card) if resp_card else {}
+            card_changes = provider.agent_card != resp_card
             state_changed = state != provider.unmanaged_state
 
             if card_changes or state_changed:
--- a/apps/agentstack-server/src/agentstack_server/utils/a2a.py
+++ b/apps/agentstack-server/src/agentstack_server/utils/a2a.py
@@ -1,15 +1,7 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Any
-
 from a2a.types import AgentCard, AgentExtension
-from pydantic import BaseModel
-
-
-class FieldChange(BaseModel):
-    old: Any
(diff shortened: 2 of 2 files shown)


apps/agentstack-server/src/agentstack_server/jobs/crons/provider.py
```diff
@@ -147,6 +148,18 @@ async def _check_provider(provider: Provider):
...
+                # Update agent card if it changed (after state determination to avoid mutating provider during checks)
+                card_changes = detect_card_changes(provider.agent_card, resp_card)
+                if card_changes:
+                    logger.info(f"Provider {provider.id} agent card changed: {card_changes}")
+                    try:
+                        provider.agent_card = resp_card
+                        async with uow_f() as uow:
+                            await uow.providers.update(provider=provider)
+                            await uow.commit()
+                    except Exception as ex:
```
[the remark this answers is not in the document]
aleskalfas: Great, i've updated

apps/agentstack-server/src/agentstack_server/utils/a2a.py
```diff
@@ -10,3 +18,20 @@ def get_extension(agent_card: AgentCard, uri: str) -> AgentExtension | None:
         return next(ext for ext in extensions if ext.uri == uri)
     except StopIteration:
         return None
+
+
+def detect_card_changes(old_card: AgentCard, new_card: AgentCard) -> dict[str, FieldChange]:
+    """Detect specific changes between two agent cards."""
+    old_dict = old_card.model_dump(mode="json", exclude={"signatures"})
```
jezekra1: signatures was causing an issue here?
otherwise you can just do `old_card == new_card` check

apps/agentstack-server/src/agentstack_server/utils/a2a.py
```diff
@@ -10,3 +18,20 @@ def get_extension(agent_card: AgentCard, uri: str) -> AgentExtension | None:
...
+    new_dict = new_card.model_dump(mode="json", exclude={"signatures"})
+
+    changes = {}
+    all_keys = set(old_dict.keys()) | set(new_dict.keys())
+
+    for key in all_keys:
+        old_val = old_dict.get(key)
+        new_val = new_dict.get(key)
+        if old_val != new_val:
+            changes[key] = FieldChange(old=old_val, new=new_val)
```
jezekra1: the `FieldChange` is actually not used anywhere, let's not add it before we need it, it's better to have less code to maintain :)

jezekra1 (changes requested):

apps/agentstack-server/src/agentstack_server/utils/a2a.py
aleskalfas: Signatures change every time a card is signed, even if the content is identical. They're cryptographic proof of authenticity, not semantic content so we have to omit them to check data change.

apps/agentstack-server/src/agentstack_server/utils/a2a.py
aleskalfas: It is useful for debugging.

apps/agentstack-server/src/agentstack_server/utils/a2a.py
jezekra1: I don't think that should be the case, if signed with the same key the signature should be the same but I did not read the full RFC :D https://www.rfc-editor.org/rfc/rfc7515.html

Anyway we would need to change this in more places and none of our agents are sending signatures yet, so let's implement this separately when needed

apps/agentstack-server/src/agentstack_server/utils/a2a.py
aleskalfas: Because we don't need to omit signatures attributes I will compare the objects directly and this functionality will became obsolete so I'll remove it.

PR #1531, opened 2025-11-11
Issue 1483

tedhabeck:
- Adds support for agentstack-cli automatic token refresh.

Please slack me directly if you need a working PKCE configuration to test this setup.

Automatic token refresh example:

```bash
#( 11/11/25@12:02PM )( habeck@TEDs-MacBook-Pro ):~/beeai-networking/agentstack/apps/agentstack-server@issue-1483✗✗✗
   mise agentstack-cli:run -- agent list
[agentstack-cli:run] $ uv run agentstack agent list
warning: `VIRTUAL_ENV=/Users/habeck/beeai-networking/agentstack/apps/agentstack-server/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
SHORT ID  NAME                    STATE     DESCRIPTION        INTERACTION  LOCATION               MISSING ENV  LAST ERROR
3b906a27  Chat                    ready     Agent with memor…  multi-turn   agents/chat:0.4.1-rc2  <none>       <none>    
6b27a943  RAG                     missing   RAG agent that r…  multi-turn   agents/rag:0.4.1-rc2   <none>       <none>    
1af40f1e  Single-turn Form Agent  missing   Example demonstr…  multi-turn   agents/form:0.4.1-rc2  <none>       <none>    
Finished in 9.36s
#( 11/11/25@12:17PM )( habeck@TEDs-MacBook-Pro ):~/beeai-networking/agentstack/apps/agentstack-server@issue-1483✗✗✗
   
```

cli login example:
```bash
  mise agentstack-cli:run -- server login 
[agentstack-sdk-py:setup] sources up-to-date, skipping
[helm:build:dependencies] sources up-to-date, skipping
[agentstack-cli:setup] sources up-to-date, skipping
[helm:build] sources up-to-date, skipping
[agentstack-cli:build:copy-helm-chart] sources up-to-date, skipping
[agentstack-cli:run] $ uv run agentstack server login
warning: `VIRTUAL_ENV=/Users/habeck/beeai-networking/agentstack/apps/agentstack-server/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
? Select a server, or log in to a new one: Log in to a new server
? Enter server URL: http://agentstack-cli.localhost:8333
📝 INFO: No authentication tokens found for this server. Proceeding to log in.
? Select an identity provider: IBMiD-PKCE
📝 INFO: Opening browser for login: 
https://isg-verify1.verify.ibm.com/oauth2/authorize?client_id=14d88fab-d185-4100-bea0-581461ef4fc7&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A9001%2Fcallback&scope=openid+email+profile&code_challenge=n941S9cSZlaJWj7Iw2t5KIALMJQsG1pNl2bfH1Z22aA&code_challenge_method=S256
✅ SUCCESS: Logged in to http://agentstack-cli.localhost:8333.
Finished in 44.47s
```

### DIFF
--- a/apps/agentstack-cli/src/agentstack_cli/commands/server.py
+++ b/apps/agentstack-cli/src/agentstack_cli/commands/server.py
@@ -170,7 +170,6 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] = 
                     resp = None
                     try:
                         app_name = get_unique_app_name()
-                        print("app_name=", app_name)
                         resp = await client.post(
                             registration_endpoint,
                             json={
@@ -178,7 +177,6 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] = 
                                 "grant_types": ["authorization_code", "refresh_token"],
                                 "enforce_pkce": True,
                                 "all_users_entitled": True,
-                                "access_policy": 1,
                                 "redirect_uris": [REDIRECT_URI],
                             },
                         )
@@ -187,13 +185,17 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] = 
                         client_id = data["client_id"]
                         client_secret = data["client_secret"]
                         registration_token = data["registration_access_token"]
-                    except Exception:
+                    except Exception as e:
                         if resp:
-                            error_details = resp.json()
-                            console.warning(
-                                f"error: {error_details['error']} error description: {error_details['error_description']}"
-                            )
-                        console.warning("Dynamic client registration failed. Proceed with manual input.")
+                            try:
+                                error_details = resp.json()
+                                console.warning(
+                                    f"error: {error_details['error']} error description: {error_details['error_description']}"
+                                )
+
+                            except Exception:
+                                console.info("no parsable json response.")
+                        console.warning(f" Dynamic client registration failed. Proceed with manual input.  {e!s}")
 
             if not client_id:
                 client_id = await inquirer.text(  #  type: ignore
@@ -246,11 +248,15 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] = 
                     token_resp.raise_for_status()
                     token = token_resp.json()
                 except Exception as e:
-                    if token_resp:
-                        error_details = token_resp.json()
-                        console.warning(
-                            f"error: {error_details['error']} error description: {error_details['error_description']}"
-                        )
+                    if resp:
+                        try:
+                            error_details = resp.json()
+                            console.warning(
+                                f"error: {error_details['error']} error description: {error_details['error_description']}"
+                            )
+                        except Exception:
+                            console.info("no parsable json response.")
+
                     raise RuntimeError(f"Token request failed: {e}") from e
 
             if not token:


helm/values.yaml
```diff
@@ -63,10 +63,19 @@ auth:
...
+        "app": "cli", # one of "ui" or "cli". Controls where this identity provider is presented
+        "client_id": "<oidc_client_id>",
+        "client_secret": "", # empty string for public client
+        "issuer": "<oidc_issuer>",
+      },
       {
         "name": "w3id",
         "id": "w3id",
-        "class": "IBM",
+        "app": "ui", # one of "ui" or "cli". Controls where this identity provider is presented
```
pilartomas: The CLI in not part of the infrastructure, this will always be `ui` so there is no need to add new field. The UI application should stay the same.

I assume the reason for adding such provider was to ensure the requests from the CLI are not being rejected. CLI should send requests to the `agentstack-server` directly, not through the `agentstack-ui`.

## Sidenote, out of scope of this PR

It would be useful to separate client_id for `ui` and the `server`, those can and should be different in principle. However, that would be probably best achieved by revisiting helm templates.

apps/agentstack-cli/src/agentstack_cli/commands/server.py
```diff
@@ -158,34 +164,36 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] =
...
-                ).execute_async()
-                if not client_id:
-                    raise RuntimeError("Client ID is mandatory. Action cancelled.")
-                client_secret = (
-                    await inquirer.text(  #  type: ignore
-                        message="Enter Client Secret (optional):"
-                    ).execute_async()
-                    or None
-                )
+            # registration_endpoint = oidc["registration_endpoint"]
```
pilartomas: Dynamic Client Registration should be supported.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -79,6 +127,14 @@ def load_auth_token(self) -> str | None:
         if not auth_server or not auth_server.token:
             return None
 
+        if (auth_server.token.expires_at or 0) - 60 < time.time():
+            console.warning("auth token expired")
+            console.log("Exchanging refresh token for updated access token...")
```
pilartomas: I'd drop the logs, access token expiration + refresh occurs regularly, usually every few hours. It is best to keep it silent unless it fails.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -59,14 +63,58 @@ def save_auth_token(
...
+                    self._auth.servers[self._auth.active_server or ""].authorization_servers[auth_server].client_id
+                )
+                resp = await client.post(
+                    f"{token_endpoint}",
+                    data={
+                        "grant_type": "refresh_token",
+                        "refresh_token": token.refresh_token,
+                        "scope": token.scope,
+                        "client_id": client_id,
+                    },
```
pilartomas, line 111: We must also support `client_secret` for non-public oAuth clients.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -59,14 +63,58 @@ def save_auth_token(
...
+                raise RuntimeError(f"Failed to refresh token: {e}") from e
+            self.save_auth_token(
+                self._auth.active_server or "",
+                self._auth.active_auth_server or "",
+                self._auth.servers[self._auth.active_server or ""].authorization_servers[auth_server].client_id or "",
+                "",
+                token=new_token,
+            )
+            self._save()
+            console.log("exchange_refresh_token succeeded.")
```
pilartomas: ditto

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -59,14 +63,58 @@ def save_auth_token(
...
+            except Exception as e:
+                raise RuntimeError(f"Failed to refresh token: {e}") from e
+            self.save_auth_token(
+                self._auth.active_server or "",
+                self._auth.active_auth_server or "",
+                self._auth.servers[self._auth.active_server or ""].authorization_servers[auth_server].client_id or "",
+                "",
+                token=new_token,
+            )
+            self._save()
```
pilartomas: This should be removed, `save_auth_token` already does it internally.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -59,14 +63,58 @@ def save_auth_token(
         token: dict[str, Any] | None = None,
     ) -> None:
         if auth_server is not None and client_id is not None and token is not None:
+            if token["access_token"]:
+                usetimestamp = int(time.time()) + int(token["expires_in"]) - 60
```
pilartomas: Looks like the 60 seconds leeway is duplicated here, shouldn't it be present only once, in the expiration check logic?

apps/agentstack-cli/src/agentstack_cli/commands/server.py
```diff
@@ -158,34 +164,36 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] =
...
+            #             resp.raise_for_status()
+            #             data = resp.json()
+            #             client_id = data["client_id"]
+            #             client_secret = data["client_secret"]
+            #         except Exception as err:
+            #             console.warning("Dynamic client registration failed. Proceed with manual input.")
+            #             console.warning(f"details: {err!s}")
+
+            # if not client_id:
+            #     client_id = await inquirer.text(  #  type: ignore
```
pilartomas: Interactive manual client input should be supported.

apps/agentstack-cli/src/agentstack_cli/commands/server.py
```diff
@@ -135,18 +135,24 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] =
...
 
-        client_id = config.client_id
+        client_id = None
         client_secret = config.client_secret
 
         if auth_servers:
             if len(auth_servers) == 1:
-                auth_server = auth_servers[0]
+                auth_server = auth_servers[0].get("server", "")
+                client_id = auth_servers[0].get("client_id", "")
```
pilartomas: `auth_servers` is a list of strings as defined in https://datatracker.ietf.org/doc/html/rfc9728, it is not an object containing `server` and `client_id` fields.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
tedhabeck: We do not expose the client_secret on the client side.  This is why we use a PKCE application for the cli.

apps/agentstack-cli/src/agentstack_cli/commands/server.py
tedhabeck: Chicken and the egg problem.   You need an authorization token to perform dynamic client registation.   You *could* have a separate command e.g. `mise agentstack-cli:run -- server register-client` after login.  In that way a client could perform a dynamic registration using the authorization token provided that they were granted that permission with the identity provider.

Without a reference to the use case, I'm struggling to understand this requirement.

apps/agentstack-cli/src/agentstack_cli/commands/server.py
tedhabeck: OK, was not aware you were trying to follow a spec here (Clearly, it works fine as is though). I'll add a new metadata parameter called "client_data" to hold the dictionary.  The spec states, `"Additional protected resource metadata parameters MAY also be used."`

helm/values.yaml
tedhabeck: No, not really.  I'll try to explain it more completely below.

In this example (i.e. the `providers` array in values.yaml), the first provider in the array is what the cli targets as an identity provider as it is a PKCE (public) client.  The second provider where you've left a comment is indeed an example of an non PKCE provider used by the UI.  

The  "app" field allows filtering for the APIs which serve up the available identity providers to the UI & cli.  The idea being that when cli login is invoked, and the server's URL is entered, the set of identity providers it will send back will all be PKCE providers.  The opposite is true for the UI during sign in.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
pilartomas: The CLI, once installed, runs exclusively on user's machine and the credentials are no longer shipped with the application. User must supply client credentials manually on their machine (if DCR fails) during `agentstack server login` and the CLI never discloses them to anyone else other than the authorization server. Hence we can support both public and private clients.

PKCE (with S256 challenge method) ensures that when authorization code is hijacked by an attacker, it cannot be exchanged for access token. This was indeed primarily targeting public clients at the time but now it is mandated for all in oAuth 2.1 afaik.

apps/agentstack-cli/src/agentstack_cli/commands/server.py
pilartomas: The [RFC7591](https://www.rfc-editor.org/rfc/rfc7591) states that 

```
The client registration endpoint MAY be an OAuth 2.0 [[RFC6749](https://www.rfc-editor.org/rfc/rfc6749)]
   protected resource and it MAY accept an initial access token in the
   form of an OAuth 2.0 access token to limit registration to only
   previously authorized parties.
 ```
 
Making the authorization is optional.

Our primary goal is to mimic the MCP when is comes to oAuth bootstrap process. In there, the combination of DCR and two metadata RFCs enables applications to register as oAuth clients automatically, without user interaction) based on a simple URL of the protected resource (e.g. https://mcp.stripe.com)

Therefore, so far, we've been primarily focusing on authorization servers that do not protect their registration endpoint, falling back to manual registration otherwise as suggested in https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#dynamic-client-registration.

File: helm/values.yaml (hunk 1 above)
Comment (pilartomas) [reply helm/values.yaml]: I think there is an misconception in what PKCE does. It was introduced as a way to harden the authorization code flow against hijacking. This was problematic primarily for public clients which do not send client secret during token request. However, since oAuth 2.1, the PKCE MUST be used for any authorization code flow. One cannot say anymore that public clients use PKCE and private do not.

If the resource server (agentstack-server) really wants to restrict which client(s) may enter, it must validate the `azp` field. This is potentially a security feature we might want to explore for enterprise but the default should be to allow any client in order to promote DCR.

File: apps/agentstack-cli/src/agentstack_cli/commands/server.py (hunk 2 above)
Comment (araujof) [reply apps/agentstack-cli/src/agentstack_cli/commands/server.py]: According to RFC 7591:

- DCR endpoints MAY be OAuth 2.0 protected resources requiring initial access tokens
- However, "To support open registration and facilitate wider interoperability, the client registration endpoint SHOULD allow registration requests with no authorization (which is to say, with no initial access token in the request).  These requests MAY be rate-limited or otherwise limited to prevent a denial-of-service attack on the client registration endpoint"

I suggest we adopt the following stance on this:
1. Primary path: Support unprotected DCR for initial client registration (following MCP spec guidance)
2. Fallback path: When DCR fails or is unavailable, prompt users for manual client credentials
3. Security consideration: Implement rate limiting, IP reputation checks, and validation on the registration endpoint to prevent abuse

File: helm/values.yaml (hunk 1 above)
Comment (araujof) [reply helm/values.yaml]: I think that architectural concern is valid, but the implementation is trying to address an issue in the current approach (CLI vs UI). Perhaps, the resolution depends on deployment architecture:

If CLI → Server (direct):
- No app field needed
- CLI and UI can use the same provider configuration
- Server's API returns all applicable providers

If CLI → UI → Server (indirect):
- The app field makes sense as a filter
- However, this creates unnecessary complexity

Recommendation:

1. Remove the app field distinction
2. Ensure CLI communicates directly with the backend server
3. Let the OAuth authorization server's capabilities (PKCE support, client type) determine applicability, not an app filter. If truly needed, use a more semantic field like `client_type: ["public", "confidential"]` or `supports_pkce: true`.

File: apps/agentstack-server/src/agentstack_server/service_layer/services/auth.py [hunk 10]
```diff
@@ -19,5 +19,10 @@ def protected_resource_metadata(self, *, resource: str) -> dict:
         return {
             "resource": resource,
             "authorization_servers": [str(p.issuer) for p in self._config.auth.oidc.providers if p.issuer is not None],
+            "client_data": [
+                {"server": str(p.issuer), "client_id": p.client_id, "name": p.name}
+                for p in self._config.auth.oidc.providers
+                if p.issuer is not None
+            ],
```
pilartomas, line 26: I believe this can be also cleaned up.

By the way, I noticed [Client ID Metadata RFC Draft](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00) which tries to approach this problem from a different angle but also replaces DCR in a sense. It allows `client_id` to be an arbitrary URL, previously unknown to the authorization server. The authorization server is then responsible for grabbing the client information dynamically during the flow itself (but making no persistent client, I believe).

pilartomas (approved):
LGTM

PR #1602, opened 2025-11-19
fix(tests): use self-hosted runner for e2e tests

jezekra1:
## Summary
Github runners continuously run out of disk space or into some other limitation with our e2e pipeline

## Documentation
- [x] No Docs Needed:


jezekra1:
not going to work with the current setup

PR #1606, opened 2025-11-19
feat(agentstack-server): support stdio mcp

JanPokorny:
## Summary
Refs: https://github.com/i-am-bee/agentstack/issues/1604

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-server/src/agentstack_server/service_layer/services/managed_mcp_service.py
+++ b/apps/agentstack-server/src/agentstack_server/service_layer/services/managed_mcp_service.py
@@ -30,7 +30,7 @@ class ManagedMcpService:
         return (preset := self.find_preset(connector.url)) is not None and preset.url.scheme == "mcp+stdio"
 
     def get_service_url(self, *, connector: Connector) -> str:
-        return f"http://managed-mcp-{connector.id.hex[:16]}.{self._kubectl._default_kwargs['namespace']}.svc.cluster.local:8080"
+        return f"http://managed-mcp-supergateway-{connector.id.hex[:16]}.{self._kubectl._default_kwargs['namespace']}.svc.cluster.local:8080"
 
     async def deploy(self, *, connector: Connector, preset: ConnectorPreset) -> None:
         assert preset.stdio
@@ -46,9 +46,9 @@ class ManagedMcpService:
                             "apiVersion": "apps/v1",
                             "kind": "Deployment",
                             "metadata": {
-                                "name": f"managed-mcp-{connector.id.hex[:16]}",
+                                "name": f"managed-mcp-server-{connector.id.hex[:16]}",
                                 "labels": {
-                                    "app": "managed-mcp",
+                                    "app": "managed-mcp-server",
                                     "connector-id": str(connector.id),
                                 },
                             },
@@ -56,19 +56,19 @@ class ManagedMcpService:
                                 "replicas": 1,
                                 "selector": {
                                     "matchLabels": {
-                                        "app": "managed-mcp",
+                                        "app": "managed-mcp-server",
                                         "connector-id": str(connector.id),
                                     }
                                 },
                                 "template": {
                                     "metadata": {
                                         "labels": {
-                                            "app": "managed-mcp",
+                                            "app": "managed-mcp-server",
                                             "connector-id": str(connector.id),
                                         }
                                     },
                                     "spec": {
-                                        "serviceAccountName": "managed-mcp",
+                                        "automountServiceAccountToken": False,
                                         "containers": [
                                             {
                                                 "name": "mcp-server",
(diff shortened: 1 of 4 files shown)


apps/agentstack-server/src/agentstack_server/service_layer/services/connector.py
```diff
@@ -54,13 +57,17 @@ async def create_connector(
         client_secret: str | None,
         metadata: Metadata | None,
         match_preset: bool = True,
+        bearer_token: str | None = None,
```
pilartomas: The `bearer_token` should be added to `connect_connector` rather than `create_connector`, connector can be "reconnected" with different token if necessary.

apps/agentstack-server/src/agentstack_server/service_layer/services/connector.py
```diff
@@ -371,10 +376,19 @@ def client_factory(headers=None, timeout=None, auth=None):
...
+                        (
+                            f"{await stack.enter_async_context(self._stdio_gateway(connector=connector, preset=preset))}/mcp"
+                            if (preset := self._find_preset(url=connector.url))
+                            and str(preset.url).startswith("mcp+stdio://")
+                            else str(connector.url)
+                        ),
+                        httpx_client_factory=client_factory,
+                    )
+                )
+                session = await stack.enter_async_context(ClientSession(read, write))
```
pilartomas: I don't think there is a bug in there but sticking with `async with` would really improve readability imo.

apps/agentstack-server/src/agentstack_server/service_layer/services/connector.py
```diff
@@ -392,17 +407,24 @@ async def mcp_proxy(self, *, connector_id: UUID, request: Request, user: User |
...
+                    target_url,
                     headers=forward_headers
                     | (
                         {"authorization": f"Bearer {connector.auth.token.access_token}"}
                         if connector.state == ConnectorState.connected
                         and connector.auth
                         and connector.auth.token
                         and connector.auth.token.token_type == "bearer"
+                        else {"authorization": f"Bearer {preset.bearer_token}"}
+                        if preset and preset.bearer_token
```
pilartomas: The token itself must be taken from `connector`. It would be best to reuse the existing connector structure the rest of the connectors service should then treat the token uniformly when probing, clearing, etc.

apps/agentstack-server/src/agentstack_server/service_layer/services/connector.py
pilartomas: Probing must also use the bearer token

JanPokorny:
TODO in this PR:
- [x] Working implementation using pods with supergateway sidecar
- [x] Split OAuth (external) and Kubernetes (internal) MCP server logic into separate services
- [ ] Start managed MCP server on "connect", not "create"
- [ ] Revise connect/disconnect/probe logic to be applicable to both
- [ ] Review feedback
- [ ] Propagate bearer token to env vars of provider
- [ ] Docs

For future PR:
- [ ] Allow `http` MCP servers to be managed too
- [ ] Allow running MCP servers in an external cluster (like provider builds)

apps/agentstack-server/src/agentstack_server/domain/models/connector.py
```diff
@@ -37,6 +37,7 @@ class Authorization(BaseModel):
     flow: AuthFlow | None = None
     token: Token | None = None
     token_endpoint: AnyUrl | None = None
+    auth_token: str | None = None  # Simple token-based auth (Bearer token or env var)
```
JanPokorny: Not sure if separate field for "plain token" is better or whether I should piggyback it on `token` here

pilartomas: piggyback, both is bearer token and the authlib client knows how to work with it.

apps/agentstack-server/src/agentstack_server/api/routes/connectors.py
```diff
@@ -96,6 +97,7 @@ async def connect_connector(
             user=user.user,
             redirect_url=connect_request.redirect_url,
             callback_uri=str(request.url_for(oauth_callback.__name__)),
+            auth_token=connect_request.auth_token,
```
pilartomas: Do not return the token, improves security.

JanPokorny: I'm not returning it to client, `_to_response` does not propagate it to response

apps/agentstack-server/src/agentstack_server/domain/models/connector.py
pilartomas: also, it will make it applicable to remote MCP servers that accept PATs etc.

apps/agentstack-server/src/agentstack_server/configuration.py
```diff
@@ -219,11 +219,20 @@ class ManagedProviderConfiguration(BaseModel):
...
+    env: dict[str, str] = Field(default_factory=dict)
+    auth_token_env_name: str | None = None
+
+
 class ConnectorPreset(BaseModel):
     url: AnyUrl
     client_id: str | None = None
     client_secret: str | None = None
     metadata: dict[str, str] | None = None
+    stdio: ConnectorStdioPreset | None = None
```
pilartomas, line 235: Nice to have: Validation that url has specific schema when stdio is used and vice-versa

apps/agentstack-server/src/agentstack_server/service_layer/services/connector.py
```diff
@@ -60,11 +62,17 @@ async def create_connector(
...
             )
 
-        preset = self._find_preset(url=url) if match_preset else None
+        preset = (
+            next((p for p in self._configuration.connector.presets if str(p.url) == str(url)), None)
+            if match_preset
+            else None
+        )
+        if not preset and url.scheme not in {"http", "https"}:
+            raise PlatformError("Unknown connector preset", status_code=status.HTTP_400_BAD_REQUEST)
```
pilartomas: Nitpick: Unknown preset should not be an error, it is completely valid not to have one so this error is misleading. Prefer `Invalid connector URL`.

apps/agentstack-server/src/agentstack_server/service_layer/services/connector.py
```diff
@@ -60,11 +62,17 @@ async def create_connector(
...
+
         if preset:
-            if not client_id:
-                client_id = preset.client_id
-                client_secret = preset.client_secret
+            client_id = client_id or preset.client_id
+            client_secret = client_secret or preset.client_secret
```
pilartomas: Must ensure that secret is not used with incorrect client_id.

Issue #1558, opened 2025-11-13
Investigate Carbon AI Chat compatibility and messaging layer integration


jenna-winkler:
## Description

We received a question about compatibility with Carbon AI Chat, IBM's open-source chat UI component. This raises an important architectural question about how Agent Stack can serve as backend infrastructure for Carbon while maintaining our existing lightweight GUI for demos and prototyping.

## Background

Carbon AI Chat is IBM's UI solution for internal deployments. Teams need a "messaging layer that is compatible with Carbon AI Chat" that can connect to Agent Stack's infrastructure.

Agent Stack's current GUI was built as a lightweight, open-source solution for the broader community with direct A2A protocol integration.

Based on initial investigation of [[Carbon AI Chat docs](https://chat.carbondesignsystem.com/tag/latest/docs/documents/Overview.html), it appears we can connect Carbon to the platform. The client SDK (#1229) would likely help facilitate this integration.

## Related Issues

- #1229 (may help with Carbon integration)

PR #1609, opened 2025-11-19
feat: extending TS sdk

tomkis:
## Summary
Extends the TypeScript SDK with:
- Server API client
- Simple implementation of LLM fullfilment resolver
- Helper to build A2A Agent Stack messages

## Linked Issues
Preparation work for #1558 

## Documentation
- [x] No Docs Needed:

Docs will be handled separately in https://github.com/i-am-bee/agentstack/issues/1617

### DIFF
--- a/apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
+++ b/apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
@@ -5,8 +5,20 @@
 
 import type { z } from 'zod';
 
-import type { ContextPermissionGrant, GlobalPermissionGrant, ModelCapability } from './types';
+import type { ContextPermissionsGrant, GlobalPermissionsGrant, ModelCapability } from './types';
 import { contextSchema, contextTokenSchema, modelProviderMatchSchema } from './types';
+
+interface MatchProvidersParams {
+  suggestedModels: string[] | null;
+  capability: ModelCapability;
+  scoreCutoff: number;
+}
+
+interface CreateContextTokenParams {
+  contextId: string;
+  globalPermissions: GlobalPermissionsGrant;
+  contextPermissions: ContextPermissionsGrant;
+}
 
 export const buildApiClient = ({ baseUrl }: { baseUrl: string } = { baseUrl: '' }) => {
   async function callApi<T>(method: 'POST', url: string, data: Record<string, unknown>, resultSchema: z.ZodSchema<T>) {
@@ -26,26 +38,16 @@ export const buildApiClient = ({ baseUrl }: { baseUrl: string } = { baseUrl: '' 
   }
 
   const createContext = async (providerId: string) => {
-    const { id: contextId } = await callApi(
-      'POST',
-      '/api/v1/contexts',
-      { metadata: {}, provider_id: providerId },
-      contextSchema,
-    );
+    const context = await callApi('POST', '/api/v1/contexts', { metadata: {}, provider_id: providerId }, contextSchema);
 
-    return contextId;
+    return context;
   };
 
-  const createContextToken = async (
-    contextId: string,
-    globalPermissions: GlobalPermissionGrant,
-    contextPermissions: ContextPermissionGrant,
-  ) => {
+  const createContextToken = async ({ contextId, globalPermissions, contextPermissions }: CreateContextTokenParams) => {
     const token = await callApi(
       'POST',
       `/api/v1/contexts/${contextId}/token`,
       {
-        context_id: contextId,
         grant_global_permissions: globalPermissions,
         grant_context_permissions: contextPermissions,
       },
@@ -55,7 +57,7 @@ export const buildApiClient = ({ baseUrl }: { baseUrl: string } = { baseUrl: '' 
     return { token, contextId };
   };
 
-  const matchProviders = async (suggestedModels: string[], capability: ModelCapability, scoreCutoff: number) => {
+  const matchProviders = async ({ suggestedModels, capability, scoreCutoff }: MatchProvidersParams) => {
     return await callApi(
       'POST',
       '/api/v1/model_providers/match',
--- a/apps/agentstack-sdk-ts/src/client/api/types.ts
+++ b/apps/agentstack-sdk-ts/src/client/api/types.ts
@@ -7,11 +7,11 @@ import { z } from 'zod';
 
 export const contextSchema = z.object({
   id: z.string(),
-  created_at: z.string().optional(),
-  created_by: z.string(),
-  last_active_at: z.string(),
-  provider_id: z.string().nullable(),
+  created_at: z.string(),
   updated_at: z.string(),
+  last_active_at: z.string(),
+  created_by: z.string(),
+  provider_id: z.string().nullable(),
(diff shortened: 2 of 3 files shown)


apps/agentstack-sdk-ts/src/client/api/types.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const contextSchema = z.object({
+  id: z.string(),
+  created_at: z.string().optional(),
+  created_by: z.string(),
+  last_active_at: z.string(),
+  provider_id: z.string().nullable(),
+  updated_at: z.string(),
+  metadata: z.record(z.string(), z.unknown()).nullable(),
+});
```
PetrBulanek, line 16: Is everything typed correctly? It seems strange to me that, for example, `created_at` is optional, but `updated_at` is not. Where is the source of truth for this within the Python SDK?

apps/agentstack-sdk-ts/src/client/api/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  feedback?: 'write'[];
+  files?: ('read' | 'write' | 'extract' | '*')[];
+  llm?: ('*' | ResourceIdPermission)[];
+  mcp_providers?: ('read' | 'write' | '*')[];
+  mcp_proxy?: '*'[];
+  mcp_tools?: ('read' | '*')[];
+  model_providers?: ('read' | 'write' | '*')[];
+  provider_variables?: ('read' | 'write' | '*')[];
+  providers?: ('read' | 'write' | '*')[];
+  vector_stores?: ('read' | 'write' | 'extract' | '*')[];
```
PetrBulanek: `vector_stores` don't have `extract`. See this change https://github.com/i-am-bee/agentstack/pull/1576/files

apps/agentstack-sdk-ts/src/client/api/types.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export interface ContextPermissionGrant {
+  context_data?: ('read' | 'write' | '*')[];
+  files?: ('read' | 'write' | 'extract' | '*')[];
+  vector_stores?: ('read' | 'write' | 'extract' | '*')[];
```
PetrBulanek: Same as above.

apps/agentstack-sdk-ts/src/client/api/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  mcp_providers?: ('read' | 'write' | '*')[];
+  mcp_proxy?: '*'[];
+  mcp_tools?: ('read' | '*')[];
+  model_providers?: ('read' | 'write' | '*')[];
+  provider_variables?: ('read' | 'write' | '*')[];
+  providers?: ('read' | 'write' | '*')[];
+  vector_stores?: ('read' | 'write' | 'extract' | '*')[];
+}
+
+export interface ContextPermissionGrant {
```
PetrBulanek: It is actually called `ContextPermissionsGrant` (plural) in Python SDK. I think we should standardize it in both cases. However, I am not sure whether it is better to use the singular or plural form.

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const createContext = async (providerId: string) => {
+    const { id: contextId } = await callApi(
+      'POST',
+      '/api/v1/contexts',
+      { metadata: {}, provider_id: providerId },
+      contextSchema,
+    );
+
+    return contextId;
```
PetrBulanek: Shouldn't we return the entire context object and not just its id?

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -0,0 +1,74 @@
...
+  const createContextToken = async (
+    contextId: string,
+    globalPermissions: GlobalPermissionGrant,
+    contextPermissions: ContextPermissionGrant,
+  ) => {
+    const token = await callApi(
+      'POST',
+      `/api/v1/contexts/${contextId}/token`,
+      {
+        context_id: contextId,
```
PetrBulanek: `context_id` should not be needed here. It is included in the path.

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -0,0 +1,74 @@
...
+      contextSchema,
+    );
+
+    return contextId;
+  };
+
+  const createContextToken = async (
+    contextId: string,
+    globalPermissions: GlobalPermissionGrant,
+    contextPermissions: ContextPermissionGrant,
```
PetrBulanek: Can't these params be optional?

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const matchProviders = async (suggestedModels: string[], capability: ModelCapability, scoreCutoff: number) => {
+    return await callApi(
+      'POST',
+      '/api/v1/model_providers/match',
+      {
+        capability,
+        score_cutoff: scoreCutoff,
+        suggested_models: suggestedModels,
```
PetrBulanek, line 67: `suggested_models` can also be optional, right?

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -0,0 +1,74 @@
...
+        grant_global_permissions: globalPermissions,
+        grant_context_permissions: contextPermissions,
+      },
+      contextTokenSchema,
+    );
+
+    return { token, contextId };
+  };
+
+  const matchProviders = async (suggestedModels: string[], capability: ModelCapability, scoreCutoff: number) => {
```
PetrBulanek: Shouldn't we use an object as a parameter for each API function? I think it would be easier to read when using it. :-)

PetrBulanek (changes requested):

apps/agentstack-sdk-ts/src/client/api/types.ts
tomkis: Good point, i forgot to check this. I did some copy paste from the schema but it seemed wrong from me, so i'll check what the server actually produces.

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
tomkis: I dont think they should. 

You can generate token without specifying grants. It's integral part of token generation.

apps/agentstack-sdk-ts/src/client/api/types.ts
tomkis: I will rename it for the sake of consistency.

However, we have `GlobalPermissionGrant` on the server which is incosistent so it still feels weird.

PetrBulanek: @jezekra1 Would it make sense to rename it on the server as well, so that the name is consistent?

PetrBulanek (approved):

PR #1616, opened 2025-11-20
fix(server): include redirect_uri into connector's token request

pilartomas:
## Summary
Ensures compliance with https://www.rfc-editor.org/rfc/rfc6749#section-4.1.3

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-server/src/agentstack_server/domain/models/connector.py
+++ b/apps/agentstack-server/src/agentstack_server/domain/models/connector.py
@@ -16,7 +16,8 @@ class AuthorizationCodeFlow(BaseModel):
     authorization_endpoint: AnyUrl
     state: str
     code_verifier: str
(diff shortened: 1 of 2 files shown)


apps/agentstack-server/src/agentstack_server/domain/models/connector.py
```diff
@@ -16,7 +16,8 @@ class AuthorizationCodeFlow(BaseModel):
     authorization_endpoint: AnyUrl
     state: str
     code_verifier: str
-    redirect_url: AnyUrl | None
+    redirect_uri: str
+    client_redirect_uri: AnyUrl | None
```
pilartomas: We could do this in non-breaking fashion but given flow is short-lived I opted for renaming to promote clarity.

PR #1625, opened 2025-11-21
feat(ui): add agent management under Providers feature flag

PetrBulanek:
## Summary
Removes the “Add Agent” button from the GUI for all roles to reduce confusion and keep the interface focused on end-user actions, with the CLI remaining the primary method for managing agents. Agent management is actually not removed from the UI entirely - it is now hidden behind the `Providers` feature flag.

## Linked Issues
Closes: https://github.com/i-am-bee/agentstack/issues/1563

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/modules/settings/SettingsView.tsx
+++ b/apps/agentstack-ui/src/modules/settings/SettingsView.tsx
@@ -16,9 +16,8 @@ import { ViewStack } from '#components/ViewStack/ViewStack.tsx';
 import { useApp } from '#contexts/App/index.ts';
 import { ConnectorsView } from '#modules/connectors/components/ConnectorsView.tsx';
 import { ProvidersView } from '#modules/providers/components/ProvidersView.tsx';
+import { useCanManageProviders } from '#modules/providers/hooks/useCanManageProviders.ts';
 import { VariablesView } from '#modules/providers/variables/components/VariablesView.tsx';
-import { useUser } from '#modules/users/api/queries/useUser.ts';
-import { isUserAdminOrDev } from '#modules/users/utils.ts';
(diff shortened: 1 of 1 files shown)


apps/agentstack-ui/src/components/Sidebar/AgentsNav.tsx
```diff
@@ -32,13 +35,13 @@ export function AgentsNav({ className }: Props) {
 
   const action = useMemo(
     () =>
-      isAdminOrDev
+      featureFlags.Providers && isAdminOrDev
```
tomkis: Would be nice to extract this into some feature flags / business logic ruleset hook to make it DRY. I can easily see someone missing either of the rules.

PetrBulanek: Good idea, I've updated it.

apps/agentstack-ui/src/modules/settings/SettingsView.tsx
```diff
@@ -25,10 +27,20 @@ export function SettingsView() {
...
+
+  const isAdminOrDev = isUserAdminOrDev(user);
 
   const items = useMemo(
-    () => ITEMS.filter(({ featureName }) => !featureName || featureFlags[featureName]),
-    [featureFlags],
+    () =>
+      ITEMS.filter(({ featureName }) => {
+        if (featureName === 'Providers') {
+          return featureFlags[featureName] && isAdminOrDev;
```
kapetr: I think `useCanManageProviders` can be used here as well.

PetrBulanek: I missed that, good catch! :-)

Issue #1543, opened 2025-11-12
Docs - OIDC configuration in Helm


tomkis:
The platform currently supports user auth via OIDC provider.

We need to document how to set it up in the Helm, as well as how to get this running in local installations.

PR #1660, opened 2025-11-28
chore: refactoring auth providers config & allow auth0 support

### LINKED ISSUE #1659: Provide audience for Auth0 auth provider
Integrating auth0 with agentstack is not possible due to fact that Auth0 requires audience in auth parameter in order to issue JWT that can be checked via JWKS.

When not provided the auth0 provides opaque token and does not provide introspect endpoint to verify.

tomkis:
## Summary
- Refactoring of auth provider configs (validation)
- Introduced support for Auth0 via audience param

## Linked Issues
Fixes #1659 

## Documentation
- [x] No Docs Needed:

Docs wil be handled in #1543


apps/agentstack-ui/src/app/(auth)/providers/ibm.ts
```diff
@@ -111,13 +113,17 @@ export interface IbmProfile extends Record<string, any> {
...
-export function IBMProvider(config: OIDCConfig<IbmProfile>): OIDCConfig<IbmProfile> {
+export function IBMProvider(config: ProviderConfig): OIDCConfig<IbmProfile> {
   return {
     id: config.id,
     name: config.name,
     type: 'oidc',
     idToken: true,
     style: { text: '#ffffff', bg: '#252525' },
-    options: config,
+    options: {
```
tomkis: review carefuly

apps/agentstack-ui/src/app/(auth)/auth.ts
```diff
@@ -4,63 +4,66 @@
...
 import { getTokenRefreshSchedule, jwtWithRefresh, RefreshTokenError } from './utils';
 
 let providersConfig: ProviderConfig[] = [];
 
-const providers: ProviderWithId[] = [];
-
 export const AUTH_COOKIE_NAME = 'agentstack';
 
 const { isAuthEnabled } = runtimeConfig;
+3;
```
kapetr: Typo?

apps/agentstack-ui/src/app/(auth)/providers/ibm.ts
kapetr: While these are sufficient for basic OIDC authentication, this might be too restrictive for some use cases, users might need to include other OIDC allowed options (https://next-auth.js.org/configuration/providers/oauth#options).
I'd revert this to full config spread, or allow additional options via `options` prop in provider config JSON.

apps/agentstack-ui/src/app/(auth)/auth.ts
```diff
@@ -4,63 +4,66 @@
...
     if (!providersJson) {
       throw new Error('No OIDC providers configured. Set OIDC_PROVIDERS with at least one provider.');
     }
 
-    providersConfig = JSON.parse(providersJson);
+    providersConfig = z.array(providerConfigSchema).parse(JSON.parse(providersJson));
+
+    return providersConfig.map((providerConfig) => {
+      return match(providerConfig)
+        .with({ name: 'w3id' }, { name: 'ibmid' }, { name: 'ibm' }, { name: 'ibmid-pkce' }, (provider) => {
```
kapetr: I think we might want to keep this case insensitive, it didn't match my provider with `name: 'IBMid'`.

kapetr: Otherwise we'd probably have to to add `label` to provider config to control the name rendered on the UI. It might be good idea to add that anyway with a fallback to `name` 🤔 

Image

tomkis: Yeah exactly. Lets separate name from class.

apps/agentstack-ui/src/app/(auth)/types.ts
```diff
@@ -4,15 +4,41 @@
...
+  issuer: z.string(),
+});
 
-export type ProviderWithId = Provider & {
-  id: string;
-};
+const baseProviderConfigSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  type: z.unknown(),
```
kapetr: Shouldn't this be `z.string()`?

apps/agentstack-ui/.env.example
```diff
@@ -22,7 +22,7 @@ NEXT_PUBLIC_APP_FAVICON_SVG=
 
 # Default false
 OIDC_ENABLED=
-OIDC_PROVIDERS='[{"name": "w3id","id": "w3id","type": "oidc","client_id": "<your_client_id>","client_secret": "<your_client_secret>","issuer": "your_issuer"}]'
```
tomkis, line 25: This exists to keep backwards compatibility. `type` is no longer needed, newly added custom is discriminated union for scehma validation.

apps/agentstack-ui/src/app/(auth)/auth.ts
```diff
@@ -4,63 +4,80 @@
...
+      },
+    }))
+    .exhaustive();
+
+  return {
+    id: config.id,
+    name: config.name,
+    type: 'oidc',
+    idToken: true,
+    options,
```
kapetr, line 55: This could be a breaking change - in the previous version, the entire config object was passed into the `options`, now only some selected props are used. OIDCConfig allow configuring more than what is currently exposed to the `options`, and the previous implementation supported that.

tomkis: Understood but honestly i think that this was wrong. It was just too unpredictable and if someone relied on that then we cant guarantee non-breaking changes. 

We need to mention this though in our release notes cc @jenna-winkler

PR #1676, opened 2025-12-03
fix(ui): stabilize mermaid diagram rendering while streaming

### LINKED ISSUE #1672: UI - Investigate Streamdown for streaming markdown rendering
## Problem

Currently experiencing flickering issues when rendering streaming markdown content, particularly with Mermaid diagrams and other complex markdown elements during agent responses. The current implementation renders content continuously as it streams, causing visual instability.

## Current hotfix

We've implemented a temporary solution that delays rendering until after streaming completes - showing only raw code blocks during streaming and rendering the final diagram/markdown once the stream finishes. While this eliminates flicker, it's not ideal for user experience as it doesn't provide progressive visual feedback.

## Proposed investigation

Evaluate [Streamdown](https://streamdown.ai/) as a potential long-term replacement for our current markdown rendering approach. Streamdown is an open-source library from Vercel specifically designed to handle streaming markdown with built-in support that should eliminate flicker issues.

## Key Considerations

- Need to verify all markdown features we currently rely on render correctly with Streamdown
- Not a trivial migration - requires thorough testing of all markdown use cases
- May benefit Canvas feature as well

kapetr:
### Proposed solution with Streamdown
The suggested Streamdown-based approach turned out not to be a good fit for our use case. While it works well on its own, it relies on Tailwind CSS and offers limited customization. In particular, its Mermaid and code block components cannot be used independently — it’s an all-or-nothing setup. To use their Mermaid support, we would have had to drop our custom Code component.

### Final solution
Fortunately, our react-markdown (currently used) already supports rendering incomplete blocks. Instead of adopting Streamdown, custom Mermaid diagram caching was implemented to stabilize diagram rendering, eliminate flicker during message streaming, and ensure better streaming experience.

## Linked Issues
Closes #1672

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
+++ b/apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
@@ -30,8 +30,6 @@ export function MermaidDiagram({ children, mermaidIndex, isStreaming }: MermaidD
 
   const diagram = diagrams.get(index);
 
-  console.log({ index, isStreaming, diagram });
-
   useEffect(() => {
     mermaid.initialize({
       startOnLoad: false,


apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
```diff
@@ -3,26 +3,27 @@
...
 
-export function MermaidDiagram({ showDiagram = true, children }: MermaidDiagramProps) {
+export function MermaidDiagram({ children, ...props }: MermaidDiagramProps) {
   const id = useId();
-  const [diagram, setDiagram] = useState<string | null>(null);
-
   const { theme } = useTheme();
+  const { diagrams, setDiagram } = useMermaid();
+
+  const index = props.mermaidIndex ?? 0;
```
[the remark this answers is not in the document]
kapetr: While that’s true, I think it’s better to let it render and simply log the error, since in most cases there will be at most one Mermaid diagram in the response message.

apps/agentstack-ui/src/components/MarkdownContent/MarkdownContent.tsx
```diff
@@ -40,24 +39,26 @@ export function MarkdownContent({
     () => ({
       ...components,
       code: ({ ...props }) => <Code {...props} forceExpand={codeBlocksExpanded} />,
-      mermaidDiagram: ({ ...props }) => <MermaidDiagram {...props} showDiagram={showMermaidDiagrams} />,
+      mermaidDiagram: (props) => <MermaidDiagram {...props} />,
```
PetrBulanek: As we don't need to customize the props any more, we can move the component back here `apps/agentstack-ui/src/components/MarkdownContent/components/index.ts` to default components. Actually, it is already there, so we can just remove it. :-)

apps/agentstack-ui/src/components/MarkdownContent/remark/remarkMermaidIndex.ts
PetrBulanek, line 1: Why not to make it part of the `remarkMermaid` function?

PetrBulanek (changes requested):

apps/agentstack-ui/src/components/MarkdownContent/MarkdownContent.tsx
kapetr: Turns out we do need it in the end :)

apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
```diff
@@ -3,48 +3,61 @@
...
+  const { diagrams, setDiagram } = useMermaid();
+
+  if (mermaidIndex === undefined) {
+    console.error('MermaidDiagram component requires a `mermaidIndex` prop.');
+  }
+  const index = mermaidIndex ?? 0;
+
+  const diagram = diagrams.get(index);
+
+  console.log({ index, isStreaming, diagram });
```
PetrBulanek: Forgotten console.log.

Issue #1694, opened 2025-12-09
[Feature]: Auto-inject MCP fulfillments from connected OAuth connectors


tomkis:
**Is your feature request related to a problem? Please describe.**
Agents declare MCP demands but users must manually construct fulfillments. When connector requires OAuth (Stripe, GitHub, Box), agent can't act on user's behalf without passing connected connector reference.

**Describe the solution you'd like**
Auto-match agent MCP demands to user's connected connectors via `metadata.name`. Transparently inject fulfillments pointing to connector MCP proxy at agent runtime. No UI changes - server-side only.

Implementation approach:
- Case-insensitive name matching between MCP demands and connector metadata
- Multiple connectors with same name: use latest (most recently created)
- Client-provided fulfillments take precedence over auto-injected
- No logging for injection events

**Additional context**
Benefits:
- Agents automatically use OAuth-connected services
- No manual fulfillment construction
- Leverage existing connector auth flow

Risks to address:
- Multiple connectors same name creates ambiguity (mitigated by using latest)
- Connector disconnects mid-run causes auth error (acceptable, existing error paths handle)
- Name collision between different services (user responsibility to manage)
- Client-provided fulfillments may conflict (mitigated by client precedence)

Issue #1542, opened 2025-11-12
Docs - consolidate MCP documentation


tomkis:
The current problems with the MCP docs:

- https://agentstack.beeai.dev/guides/connectors should only contain details for integrators, not devleopers nor consumers
- https://agentstack.beeai.dev/extensions/mcp - needs complete overhaul and unification with https://agentstack.beeai.dev/extensions/mcp-oauth
- We need to clearly articulate what's the difference between connectors and MCP OAuth and how it's being used
- Ideally just one page about MCP, which mentions options for OAuth.

PR #1695, opened 2025-12-09
feat: Auto inject connected MCPs in UI

tomkis:
## Summary
- Adjusted the GUI code to automatically resolve MCP demands using connectors if available
- Removed old approach of adding MCPs via textbox
- Reworked MCP extension to optionally provide None for created MCP client instead of throwing error when no fullfilment is provided
- Added Github MCP agent example
- Consolidated docs for MCP #1542
- Introduced cursor commands for ticket filing as well as documentation

## Linked Issues
Solves #1694, #1542  

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/modules/connectors/components/ConnectorActionButton.tsx
+++ b/apps/agentstack-ui/src/modules/connectors/components/ConnectorActionButton.tsx
@@ -6,6 +6,7 @@
 import type { CarbonIconType } from '@carbon/icons-react';
 import { Credentials, Link, Unlink } from '@carbon/icons-react';
 import { IconButton } from '@carbon/react';
+import { ConnectorState } from 'agentstack-sdk';
 import { useMemo } from 'react';
 import { match } from 'ts-pattern';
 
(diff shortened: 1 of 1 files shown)


apps/agentstack-sdk-py/src/agentstack_sdk/a2a/extensions/services/mcp.py
```diff
@@ -39,7 +39,7 @@ class StdioTransport(pydantic.BaseModel):
 class StreamableHTTPTransport(pydantic.BaseModel):
     type: Literal["streamable_http"] = "streamable_http"
 
-    url: pydantic.AnyHttpUrl
+    url: str
```
tomkis, line 42: The problem is that client can provide `{platform_url}/foobar` like in other url fields.

jezekra1 (approved):

apps/agentstack-sdk-ts/src/client/api/types.ts
kapetr, line 1: Do we want to keep all api types and schemas in one file?

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -86,7 +105,11 @@ export const buildApiClient = (
     );
   };
 
-  return { createContextToken, createContext, matchProviders };
+  const listConnectors = async () => {
+    return await callApi('GET', '/api/v1/connectors', null, listConnectorsResponseSchema);
+  };
```
kapetr, line 110: Just wondering, do we want to keep all API calls in this file? Nothing critical, but I'd consider splitting them  by modules.

apps/agentstack-ui/src/modules/connectors/api/index.ts
```diff
@@ -72,5 +74,5 @@ interface Created {
 
 export async function listConnectors(): Promise<ListConnectorsResponse | undefined> {
   const response = await api.GET('/api/v1/connectors', {});
-  return ensureData(response);
+  return ensureData(response) as ListConnectorsResponse;
```
kapetr: Why not using `agentstackClient.listConnectors()`? Or delete the function altogether and use it directly.

PR #1701, opened 2025-12-10
feat(ui): add missing embedding model error page

### LINKED ISSUE #1673: UI - Missing embedding model error state
## Current behavior

If a user configures an LLM but does not configure an embedding model, any agent that depends on embeddings shows a generic error screen, which doesn’t indicate the real problem.

## Expected behavior

The UI should clearly tell the user that an embedding model must be configured, similar to the existing UI message shown when an LLM is missing (see reference below).

However, it should say:
> There is no embedding model selected.
> You can configure an embedding model by running `agentstack model add embedding` in your terminal.

## Steps to reproduce

1. Configure an LLM only (leave embedding model unconfigured)
2. Open the out-of-the-box RAG agent
3. View the agent error state

## Screenshot

Image

## Reference (LLM missing state)

Image

kapetr:
## Summary
Check the selected models against the actual agent demands, and provide an error message for a missing embedding model configuration, analogous to the existing error for a missing main model.

## Linked Issues
Closes #1673

## Documentation
- [x] No Docs Needed: a fix

### DIFF
--- a/apps/agentstack-ui/src/app/(main)/run/[providerId]/c/[contextId]/page.tsx
+++ b/apps/agentstack-ui/src/app/(main)/run/[providerId]/c/[contextId]/page.tsx
@@ -11,7 +11,7 @@ import { fetchContextHistory } from '#modules/platform-context/api/index.ts';
 import { PlatformContextProvider } from '#modules/platform-context/contexts/PlatformContextProvider.tsx';
 import { RunView } from '#modules/runs/components/RunView.tsx';
 
-import { ensureModelSelected } from '../../model-check';
+import { ensureModelSelected } from '../../ensure-model-selected';
 import { fetchAgent } from '../../rsc';
 
 interface Props {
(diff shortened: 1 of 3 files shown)


apps/agentstack-ui/src/app/(main)/run/[providerId]/c/[contextId]/page.tsx
```diff
@@ -18,6 +20,14 @@ interface Props {
 
 export default async function AgentRunPage({ params }: Props) {
   const { providerId, contextId } = await params;
+  const { featureFlags } = runtimeConfig;
+
+  if (featureFlags.LocalSetup) {
+    const { error } = await ensureModelSelected(providerId);
+    if (error) {
+      return error;
+    }
```
Comment (PetrBulanek) [inline apps/agentstack-ui/src/app/(main)/run/[providerId]/c/[contextId]/page.tsx]: It's not necessary to change it, but I personally would find it a little clearer if it returned the component, like this:

```
const { ErrorComponent } = await ensureModelSelected(providerId);

if (ErrorComponent) {
    return <ErrorComponent />;
}
```

At least at first glance, it indicates that something is being rendered. What do you think?

Comment (kapetr) [reply apps/agentstack-ui/src/app/(main)/run/[providerId]/c/[contextId]/page.tsx]: I like that, I was also struggling with this not being obvious.

PR #1737, opened 2025-12-18
feat(ui): move API to SDK

PetrBulanek:
## Summary
- Replace deprecated `openapi-fetch` client in UI with `agentstack-sdk` client usage across configuration, connectors, contexts, files, model providers, providers, user feedback, users and variables modules; remove legacy UI API types/utils.
- Extend the SDK with a new API core (typed ApiResult, zod validation, streaming responses, file upload FormData handling, and structured error types).
- Add A2A protocol schemas/types to the SDK and reorganize extension schemas/types; update UI A2A imports and error handling to use SDK helpers.
- Improve error handling across SDK/UI, including typed API error inspection.

## Linked Issues
Refs: https://github.com/i-am-bee/agentstack/issues/1692

## Documentation
- [x] No Docs Needed:

This PR is already huge in itself, so the documentation will be part of additional PR.

### DIFF
--- a/apps/agentstack-ui/src/modules/agents/components/cards/AgentsList.tsx
+++ b/apps/agentstack-ui/src/modules/agents/components/cards/AgentsList.tsx
@@ -4,10 +4,12 @@
  */
 
 'use client';
+
+import type { ListProvidersResponse } from 'agentstack-sdk';
+
 import { SkeletonItems } from '#components/SkeletonItems/SkeletonItems.tsx';
 import { useListAgents } from '#modules/agents/api/queries/useListAgents.ts';
 import { ListAgentsOrderBy } from '#modules/agents/api/types.ts';
-import type { ListProvidersResponse } from '#modules/providers/api/types.ts';
 
 import classes from './AgentsList.module.scss';
 import { AgentsListItem } from './AgentsListItem';


apps/agentstack-sdk-ts/src/client/a2a/extensions/services/mcp/schemas.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import z from 'zod';
+
+import { MCPTransportType } from './types';
+
+export const mcpTransportTypeSchema = z.enum(MCPTransportType);
```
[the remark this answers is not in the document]
PetrBulanek: `nativeEnum` is deprecated and was merged into `enum`.

apps/agentstack-sdk-ts/src/index.ts
```diff
@@ -3,32 +3,16 @@
...
-} from './client/a2a/extensions/handle-task-status-update';
-export { resolveUserMetadata, type UserMetadataInputs } from './client/a2a/extensions/resolve-user-metadata';
-export * from './client/a2a/extensions/services/embedding';
-export * from './client/a2a/extensions/services/form';
-export * from './client/a2a/extensions/services/llm';
-export * from './client/a2a/extensions/services/mcp';
-export * from './client/a2a/extensions/services/oauth-provider';
-export * from './client/a2a/extensions/services/platform';
-export * from './client/a2a/extensions/services/secrets';
+export * from './client/a2a/extensions';
```
tomkis: Have you been considering creating directory structure that always contains `index.ts` and then the imports are a hierarchy of imported folders? Basically each module contains `index.ts` exporting everything from the module. 

It seems a bit awkward to enumerate internal details of each module here.

Obviously my suggestion is recursive (let's not over-abuse the depth - eg one max two levels of modules would do).

I am proposing having seperate modules for:

- Client
  - extensions
  - api
  - core

apps/agentstack-sdk-ts/src/extensions.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export * from './client/a2a/extensions';
```
tomkis, line 6: basically this can be renamed to index.ts and moved to `./client/a2a/extensions` to encapsulate all the stuff.

apps/agentstack-sdk-ts/src/client/a2a/protocol/schemas.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const textPartSchema = z.object({
+  kind: z.literal('text'),
+  text: z.string(),
+  metadata: z.record(z.string(), z.unknown()).nullish(),
+});
+
+export const fileWithBytesSchema = z.object({
+  bytes: z.string(),
+  mimeType: z.string().nullish(),
```
tomkis: I believe all the `nullish` here is wrong.

The protocol defines the values as `undefined` not `nullish`.

This won't work in environments where plain a2a is used. 

See https://a2a-protocol.org/latest/specification/#57-field-presence-and-optionality for further reference. cc @jezekra1

apps/agentstack-sdk-ts/src/client/a2a/runtime/buildLLMExtensionFulfillmentResolver.ts
```diff
@@ -3,23 +3,26 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import type { AgentstackClient } from '../../../api/build-api-client';
-import type { ContextToken } from '../../../api/types';
-import { ModelCapability } from '../../../api/types';
-import type { LLMDemands, LLMFulfillments } from '../services/llm';
+import type { ContextToken } from '../../api/types';
```
tomkis: I'd move this in `fulfillment-resolvers` folder, there will potentialy be more of those and it's sort of separate module.

apps/agentstack-sdk-ts/src/client/a2a/runtime/buildMessageBuilder.ts
```diff
@@ -5,8 +5,8 @@
 
 import type { AgentCapabilities, Message } from '@a2a-js/sdk';
 
-import type { Fulfillments } from '../handle-agent-card';
-import { handleAgentCard } from '../handle-agent-card';
+import { handleAgentCard } from './handleAgentCard';
```
tomkis: This is an `util` - i'd create here separate `util` folder and move it there.

apps/agentstack-sdk-ts/src/client/core/extensions/extract.ts
```diff
@@ -11,7 +11,7 @@ export function extractUiExtensionData<U extends string, D>(extension: A2AUiExte
   const schema = extension.getMessageMetadataSchema();
```
tomkis, line 10: I'd create extensions folder for everything related to extensions (that is not a2a schema definition).

so eg. this extract, fulfill

apps/agentstack-sdk-ts/src/client/core/extensions/resolve-user-metadata.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { canvasExtension } from '../extensions/ui/canvas';
+import { formRequestExtension } from '../extensions/ui/form-request';
+import type { UserMetadataInputs } from './types';
+
+export const resolveUserMetadata = async (inputs: UserMetadataInputs) => {
```
tomkis, line 10: Another extension specific file.

apps/agentstack-sdk-ts/src/client/api/core/client.ts
```diff
@@ -0,0 +1,147 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
```
tomkis, line 2: Shouldnt this whole module be owned by the api folder? It's api after all?

apps/agentstack-sdk-ts/src/client/api/core/errors/utils.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
tomkis, line 1: Another one that could be owned by api module.

apps/agentstack-ui/src/api/a2a/client.ts
```diff
@@ -42,7 +42,8 @@ function handleStatusUpdate<UIGenericPart = never>(
     return [];
   }
 
-  const metadataParts = processMessageMetadata(message);
+  // TODO: Resolve so that we don't have to use typecasting
+  const metadataParts = processMessageMetadata(message as Message);
```
kapetr: Can we do it now, or is that a note for later?

apps/agentstack-ui/src/api/agentstack-client.ts
```diff
@@ -7,12 +7,11 @@ import { buildApiClient } from 'agentstack-sdk';
...
 import { runtimeConfig } from '#contexts/App/runtime-config.ts';
-import { listConnectorsResponseSchema } from '#modules/connectors/api/types.ts';
 import { getBaseUrl } from '#utils/api/getBaseUrl.ts';
 
-import { getProxyHeaders, handleFailedResponse } from './utils';
+import { getProxyHeaders } from './utils';
 
-function buildAuthenticatedAgentstackClient() {
+function buildAuthenticatedAgentStackClient() {
   const { isAuthEnabled } = runtimeConfig;
```
kapetr, line 15: Just a note, not the subject of this PR - it works well, but the code is a bit misleading, because `isAuthEnabled` is always `false` on the client. Might be worth a small refactor, I think the `authenticatedFetch` function should be used on the server only.

apps/agentstack-ui/src/api/schema.d.ts
kapetr, line 1: Do we plan to keep the schema for future use? If not, I think we could remove it along with the generating task.

apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx
Comment (kapetr) [inline apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx:1]: I noticed that camelCase naming is used for functions in the SDK (e.g. `handleTaskStatusUpdate.ts`). Should we switch to the same convention in the agentstack-ui, or vice versa? (I don't have a preference 😄 )

apps/agentstack-sdk-ts/src/client/api/core/client.ts
```diff
@@ -0,0 +1,155 @@
...
+            request,
+            response,
+            details: parseError,
+          } satisfies ParseError,
+        } satisfies ApiFailure;
+      }
+
+      const { data: result, success, error } = schema.safeParse(parsedBody);
+
+      if (!success) {
```
kapetr, line 69: Maybe we could add a warning to the dev console to help with debugging.

kapetr (changes requested):
The SDK fails to validate `/providers` response as the API doesn't align with a2a types on `null` values, while the SDK schema does.

apps/agentstack-ui/src/api/a2a/client.ts
PetrBulanek: Leftover from gradual refactoring. No longer needed. Thank you!

apps/agentstack-ui/src/api/schema.d.ts
PetrBulanek: Actually, I thought about removing it completely, but I figured it would be better to discuss it first. :-)

apps/agentstack-ui/src/api/agentstack-client.ts
PetrBulanek: We can definitely adjust it. Will you take a look at it once we merge this PR or shall we create an issue?

apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx
Comment (PetrBulanek) [reply apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx:1]: I'm probably just used to using camelCase in the JS/TS environments, but the fact is that Next.js itself uses kebab-case, and that's how most of our files are, so I'll probably change it in the SDK too, to keep things consistent. So I'd leave kebab-case everywhere and only use PascalCase for React components. Agreed?

apps/agentstack-sdk-ts/src/client/api/core/client.ts
PetrBulanek: The SDK already returns everything, so I don't think it's necessary to log it here. Some people might not want that, so I would rather log it on the client side.

apps/agentstack-ui/src/api/agentstack-client.ts
kapetr: I will look at it.

apps/agentstack-sdk-ts/src/client/api/core/client.ts
kapetr: Yes, definitely client side (in agentstack-ui), wrongly placed comment. But we can do that later, if we realize we run into parsing issues frequently :)

apps/agentstack-sdk-ts/src/client/a2a/protocol/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+      return undefined;
+    }
+
+    return normalizeOptionalNulls(schema.unwrap(), data);
+  }
+
+  if (schema instanceof z.ZodObject) {
+    if (data === null || typeof data !== 'object' || Array.isArray(data)) {
+      return data;
+    }
```
kapetr, line 26: Wouldn't this bypass normalization for objects inside arrays?

Issue #1734, opened 2025-12-17
User Role Management - No API/CLI method to upgrade user to admin role


jenna-winkler:
## Problem

Currently there’s no way to upgrade a user from “user” to “admin” role without direct database access. When `/api/v1/user` returns `"role": "user"`, administrators cannot promote that user through API or CLI methods.

## Current behavior

- Adding user email to `AUTH__OIDC__ADMIN_EMAILS` environment variable and restarting does not upgrade existing users
- The env var appears to only be checked during initial user creation
- Role modification requires direct database access, which may not be available in all deployment scenarios (e.g., DB not on public network)

## Expected behavior

Administrators should be able to promote users to higher roles through API endpoints or CLI commands without requiring direct database access.

## Proposed solution

- Add API endpoint allowing admins to promote other users to higher roles
- Add endpoint for listing users (useful for identifying users to promote)
- Expose this functionality through agentstack CLI

PR #1740, opened 2025-12-19
feat: permission elevation

tomkis:
## Summary
Adds whole CLI section for listing users as well as setting role to individual users.

## Linked Issues
Solves #1734 

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-cli/src/agentstack_cli/commands/user.py
+++ b/apps/agentstack-cli/src/agentstack_cli/commands/user.py
@@ -83,9 +83,7 @@ async def set_role(
 
         role_display = ROLE_DISPLAY.get(result.new_role, result.new_role)
 
-        console.success(
-            f"User role updated to [cyan]{role_display}[/cyan] (version [yellow]{result.role_version}[/yellow])"
-        )
+        console.success(f"User role updated to [cyan]{role_display}[/cyan]")
 
 
 def _format_date(dt: datetime | None) -> str:
--- a/apps/agentstack-server/src/agentstack_server/api/auth/auth.py
+++ b/apps/agentstack-server/src/agentstack_server/api/auth/auth.py
@@ -92,7 +92,7 @@ class ParsedToken(BaseModel):
     context_permissions: Permissions
     context_id: UUID
     user_id: UUID
-    role_version: int
+    iat: int
     raw: dict[str, Any]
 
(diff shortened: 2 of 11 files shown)


apps/agentstack-cli/src/agentstack_cli/commands/user.py
```diff
@@ -0,0 +1,94 @@
...
+    console.print(table)
+
+    if has_more and next_page_token:
+        console.print(f"\n[dim]Use --after {next_page_token} to see more[/dim]")
+
+
+@app.command("set-role")
+async def set_role(
+    user_id: typing.Annotated[str, typer.Argument(help="User UUID")],
+    role: typing.Annotated[UserRole, typer.Option("--role", "-r", help="Target role")],
```
JanPokorny: This should be an argument, since it's integral

apps/agentstack-server/src/agentstack_server/api/auth/auth.py
```diff
@@ -119,6 +121,7 @@ def issue_internal_jwt(
             "global": global_permissions.model_dump(mode="json"),
             "context": context_permissions.model_dump(mode="json"),
         },
+        "role_version": role_version,
```
pilartomas: Why is it necessary to have the role version? Wouldn't it be better to compare `iat` with `role_updated_at`?

tomkis: Good point 👍

apps/agentstack-server/src/agentstack_server/application.py
```diff
@@ -118,6 +119,7 @@ async def custom_http_exception_handler(request: Request, exc: Exception):
 def mount_routes(app: FastAPI):
     server_router = APIRouter()
     server_router.include_router(user_router, prefix="/user")
+    server_router.include_router(users_router, prefix="/users")
```
pilartomas, line 122: Would be nice to deprecate `/user` endpoint and add one of the following (or both):

- `/users/me`
- `/users/{id}`

JanPokorny: I'll leave this for another PR since it involves deprecations

JanPokorny (approved):

Issue #1637, opened 2025-11-25
Secure A2A server in the SDK to validate Context Token


tomkis:
**Is your feature request related to a problem? Please describe.**
Currently, the AgentStack SDK lacks an authentication mechanism for its A2A endpoints. This means any internal service or component could potentially invoke these endpoints without being authenticated, leading to potential security concerns and unauthorized access.

**Describe the solution you'd like**
Introduce an authentication mechanism for all A2A endpoints in the AgentStack SDK. Every A2A request should include an Authorization header with a bearer token - that is platform's Context Token. The platform should validate this token to ensure that the calling service is authorized. This would make the SDK compliant with standard A2A authentication practices. 

**Additional context**
See https://a2a-protocol.org/latest/specification/#452-apikeysecurityscheme for further reference.

PR #1741, opened 2025-12-19
Add authorization to agents using cotext token

jezekra1:
## Summary

Important changes:
- a2a clients **can now use platform context token in the authorization header** instead of sending 2 tokens - user OIDC token in the header and context token in the platform api extension fullfilment 
- platform api extension fullfilment body is now optional, if a token is provided in the header and is a context token, it will be used
- a2a proxy on the server does "token exchange" -> if it receives a context token it creates a new context token with the same claims and expiration, but targeted "audience" in `aud` claim.
  * This is passed to the SDK using the `PLATFORM_AUTH__PUBLIC_URL` env for managed providers and validated.
  * This ensures that agents secured by platform authentication cannot use the token to call other agents directly, they must always go through the platform proxy.
  * This is important because direct access be a security hole where it's possible to exfiltrate data of other users using `task/list` or `task/get` methods from a2a, which are not properly scoped per user.  The a2a proxy is tracking task ownership and won't allow these methods if unauthorized.
 - All changes should be backwards compatible
 
Todos:
- [x] test backwards compatibility
- [x] test managed agents
- [ ] add sdk tests
- [x] improve frontend implementation

## Linked Issues

Ref: #1637 

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/agents/chat/pyproject.toml
+++ b/agents/chat/pyproject.toml
@@ -7,7 +7,7 @@ authors = [
 ]
 requires-python = ">=3.13,<3.14"
 dependencies = [
-    "beeai-framework[duckduckgo,wikipedia,a2a]>=0.1.74",
+    "beeai-framework[duckduckgo,wikipedia,a2a]==0.1.70",
     "tiktoken>=0.12.0",  # constraint for beeai-framework dependency (first version with musl arm64 wheel)
     "fastuuid>=0.14.0",  # constraint for beeai-framework dependency (first version with musl arm64 wheel)
     "openai>=1.107.1",
--- a/agents/chat/uv.lock
+++ b/agents/chat/uv.lock
@@ -238,7 +238,7 @@ wheels = [
 
 [[package]]
 name = "beeai-framework"
-version = "0.1.74"
+version = "0.1.70"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "aiofiles" },
@@ -252,9 +252,9 @@ dependencies = [
     { name = "pydantic-settings" },
     { name = "requests" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/4b/d5/19d2a202edf61e629148aa63ec35e0ca2a9073183a66499aaa8f36b1ff9f/beeai_framework-0.1.74.tar.gz", hash = "sha256:1b0dd9114e0099b0f43e63d8951f05407499e4c371e387e27ef123b803211c69", size = 189160, upload-time = "2025-12-22T12:57:49.157Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/a5/27/a26b1a55444de2271325d5471ba09e8a5f8c40bee25bf0f891eff52942ab/beeai_framework-0.1.70.tar.gz", hash = "sha256:7fe1b4f22015f8ea3ceffdbbaf1bba4ed390065d23c17aa3d406e25302ea24eb", size = 184126, upload-time = "2025-11-26T11:47:07.117Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/5c/b0/62b5646b80f91930363fc01d12dea595d401d28445dd5e0744c97f0cf3bc/beeai_framework-0.1.74-py3-none-any.whl", hash = "sha256:9958f79ba917b3769572df6507513bdecf5ca4d2f62d22389e93a3f4b7dd9ac6", size = 347954, upload-time = "2025-12-22T12:57:47.806Z" },
+    { url = "https://files.pythonhosted.org/packages/bb/50/fb1172b5c1625832b6589a1c6150197940b73c9aaeadfda4d7fef2067763/beeai_framework-0.1.70-py3-none-any.whl", hash = "sha256:a0e9350cc9a8360f2b8f2251ac96bcc41511ed24ac6f27bc7fb6c3e5dbc260e2", size = 340421, upload-time = "2025-11-26T11:47:05.498Z" },
 ]
 
 [package.optional-dependencies]
@@ -396,7 +396,7 @@ dev = [
 [package.metadata]
 requires-dist = [
     { name = "agentstack-sdk", editable = "../../apps/agentstack-sdk-py" },
-    { name = "beeai-framework", extras = ["duckduckgo", "wikipedia", "a2a"], specifier = ">=0.1.74" },
+    { name = "beeai-framework", extras = ["duckduckgo", "wikipedia", "a2a"], specifier = "==0.1.70" },
     { name = "fastuuid", specifier = ">=0.14.0" },
     { name = "openai", specifier = ">=1.107.1" },
     { name = "openinference-instrumentation-beeai", specifier = ">=0.1.12" },


apps/agentstack-ui/src/app/api/[...path]/route.ts
```diff
@@ -31,8 +31,10 @@ async function handler(request: NextRequest, context: RouteContext) {
     targetUrl += '/';
   }
   targetUrl += search;
-
-  if (isAuthEnabled) {
+  if (request.url.includes('/api/v1/a2a')) {
+    console.log(path, headers);
```
Comment (tomkis) [inline apps/agentstack-ui/src/app/api/[...path]/route.ts]: Dont forget to remove this.

apps/agentstack-ui/src/app/api/[...path]/route.ts
```diff
@@ -31,8 +31,10 @@ async function handler(request: NextRequest, context: RouteContext) {
...
+  }
+  if (isAuthEnabled && !(path[0] == 'v1' && path[1] == 'a2a')) {
```
Comment (tomkis) [inline apps/agentstack-ui/src/app/api/[...path]/route.ts]: What is the reason behind this ? Is that so UI won't send wrong tokens? Shouldn't we rather forbid this endpoint entirely from the client and access the server directly?

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
```diff
@@ -79,6 +79,10 @@ export const buildApiClient = (
     await callApi('POST', '/api/v1/contexts', { metadata: {}, provider_id: providerId }, contextSchema);
 
   const createContextToken = async ({ contextId, globalPermissions, contextPermissions }: CreateContextTokenParams) => {
+    if (!globalPermissions.a2a_proxy?.includes('*') && globalPermissions.a2a_proxy?.length === 0) {
```
tomkis: I'd move this validation to zod schema refinment in the types.ts - so that its not even possible to construct the permissions.

apps/agentstack-sdk-ts/src/client/api/types.ts
```diff
@@ -65,7 +65,7 @@ export const globalPermissionsGrantSchema = contextPermissionsGrantSchema.extend
   embeddings: z.array(z.union([z.literal('*'), resourceIdPermissionSchema])).optional(),
   model_providers: z.array(z.literal(['read', 'write', '*'])).optional(),
 
-  a2a_proxy: z.array(z.literal('*')).optional(),
+  a2a_proxy: z.array(z.union([z.literal('*'), z.string()])).optional(),
```
tomkis: I'd go with:

1. ['*']
2. string[] which does not include '*'
3. undefined.

Using refinments.

apps/agentstack-ui/src/api/a2a/agent-card.ts
```diff
@@ -8,10 +8,17 @@ import { A2AClient } from '@a2a-js/sdk/client';
 import { UnauthenticatedError } from '#api/errors.ts';
 import { getBaseUrl } from '#utils/api/getBaseUrl.ts';
 
-export async function getAgentClient(providerId: string) {
+export async function getAgentClient(providerId: string, token?: string) {
   const agentCardUrl = `${getBaseUrl()}/api/v1/a2a/${providerId}/.well-known/agent-card.json`;
 
-  return await A2AClient.fromCardUrl(agentCardUrl, { fetchImpl: clientFetch });
+  const fetchImpl = token
```
tomkis: This is something that has to be extracted to the SDK. 

We are effectively moving the platform auth from the extension to headers, so far we've been telling users to build their own A2A client and use extensions but this approach won't work.

apps/agentstack-ui/src/api/schema.d.ts
```diff
@@ -9,6 +9,23 @@
  */
 
 export interface paths {
+  '/.well-known/jwks': {
```
tomkis, line 12: This file will be deprecated in favour of the SDK.

apps/agentstack-server/src/agentstack_server/configuration.py
```diff
@@ -129,8 +130,19 @@ def validate_auth(self):
...
+            raise ValueError("JWT private and public keys must be provided if authentication is enabled")
+        return self
+
+    @model_validator(mode="after")
+    def set_default_jwt_keys(self):
+        if self.jwt_private_key.get_secret_value() == "dummy" or self.jwt_public_key.get_secret_value() == "dummy":
+            logger.warning("JWT private and public keys are not set. Generating default keys.")
+            from authlib.jose import JsonWebKey
+
+            key = JsonWebKey.generate_key("RSA", 1024, is_private=True)
```
pilartomas: Too small, you should use at least 2048 bits

apps/agentstack-server/src/agentstack_server/configuration.py
```diff
@@ -129,8 +130,19 @@ def validate_auth(self):
...
+            self.jwt_private_key = Secret(key.as_pem(is_private=True).decode("utf-8"))
+            self.jwt_public_key = Secret(key.as_pem(is_private=False).decode("utf-8"))
         return self
```
pilartomas, line 152: Would be nice to verify keypair in the `else` branch. Mismatch will be a common configuration error.

apps/agentstack-server/src/agentstack_server/api/routes/auth.py
```diff
@@ -19,3 +21,10 @@ def protected_resource_metadata(
     resource: str = "",
 ):
     return auth_service.protected_resource_metadata(resource=create_resource_uri(request.url.replace(path=resource)))
+
+
+@well_known_router.get("/jwks")
+def jwks():
+    config = get_configuration()
+    key = JsonWebKey.import_key(config.auth.jwt_public_key.get_secret_value(), {"use": "sig", "alg": "RS256"})
+    return {"keys": [key.as_dict()]}
```
pilartomas, line 30: There is also `KeySet` class in the same module if you'd like to use it to make it nicer.

apps/agentstack-ui/src/modules/runs/contexts/agent-demands/AgentDemandsProvider.tsx
```diff
@@ -115,28 +108,8 @@ export function AgentDemandsProvider<UIGenericPart>({
...
-    if (!contextToken) {
-      throw new Error('Could not generate context token');
-    }
-
-    return contextToken;
-  }, [contextId, contextTokenPermissions, createContextToken]);
-
   const getFulfillments = useCallback(
     async (fulfillmentsContext: FulfillmentsContext) => {
-      const contextToken = await getContextToken();
```
tomkis, line 138: Context token in fullfilments should be marked as deprecated.

apps/agentstack-sdk-ts/src/client/api/build-api-client.ts
jezekra1: sounds good, is it possible to do with the pre-generated schema?

tomkis: We are actually handwriting the zod validation schemas ourselves just beacuse the SDK needs runtime validation (openapi specs is not enough). I believe there's an existing project that would allow generating zod schemas on the fly based on openapi spec, we are not using it.

PR #1789, opened 2026-01-08
feat(ui): gui shell redesign, add agent detail, settings page

### LINKED ISSUE #1565: UI - Left panel scrollability not obvious
## Description

Several users have mentioned that it’s not immediately clear that the left panel is scrollable to view more than 5 agents. Currently, the UI doesn’t provide a visual cue that there are additional agents off-screen.

## Screenshot

Image

## Designs

- Redesigned navigation

## Figma

https://www.figma.com/design/oaJEgk4ZcH3mI1juPGITsx/Agent-Testbed-dev-ready--external-?node-id=1533-8725&p=f&t=c4QShobVWTStNXLI-0

kapetr:
## Summary
Implements latest GUI shell redesign changes including the new agent detail page and the agent settings page.

## Linked Issues
Closes #1565

## Documentation
- [x] No Docs Needed: mostly design changes


apps/agentstack-ui/src/utils/router.ts
```diff
@@ -11,12 +11,19 @@ export const routes = {
   signIn: ({ callbackUrl }: { callbackUrl?: string } = {}) =>
     `/signin${callbackUrl ? `?callbackUrl=${encodeURIComponent(callbackUrl)}` : ''}`,
   notFound: () => '/not-found' as const,
+  agentRoute: ({ providerId }: AgentRouteParams) => `/${encodeURIComponent(providerId)}`,
```
PetrBulanek: Don't we want to add a segment to the URL that says it's an agent URL? Right now, any `/whatever` doesn't end up with a general 404, but with a page saying that the agent wasn't found, which I find a bit confusing.

apps/agentstack-ui/src/utils/router.ts
```diff
@@ -11,12 +11,19 @@ export const routes = {
...
   agentRun: ({ providerId, contextId }: AgentRunParams) =>
-    `/run/${encodeURIComponent(providerId)}${contextId ? `/c/${encodeURIComponent(contextId)}` : ''}`,
-  settings: () => '/settings' as const,
+    `${routes.agentRoute({ providerId })}${contextId ? `/c/${encodeURIComponent(contextId)}` : ''}`,
+  agentSettings: (params: AgentRouteParams) => `${routes.agentRoute(params)}/settings`,
+  agentAbout: (params: AgentRouteParams) => `${routes.agentRoute(params)}/about`,
+  settings: ({ providerId }: Partial<AgentRouteParams> = {}) =>
+    providerId ? `${routes.agentRoute({ providerId })}/global-settings` : '/settings',
```
PetrBulanek, line 20: Is it intentional that these global settings are available at two different URL addresses?

apps/agentstack-ui/src/modules/home/components/HomeHeading.tsx
PetrBulanek, line 1: In the end, it's probably an almost useless component. :-D But it doesn't matter, I would just remove the unnecessary 'use client' directive. :-)

apps/agentstack-ui/src/modules/history/components/SessionsNav.tsx
```diff
@@ -45,14 +45,12 @@ export function SessionsNav({ className }: Props) {
...
 
         const heading = (metadata?.title || created_at) ?? contextId;
-        const subHeading = agent.name;
         const isActive = contextIdUrl === contextId;
 
         return {
           contextId,
           providerId,
           heading,
-          subHeading,
```
PetrBulanek, line 55: We can probably remove this also from the `SessionItem` itself? :)

apps/agentstack-ui/src/modules/agents/components/detail/AgentSection.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface Props {
+  title: string;
+}
+
+export function AgentSection({ title, children }: PropsWithChildren<Props>) {
+  return (
+    <div className={classes.root}>
+      <h2>{title}</h2>
+      {children}
```
PetrBulanek: The children should be wrapped with a div due to styling of the .root as display: flex. Otherwise, it may unexpectedly affect the children elements.

apps/agentstack-ui/src/modules/agents/components/detail/AgentSettingsView.module.scss
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright 2025 © BeeAI a Series of LF Projects, LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+.root {
+  h2 {
+    @include type-style('heading-02');
+  }
```
PetrBulanek: Right now, this is the same as styles for h2 inside AgentSection.module.scss, but if we change one of them later, it might actually conflict unexpectedly. I would try to avoid styling specific tags and rather use a class instead.

apps/agentstack-ui/src/app/(main)/[providerId]/rsc.tsx
```diff
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-'use server';
+'server only';
```
Comment (PetrBulanek) [inline apps/agentstack-ui/src/app/(main)/[providerId]/rsc.tsx]: I am not sure this actually does anything. We probably need this https://nextjs.org/docs/app/guides/data-security#preventing-client-side-execution-of-server-only-code.

Comment (PetrBulanek) [reply apps/agentstack-ui/src/app/(main)/[providerId]/rsc.tsx]: Check also `apps/agentstack-ui/src/app/(auth)/token-endpoint.ts` and `apps/agentstack-ui/src/app/(auth)/rsc.tsx`.

apps/agentstack-ui/src/components/Navbar/FooterNav.module.scss
PetrBulanek, line 1: Seems like those styles are not used anywhere.

apps/agentstack-ui/src/components/Navbar/Navbar.tsx
PetrBulanek, line 1: Opening or closing the Navbar makes the session history flicker, as opposed to other elements that "animate" nicely.

apps/agentstack-ui/src/components/Navbar/Navbar.tsx
```diff
@@ -5,30 +5,34 @@
...
   const [isAnimating, setIsAnimating] = useState(false);
 
-  const {
-    config: { isAuthEnabled },
-    sidebarOpen,
-  } = useApp();
+  const router = useRouter();
+  const { providerId } = useParamsFromUrl();
+
+  const { sidebarOpen } = useApp();
```
PetrBulanek: We could change the name of this variable to match the renaming of Sidebar => Navbar.

apps/agentstack-sdk-py/examples/citation_agent.py
PetrBulanek, line 1: `agentstack-sdk-py:check:ruff-check` fails. There are some linting issues that need to be fixed.

apps/agentstack-sdk-py/examples/secrets_agent.py
PetrBulanek, line 1: Same as `apps/agentstack-sdk-py/examples/citation_agent.py`.

apps/agentstack-ui/src/components/Navbar/NavbarButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  onClick?: () => void;
+  href?: string;
+  icon: ComponentType;
+  hoverIcon?: ComponentType;
+  label: string;
+}
+
+export function NavbarButton({ onClick, href, icon: Icon, hoverIcon: HoverIcon, label }: Props) {
+  return (
+    <Button className={classes.root} kind="ghost" size="sm" onClick={() => onClick?.()} href={href}>
```
PetrBulanek: That anonymous function might not be necessary here.

apps/agentstack-ui/src/modules/agents/components/cards/AgentsListItem.module.scss
```diff
@@ -0,0 +1,29 @@
...
+.link {
+  @include link-mask();
+  @include type-style(heading-03);
+  text-decoration: none;
+  color: inherit;
+  display: flex;
+  align-items: center;
+  gap: $spacing-02;
+  &:focus-visible {
+    outline: none;
```
PetrBulanek: Can we leave the option to focus using the keyboard? I use it quite a lot. Or is there a special reason for this? :-)

apps/agentstack-ui/src/modules/agents/components/detail/AgentDetailView.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <AnimatePresence>
+          <motion.div {...agentDetailFadeProps} key="header">
+            <ViewHeader heading="Agent details" />
+          </motion.div>
+
+          <div className={classes.main}>
+            <div className={classes.mainInfo}>
+              {documentationUrl && (
+                <motion.div {...agentDetailFadeProps} key="documentationLink">
+                  <ExternalLink href={documentationUrl} className={classes.documentationLink}>
```
PetrBulanek, line 45: I'd add some hover style. Maybe underline, or just slight color change ($text-dark).

apps/agentstack-ui/src/modules/agents/components/detail/AgentSecrets.module.scss
```diff
@@ -3,8 +3,46 @@
...
+
+  > div {
+    display: flex;
+    flex-direction: column;
+    gap: $spacing-02;
+
+    :global(.cds--label) {
+      @include type-style(body-02);
+      color: $text-primary;
+      // display: none;
```
PetrBulanek: Probably a leftover from some debugging.

apps/agentstack-ui/src/modules/agents/components/detail/AgentTool.module.scss
```diff
@@ -27,17 +27,18 @@
...
 
 .name {
-  @include type-style('body-compact-01');
+  @include type-style('heading-02');
   @include line-clamp();
   color: $text-dark;
   flex-grow: 1;
 }
 
 .description {
```
PetrBulanek, line 36: There might be some issue with text truncating vs. font-size. I did not investigate farther, but what happens is that the "Show more/less" button shows even if there is actually no more content to show. Check ChatV2 agent and the File Creator tool.

image
image

apps/agentstack-ui/src/utils/router.ts
kapetr: I missed that. Let's use `/agent/...`  then, because the previous `/run/...` would be confusing for the new routes.

apps/agentstack-ui/src/utils/router.ts
kapetr: Yes it is. It comes from the new design - agent page opens in a new window/tab and global settings should be accessible here, but you don't want to lose agent context, when visiting the settings.
You can imagine I'm not happy about this :D But it seems like the cleanest solution given the desired outcome.

apps/agentstack-ui/src/modules/home/components/HomeHeading.tsx
kapetr: Agree, removing.

apps/agentstack-ui/src/modules/history/components/SessionsNav.tsx
kapetr: I left it there on purpose, since we might need it again soon :) But I guess, you're right, let's keep the components minimal. We can always add it back in seconds.

apps/agentstack-ui/src/modules/agents/components/detail/AgentSettingsView.module.scss
kapetr: Leftover of different wip component structure. But good point.

apps/agentstack-ui/src/modules/agents/components/cards/AgentsListItem.module.scss
kapetr: This is just a leftover from copying previous agent list item (`AgentCard`) style, removing.

apps/agentstack-ui/src/modules/agents/components/detail/AgentTool.module.scss
kapetr: Fixed in the `LineClampText` component

PR #1796, opened 2026-01-12
feat(api, sdk): adding 'delete from selected id onward' endpoint and …

jjeliga:
feat(api, sdk): adding 'delete from selected id onward' endpoint and sdk functionality

## Summary
Adding the DELETE history endpoint which deletes all messages from conversation history from the given message onwards (all newer messages) including the message with the given id. The functionality is implemented for the sdk as well. Example usage is documented.

## Linked Issues
Closes https://github.com/i-am-bee/agentstack/issues/1786

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-sdk-py/src/agentstack_sdk/server/context.py
+++ b/apps/agentstack-sdk-py/src/agentstack_sdk/server/context.py
@@ -1,8 +1,8 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-
-from collections.abc import AsyncIterator
+from collections.abc import AsyncGenerator
+from typing import Literal, overload
 from uuid import UUID
 
 import janus
@@ -32,7 +32,19 @@ class RunContext(BaseModel, arbitrary_types_allowed=True):
             data = data.model_copy(deep=True, update={"context_id": self.context_id, "task_id": self.task_id})
         await self._store.store(data)
 
-    async def load_history(self, load_history_items: bool = False) -> AsyncIterator[ContextHistoryItem | Message | Artifact]:
+    @overload
+    async def load_history(
+        self, load_history_items: Literal[False] = False
+    ) -> AsyncGenerator[Message | Artifact, None]:
+        yield ...  # type: ignore
+
+    @overload
+    async def load_history(self, load_history_items: Literal[True]) -> AsyncGenerator[ContextHistoryItem, None]:
+        yield ...  # type: ignore
+
+    async def load_history(
+        self, load_history_items: bool = False
+    ) -> AsyncGenerator[ContextHistoryItem | Message | Artifact]:
         if not self._store:
             raise RuntimeError("Context store is not initialized")
         async for item in self._store.load_history(load_history_items=load_history_items):
--- a/apps/agentstack-sdk-py/src/agentstack_sdk/server/store/memory_context_store.py
+++ b/apps/agentstack-sdk-py/src/agentstack_sdk/server/store/memory_context_store.py
@@ -18,17 +18,17 @@ class MemoryContextStoreInstance(ContextStoreInstance):
         self.context_id = context_id
         self._history: list[ContextHistoryItem] = []
 
-    async def load_history(self, load_history_items: bool = False) -> AsyncIterator[ContextHistoryItem | Message | Artifact]:
-        for message in self._history.copy():
+    async def load_history(
+        self, load_history_items: bool = False
+    ) -> AsyncIterator[ContextHistoryItem | Message | Artifact]:
+        for item in self._history.copy():
             if load_history_items:
-                yield message.model_copy(deep=True)
+                yield item.model_copy(deep=True)
             else:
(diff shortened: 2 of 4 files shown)


docs/development/agent-integration/multi-turn.mdx
```diff
@@ -276,3 +276,46 @@ This advanced example demonstrates several key concepts:
...
+from a2a.types import Message, Part, Role, TextPart
+from agentstack_sdk.platform.context import Context
+from agentstack_sdk.server.context import RunContext
+
+async def edit_message_in_context(run_context: RunContext, message_id: str, new_text: str, metadata: dict[str, Any] | None = None):
+    """Edit a message in the conversation history"""
+    context = await Context.get(run_context.context_id)
+
+    # Step 1: Delete from this message onwards
+    await context.delete_history_from_id(from_id=message_id)
```
jezekra1: This example is a bit convoluted, I wouldn't combine `context` and `run_context` methods when `RunContext` also has a `delete_history_from_id` method

apps/agentstack-sdk-py/src/agentstack_sdk/server/store/platform_context_store.py
```diff
@@ -46,3 +47,7 @@ async def load_history(self) -> AsyncIterator[Message | Artifact]:
     async def store(self, data: Message | Artifact) -> None:
         async with self._platform_extension.use_client():
             await Context.add_history_item(self._context_id, data=data)
+
+    async def delete_history_from_id(self, from_id: UUID) -> None:
+        async with self._platform_extension.use_client():
+            await Context.delete_history_from_id(self._context_id, from_id=from_id)
```
jezekra1, line 58: This is incosinstent with `MemoryContextStore` which takes a "message id" as `from_id`, but here you expect "history item id" instead  breaking the polymorphism.

This means that switching from
`store=MemoryContextStore()` to `store=PlatformContextStore()` would break the agent

There are 3 possible solutions to this:
- Expose a context item wrapper - e.g. `HistoryItem` (`id: UUID`, data: `Message | Artifact`) to the agent developers (output of `RunContext.load_data` would be `HistoryItem` instead of directly `Message | Artifact`,  then this method would take item id, not message_id. This would be backwards incompatible - need to update all agents that work with history (I'd say not worth it for this niche feature)
- Allow deleting also by a "data query" - add `from_data_id` in addition from `from_id`, this would be a json query in postgres trying to find the message by id in the `data` column. The SDK method would allow both `Context.delete_history_from_id(from_id=..., from_data_id=...)` (mutually exclusive)
- `PlatformContextStore.delete_history_from_id` would accept "message_id" and then it would load all history items from the platform and do the mapping "message_id" -> "item_id" locally and send `from_id=item_id` to the platform api

apps/agentstack-sdk-py/src/agentstack_sdk/server/context.py
```diff
@@ -30,12 +32,17 @@ async def store(self, data: Message | Artifact):
             data = data.model_copy(deep=True, update={"context_id": self.context_id, "task_id": self.task_id})
         await self._store.store(data)
 
-    async def load_history(self) -> AsyncIterator[Message | Artifact]:
+    async def load_history(self, load_history_items: bool = False) -> AsyncIterator[ContextHistoryItem | Message | Artifact]:
```
jezekra1: create type `@overload` for this method, to properly advertise the response type

```
@overload
def load_hostory(self, load_history_items: False) -> AsyncIterator[Message | Artifact]: ...

@overload
def load_hostory(self, load_history_items: True) -> AsyncIterator[ContextHistoryItem]: ...
```
something like this should be possible

apps/agentstack-sdk-py/src/agentstack_sdk/server/store/memory_context_store.py
```diff
@@ -3,25 +3,41 @@
...
 
 class MemoryContextStoreInstance(ContextStoreInstance):
     def __init__(self, context_id: str):
         self.context_id = context_id
-        self._history: list[Message | Artifact] = []
+        self._history: list[ContextHistoryItem] = []
 
-    async def load_history(self) -> AsyncIterator[Message | Artifact]:
+    async def load_history(self, load_history_items: bool = False) -> AsyncIterator[ContextHistoryItem | Message | Artifact]:
         for message in self._history.copy():
```
jezekra1: nitpick: let's call it `item`

docs/development/agent-integration/multi-turn.mdx
```diff
@@ -276,3 +276,46 @@ This advanced example demonstrates several key concepts:
...
+```python
+import uuid
+from uuid import UUID
+from typing import Any
+
+from a2a.types import Message, Part, Role, TextPart
+from agentstack_sdk.platform.context import Context
+from agentstack_sdk.server.context import RunContext
+
+async def edit_message_in_context(run_context: RunContext, id: UUID, new_text: str, metadata: dict[str, Any] | None = None):
```
jezekra1, line 298: maybe we should mention that you need to use the ID from items (`load_history_item=True`) and not `message_id`

PR #1815, opened 2026-01-14
chore(ui): add css-modules eslint plugin

kapetr:
## Summary
Adds css-modules eslint plugin and fixes css classes issues. 

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/components/SidePanel/SidePanel.module.scss
+++ b/apps/agentstack-ui/src/components/SidePanel/SidePanel.module.scss
@@ -12,11 +12,11 @@ $panel-width: rem(432px);
   z-index: 1;
   inline-size: 0;
   overflow: hidden;
-  transition-timing-function: motion(entrance, expressive);
+  transition: inline-size $duration-moderate-02 motion(entrance, expressive);
 
   &.isOpen {
     inline-size: $panel-width;
-    transition: inline-size $duration-moderate-02 motion(exit, expressive);
+    transition-timing-function: motion(exit, expressive);
   }
 }
 
--- a/apps/agentstack-ui/src/modules/runs/components/ChatMarkdownContent/CitationLink/InlineCitations.module.scss
+++ b/apps/agentstack-ui/src/modules/runs/components/ChatMarkdownContent/CitationLink/InlineCitations.module.scss
@@ -17,10 +17,8 @@
 
   &.isActive,
   &:hover {
-    .root:has(&) {
-      .content {
-        text-decoration: underline;
-      }
+    .root:has(&) .content {
+      text-decoration: underline;
     }
   }
 }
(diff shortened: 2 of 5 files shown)


apps/lint-config/eslint.config.js
```diff
@@ -43,6 +43,13 @@ const nextConfig = [
   ...base,
   ...compat.extends('next/core-web-vitals', 'next/typescript', 'prettier'),
   ...compat.plugins('simple-import-sort'),
+  ...compat.config({
+    plugins: ['css-modules'],
+    rules: {
+      'css-modules/no-unused-class': 'off',
```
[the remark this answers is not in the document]
kapetr: It's turned off by design, because we have quite a few components with classes defined via variable, which the plugin can't handle.

apps/agentstack-ui/src/components/SidePanel/SidePanel.module.scss
```diff
@@ -10,13 +10,13 @@ $panel-width: rem(432px);
...
   overflow: hidden;
-  transition: inline-size $duration-moderate-02 motion(exit, expressive);
+  transition: inline-size $duration-moderate-02 motion(entrance, expressive);
 
-  &:not(.isOpen) {
-    inline-size: 0;
-    transition-timing-function: motion(entrance, expressive);
+  &.isOpen {
+    inline-size: $panel-width;
+    transition: motion(exit, expressive);
```
PetrBulanek: If I'm not mistaken, this will now overwrite the transition property and duration, so the animation won't work anymore.

apps/agentstack-ui/src/modules/agents/components/detail/AgentTool.module.scss
```diff
@@ -17,15 +17,6 @@
...
 }
 
-.icon {
-  inline-size: rem(24px);
-  block-size: rem(24px);
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  flex-shrink: 0;
-}
```
PetrBulanek, line 27: These styles were preparation for the icon used in the AgentTool component. The markup is still there (but commented out). I think we can remove them as well.

apps/agentstack-ui/src/modules/runs/components/ChatMarkdownContent/CitationLink/InlineCitations.module.scss
```diff
@@ -21,4 +14,13 @@
 
 .item {
   display: flex;
+
+  &.isActive,
+  &:hover {
+    .root:has(&) {
+      .content {
```
PetrBulanek: I think we can simplify it to one line: `.root:has(&) .content`.

PetrBulanek (changes requested):

PR #1842, opened 2026-01-19
chore: TS SDK experimental server

tomkis:
## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-sdk-ts/package.json
+++ b/apps/agentstack-sdk-ts/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "build": "tsup",
     "dev": "tsup --watch",
-    "example:hello-world": "tsx watch src/examples/hello-world.ts",
+    "experimental:example:hello-world": "tsx watch src/examples/hello-world.ts",
     "test": "vitest run",
     "test:watch": "vitest"
   },
--- a/apps/agentstack-sdk-ts/src/experimental/server/core/server/index.ts
+++ b/apps/agentstack-sdk-ts/src/experimental/server/core/server/index.ts
@@ -41,7 +41,7 @@ export class Server {
       description: options.description ?? '',
       url: 'http://localhost:8000',
       version: options.version ?? '1.0.0',
-      protocolVersion: '0.2.2',
+      protocolVersion: options.protocolVersion ?? '0.3.0',
       capabilities: {
         streaming: true,
         extensions,
--- a/apps/agentstack-sdk-ts/src/experimental/server/core/server/types.ts
+++ b/apps/agentstack-sdk-ts/src/experimental/server/core/server/types.ts
@@ -32,6 +32,7 @@ export interface AgentOptions<TDeps = Record<string, never>> {
   detail?: AgentDetail;
   extensions?: ExtensionConfig<TDeps>;
   version?: string;
+  protocolVersion?: string;
   handler: AgentFunction<TDeps>;
 }
 


apps/agentstack-sdk-ts/package.json
```diff
@@ -4,8 +4,9 @@
   "version": "0.5.3",
   "type": "module",
   "scripts": {
-    "build": "rimraf dist && microbundle --format modern,cjs,umd src/*.ts",
-    "dev": "microbundle watch src/*.ts"
+    "build": "rimraf dist && microbundle --format modern,cjs,umd src/*.ts src/experimental/server.ts",
+    "dev": "microbundle watch src/*.ts src/experimental/server.ts",
```
tomkis: Let's keep this as it was. Let's make sure server has it's own file in the root instead. Ensure consistency - server.ts in the src folder would export just stuff that is now being exported from src/experimental/server/index.ts and the intermediate src/experimental/server.ts is no longner needed.

tomkis: Same goes for the build script.

apps/agentstack-sdk-ts/src/examples/hello-world.ts
```diff
@@ -0,0 +1,29 @@
...
+
+server
+  .agent({
+    name: 'HelloWorld',
+    description: 'A simple hello world agent',
+    version: '0.0.1',
+    detail: {
+      interaction_mode: 'multi-turn',
+      user_greeting: 'Hello! How can I help you?',
+      author: { name: 'BeeAI' },
```
tomkis: Auth should be Agentstack.

apps/agentstack-sdk-ts/src/examples/hello-world.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const server = new Server();
+
+server
+  .agent({
+    name: 'HelloWorld',
+    description: 'A simple hello world agent',
+    version: '0.0.1',
+    detail: {
+      interaction_mode: 'multi-turn',
```
tomkis: Instead of using string, we should use type from the SDK.

apps/agentstack-sdk-ts/src/experimental/server.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
tomkis, line 1: As mentioned previously. This file is not needed.

apps/agentstack-sdk-ts/src/experimental/server/autoregistration.ts
```diff
@@ -0,0 +1,181 @@
...
+
+export interface AutoRegistrationOptions {
+  platformUrl: string;
+  selfRegistrationId: string;
+  agentCard: AgentCard;
+  host: string;
+  port: number;
+}
+
+export class AutoRegistration {
```
tomkis: This class is doing a bit more than it should.

1. We should create configuration service that would cover all .env access - zod schema definition what to expect in  .env and then runtime parsing this - as completely separate concern of this. This class would only depend on the config service.
2. API client should be passed as dependency so that we don't have multiple instances of the API client scattered throught the code base. 
3. both sleep and withRetry should not be concern of this class but rather use some helper / library that provides this exponential backoff approach.
4. Another thing is that i don't like this to be a class - AutoRegistration does not sound right  neither class design seems correct. Let's use factory with scoped state and call this `createAutoregisterToAgentstack` - which is a functio nto be called which returns a void function that does the cleanup.

apps/agentstack-sdk-ts/src/experimental/server/autoregistration.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    try {
+      const result = await this.api.listProviderVariables({ id: this.providerId });
+
+      if (!result.ok) {
+        console.log(`Failed to load variables: ${result.error.message}`);
+        return;
+      }
+
+      const variables = result.data.variables ?? {};
```
tomkis: The whole variables resultion as well as dirty checkup is really complex and error-prone. Can't we use some lib (lodash or something similar) to simplify?

apps/agentstack-sdk-ts/src/experimental/server/autoregistration.ts
```diff
@@ -0,0 +1,181 @@
...
+      if (firstRun && this.configuredVariables.size > 0) {
+        console.log('Environment variables loaded dynamically.');
+      }
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      console.log(`Failed to load variables: ${message}`);
+    }
+  }
+
+  startVariableReload(): void {
```
tomkis: I'd prefer this to be coupled with registration to make the interface simpler and more transactional.

apps/agentstack-sdk-ts/src/experimental/server/executor.ts
```diff
@@ -0,0 +1,220 @@
...
+}
+
+interface RequestContext {
+  readonly userMessage: Message;
+  readonly task?: Task;
+  readonly taskId: string;
+  readonly contextId: string;
+}
+
+interface AgentExecutor {
```
tomkis: There is no need for this interface.

apps/agentstack-sdk-ts/src/experimental/server/types.ts
```diff
@@ -0,0 +1,43 @@
...
+  detail?: AgentDetail;
+  extensions?: ExtensionConfig<TDeps>;
+  handler: AgentFunction<TDeps>;
+  version?: string;
+}
+
+export interface ServerOptions {
+  host?: string;
+  port?: number;
+  selfRegistration?: boolean;
```
tomkis: No need to have both selfRegistration and selfRegistrationId - lets just make it that if the selfRegistrationId is omitted then dont do self reg. In otherwords make the selfRegistrationId mandatory for self registration.

apps/agentstack-sdk-ts/src/experimental/server/server.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const taskStore = new InMemoryTaskStore();
+    const executor = new AgentExecutorImpl(
+      this.agentOptions.handler,
+      this.agentOptions.extensions as ExtensionConfig<unknown> | undefined,
+    );
+
+    const requestHandler = new DefaultRequestHandler(this.agentCard, taskStore, executor);
+
+    const express = await import('express');
```
tomkis: Import this statically.

apps/agentstack-sdk-ts/package.json
```diff
@@ -4,8 +4,11 @@
   "version": "0.5.3",
   "type": "module",
   "scripts": {
-    "build": "rimraf dist && microbundle --format modern,cjs,umd src/*.ts",
-    "dev": "microbundle watch src/*.ts"
+    "build": "tsup",
+    "dev": "tsup --watch",
+    "example:hello-world": "tsx watch src/examples/hello-world.ts",
```
PetrBulanek: Do we want to have this in package.json as it's part of an experimental feature?

apps/agentstack-sdk-ts/src/experimental/server/core/server/index.ts
```diff
@@ -0,0 +1,156 @@
...
+      const detailExt = agentDetailExtension(options.detail);
+      extensions = [...(extensions ?? []), detailExt.toAgentCardExtension()];
+    }
+
+    this.agentCard = {
+      name: options.name,
+      description: options.description ?? '',
+      url: 'http://localhost:8000',
+      version: options.version ?? '1.0.0',
+      protocolVersion: '0.2.2',
```
PetrBulanek: Is this protocol version correct?

apps/agentstack-sdk-ts/package.json
tomkis: Good point, let's prefix this with experimental

apps/agentstack-sdk-ts/src/experimental/server/core/server/index.ts
kapetr: I think this might be wrong. We don't set protocolVersion in python counterpart and it's odd being hardcoded without possibility to overwrite it. If you leave it out, current version is set by default - `0.3.0`.

PR #1843, opened 2026-01-19
chore(ui): import some libs dynamically, fix file artifact triggering canvas ui

kapetr:
## Summary
Dynamically imported syntax-highlighter and mermaid libraries to reduce the initial bundle size and improve page load performance. Fixed an issue where file artifacts were incorrectly triggering canvas rendering.

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
+++ b/apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
@@ -9,6 +9,7 @@ import type { ExtraProps } from 'react-markdown';
 
 import { useTheme } from '#contexts/Theme/index.ts';
 import { Theme } from '#contexts/Theme/types.ts';
+import { usePrevious } from '#hooks/usePrevious.ts';
 
 import { useMermaid } from '../contexts';
 import { Code } from './Code';
@@ -29,8 +30,10 @@ export function MermaidDiagram({ children, mermaidIndex, isStreaming }: MermaidD
 
   const diagram = diagrams.get(index);
 
+  const themePrevious = usePrevious(theme);
   useEffect(() => {
     let isMounted = true;
+    const hasThemeChanged = themePrevious && theme !== themePrevious;
 
     async function renderDiagram() {
       if (typeof children !== 'string') {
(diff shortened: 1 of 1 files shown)


apps/agentstack-ui/src/components/SyntaxHighlighter/languages.ts
```diff
@@ -3,17 +3,27 @@
...
-import bash from 'react-syntax-highlighter/dist/esm/languages/hljs/bash';
-import javascript from 'react-syntax-highlighter/dist/esm/languages/hljs/javascript';
-import json from 'react-syntax-highlighter/dist/esm/languages/hljs/json';
-import python from 'react-syntax-highlighter/dist/esm/languages/hljs/python';
-import shell from 'react-syntax-highlighter/dist/esm/languages/hljs/shell';
-import typescript from 'react-syntax-highlighter/dist/esm/languages/hljs/typescript';
-import yaml from 'react-syntax-highlighter/dist/esm/languages/hljs/yaml';
+type RegisterableHighlighter = { registerLanguage(name: string, func: unknown): void };
+
+export async function registerLanguagesAsync(highlighter: RegisterableHighlighter) {
```
PetrBulanek: What about this typing?

```
import type { Light as Highlighter } from 'react-syntax-highlighter';

registerLanguagesAsync(highlighter: typeof Highlighter)
```

apps/agentstack-ui/src/components/SyntaxHighlighter/SyntaxHighlighterImpl.tsx
```diff
@@ -0,0 +1,38 @@
+'use client';
```
PetrBulanek: I think we should move this below the copyright.

apps/agentstack-ui/src/components/SyntaxHighlighter/SyntaxHighlighter.tsx
```diff
@@ -1,14 +1,11 @@
+'use client';
```
PetrBulanek: I think we should move this below the copyright.

apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
PetrBulanek, line 1: Due to this change, reinitialization will not work when switching to dark/light theme. Do you think it's necessary, or should we ignore it?

PetrBulanek (approved):

apps/agentstack-ui/src/components/MarkdownContent/components/MermaidDiagram.tsx
kapetr: Good catch! 👍 It's an edge case, but a fix should be easy.

PetrBulanek (approved):

PR #1857, opened 2026-01-20
feat(sdk, cli): add support for connectors

jjeliga:
## Summary
Adds the connector management functionality to python sdk and CLI. Mainly functions as the API wrapper but introduces some additional handy functionality as well.

## Linked Issues
Closes https://github.com/i-am-bee/agentstack/issues/1713

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-cli/src/agentstack_cli/commands/connector.py
+++ b/apps/agentstack-cli/src/agentstack_cli/commands/connector.py
@@ -1,21 +1,19 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 import asyncio
-import json
 import typing
-from uuid import UUID
 
 import pydantic
 import typer
-from agentstack_sdk.platform.client import PlatformClient
 from agentstack_sdk.platform.connector import Connector, ConnectorState
 from agentstack_sdk.platform.types import Metadata
 from InquirerPy import inquirer
 from InquirerPy.base.control import Choice
 
 from agentstack_cli import configuration
-from agentstack_cli.async_typer import AsyncTyper, console
+from agentstack_cli.async_typer import AsyncTyper
 from agentstack_cli.configuration import Configuration
+from agentstack_cli.console import console
 from agentstack_cli.utils import (
     announce_server_action,
     confirm_server_action,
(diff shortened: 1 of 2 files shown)


apps/agentstack-server/tests/e2e/sdk/test_connectors_sdk.py
```diff
@@ -0,0 +1,220 @@
+# Copyright 2025 © BeeAI a Series of LF Projects, LLC
```
pilartomas, line 1: Doesn't this suite belong to `apps/agentstack-sdk-py/tests`, rather?

pilartomas (approved):
Looks great 👍

apps/agentstack-server/tests/e2e/sdk/test_connectors_sdk.py
jjeliga: Thew e2e test should rather  be moved outside of the server app, since they use sdk in multiple places when testing agents. Decided to keep it as is for now

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
```diff
@@ -0,0 +1,307 @@
...
+app = AsyncTyper()
+config = Configuration()
+
+
+@app.command("create")
+async def create_connector(
+    url: typing.Annotated[str, typer.Argument(help="Agent location (public docker image or github url)")],
+    client_id: typing.Annotated[
+        str | None,
+        typer.Option("--client-id", "-cid", help="Client ID for authentication, acquired form env if not supplied"),
```
JanPokorny: Short options are supposed to be single character (so you can combine them like `ls -la`), but I wouldn't include them at all for non-obvious ones like these.

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
```diff
@@ -0,0 +1,307 @@
...
+        console.print(f"[red]Error parsing metadata JSON:[/red] {e}")
+        raise typer.Exit(code=1) from None
+    async with configuration.use_platform_client() as client:
+        connector = await Connector.create(
+            url,
+            client_id=client_id if client_id else config.client_id,
+            client_secret=client_secret if client_secret else config.client_secret,
+            metadata=pydantic.TypeAdapter(Metadata).validate_python(metadata_dict),
+            match_preset=match_preset,
+            client=client,
```
JanPokorny: If you're inside `use_platform_client`, it's not necessary to pass `client=client` as it's managed internally. (Applies to more cases below)

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
```diff
@@ -0,0 +1,307 @@
...
+        metadata_dict = json.loads(metadata) if isinstance(metadata, str) else {}
+    except json.JSONDecodeError as e:
+        console.print(f"[red]Error parsing metadata JSON:[/red] {e}")
+        raise typer.Exit(code=1) from None
+    async with configuration.use_platform_client() as client:
+        connector = await Connector.create(
+            url,
+            client_id=client_id if client_id else config.client_id,
+            client_secret=client_secret if client_secret else config.client_secret,
+            metadata=pydantic.TypeAdapter(Metadata).validate_python(metadata_dict),
```
JanPokorny: You can do `json.loads` + `TypeAdapter(...).validate_python` in one step with `TypeAdapter(...).validate_json`

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
```diff
@@ -0,0 +1,307 @@
...
+            match_preset=match_preset,
+            client=client,
+        )
+        console.print(f"Created connector for URL [blue]{connector.url}[/blue] with id: [green]{connector.id}[/green]")
```
JanPokorny: You can use these functions for a more unified look: https://github.com/i-am-bee/agentstack/blob/a10cd2087c82e1f3f3d5ab0aaf864610a2864e31/apps/agentstack-cli/src/agentstack_cli/console.py#L7-L21

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
```diff
@@ -0,0 +1,307 @@
...
+        raise ValueError(f"No matching connectors found for '{search_path}'")
+
+    if len(connector_candidates) == 1:
+        return list(connector_candidates.values())
+
+    # Multiple matches - show selection menu
+    choices = [Choice(value=c.id, name=f"{c.url} - {c.id} ({c.state})") for c in connector_candidates.values()]
+
+    selected_ids = await inquirer.checkbox(  # pyright: ignore[reportPrivateImportUsage]
+        message="Select connectors to remove (use ↑/↓ to navigate, Space to select):", choices=choices
```
JanPokorny: This says "remove" but you reuse it for "disconnect" too

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
```diff
@@ -0,0 +1,307 @@
...
+            client_secret=client_secret if client_secret else config.client_secret,
+            metadata=pydantic.TypeAdapter(Metadata).validate_python(metadata_dict),
+            match_preset=match_preset,
+            client=client,
+        )
+        console.print(f"Created connector for URL [blue]{connector.url}[/blue] with id: [green]{connector.id}[/green]")
+        console.print(f"Connector status: [yellow]{connector.state}[/yellow]")
+
+
+def search_path_match_connectors(search_path: str, connectors: list[Connector]) -> dict[UUID, Connector]:
```
JanPokorny: In all cases you iterate over the result of this so it can just return a list / iterator, no need for a dict

apps/agentstack-cli/src/agentstack_cli/commands/connector.py
jjeliga: I think I actually have to be explicit, because the definition of the method (and similarly for other methods as well) is:
```
    @staticmethod
    async def create(
        url: AnyUrl | str,
        *,
        client_id: str | None = None,
        client_secret: str | None = None,
        metadata: Metadata | None = None,
        match_preset: bool = True,
        client: PlatformClient | None = None,
    ) -> Connector:
```

PR #1871, opened 2026-01-22
chore(server): avoid wasteful probes for disconnected connectors

pilartomas:
## Summary

Disconnected connectors may now be disconnected permanently.

No resurrection attempts will be made for permanently disconnected connectors upon refresh. This primarily avoids wasting resources for connectors failing with 4xx or manually disconnected ones.

Additionally, it fixes missing `disconnect_reason` data in the database.

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-server/src/agentstack_server/domain/models/connector.py
+++ b/apps/agentstack-server/src/agentstack_server/domain/models/connector.py
@@ -4,7 +4,6 @@
 from enum import StrEnum
 from typing import Annotated, Literal
 from uuid import UUID, uuid4
-from xmlrpc.client import boolean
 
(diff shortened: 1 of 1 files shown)


apps/agentstack-server/src/agentstack_server/domain/models/connector.py
```diff
@@ -4,6 +4,7 @@
 from enum import StrEnum
 from typing import Annotated, Literal
 from uuid import UUID, uuid4
+from xmlrpc.client import boolean
```
jjeliga: Isn't this auto-generated import after specifying `disconnect_permanent: boolean` instead of `disconnect_permanent: bool`? Why use this type if not?

PR #1872, opened 2026-01-22
fix(cli): improved flow when refresh token expires

jjeliga:
## Summary
When switching servers the app first tries to refresh the token and proceeds to login when it fails due to expired refresh token. The user is notified throughout the process
## Linked Issues
CLoses https://github.com/i-am-bee/agentstack/issues/1833

## Documentation
- [x] No Docs Needed: does not change the functionality in any way


apps/agentstack-cli/src/agentstack_cli/utils.py
```diff
@@ -340,3 +307,22 @@ def is_github_url(url: str) -> bool:
...
+
+
+def get_httpx_response_error_details(response: httpx.Response | None) -> tuple[str, str] | None:
+    if response:
+        try:
+            error_json = response.json()
+            return (
+                f"error: {error_json.get('error', 'unknown')}",
+                f"error description: {error_json.get('error_description', 'No description')}",
+            )
```
pilartomas, line 320: You might want to check `authlib`, it's pretty modular and often there are data types, parser and utils for these cases.

apps/agentstack-cli/src/agentstack_cli/configuration.py
```diff
@@ -39,7 +39,8 @@ class Configuration(pydantic_settings.BaseSettings):
     server_metadata_ttl: int = 86400
 
     oidc_enabled: bool = False
-    client_id: str | None = None
+    # client_id: str | None = None
+    client_id: str = "df82a687-d647-4247-838b-7080d7d83f6c"
```
pilartomas: The public client shouldn't be included as a default, we do not ship a "default issuer" counterpart with the server.

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -112,10 +122,16 @@ async def exchange_refresh_token(self, auth_server: str, token: AuthToken) -> di
                 )
                 resp.raise_for_status()
                 new_token = resp.json()
+            except httpx.HTTPStatusError as e:
+                # Any HTTP error from the token endpoint means the refresh failed
+                # This could be expired/invalid refresh token, revoked token, etc.
+                raise TokenRefreshError("Token refresh failed") from e
```
pilartomas: Or the server might just return 5xx etc.

I'd suggest to only raise `TokenRefreshError` in 4xx range, `RuntimeError` otherwise.

apps/agentstack-cli/src/agentstack_cli/configuration.py
jjeliga: good catch! forgot to remove that

apps/agentstack-cli/src/agentstack_cli/auth_manager.py
```diff
@@ -6,9 +6,13 @@
 from collections import defaultdict
 from typing import Any
 
-import httpx
+from authlib.common.errors import AuthlibBaseError
+from authlib.integrations.httpx_client import AsyncOAuth2Client
+from authlib.oauth2.rfc6749.errors import InvalidGrantError, OAuth2Error
 from pydantic import BaseModel, Field
 
+TOKEN_EXPIRY_BUFFER = 60  # seconds
```
pilartomas: Common name is `leeway`

PR #1878, opened 2026-01-22
fix(ui): getting AgentCard without token causing 401 error

PetrBulanek:
## Summary
Fixes error 401 caused by attempting to fetch an agent card without a token.

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx
+++ b/apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx
@@ -8,19 +8,17 @@ import type { ReactElement } from 'react';
 
 import { handleApiError } from '#app/(auth)/rsc.tsx';
 import { ErrorPage } from '#components/ErrorPage/ErrorPage.tsx';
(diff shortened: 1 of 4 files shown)


apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx
```diff
@@ -6,22 +6,21 @@
...
+import { readProvider } from '#modules/providers/api/index.ts';
 import { ModelType, NoModelSelectedErrorPage } from '#modules/runs/components/NoModelSelectedErrorPage.tsx';
 
 export async function ensureModelSelected(providerId: string) {
   let ErrorComponent: ReactElement | null = null;
 
   try {
-    const client = await getAgentClient(providerId);
-    const card = await client.getAgentCard();
+    const { agent_card } = await readProvider({ id: providerId });
```
Comment (kapetr) [inline apps/agentstack-ui/src/app/(main)/agent/[providerId]/ensure-model-selected.tsx]: I think we could pass the `agent_card` to this function as parameter, because it's fetched in both pages where this is used. Or use `unstable_cache`, so it's not fetched twice.

PR #1879, opened 2026-01-22
feat(infra): add keycloak as a single central OIDC

jezekra1:
## Summary
Adds keycloak as a central OIDC/auth provider:

**Helm changes**
- add keycloak when auth.enabled = true
- add temporary keycloak carbon login theme (will be replaced by a better one in a separate PR)
- simplify auth configuration - it is now on or off, everything is configured automatically (except external domains)

**Server changes**
- remove user management API (users should be managed in keycloak now), possibly reintroduce later through keycloak api
- integrate with keycloak as single oidc provider
- remove old "admin basic auth", replace with proper basic auth through keycloak

**CLI changes**
- remove user management commands
- change admin password env to basic auth env
- [ ] TBD: check token refreshing

**UI changes**
- remove login screen, instead redirect directly to keycloak
- change providers.json variables to a single provider configured by OIDC_* variables
- federate sing out action to keycloak (logout SSO session)

## Documentation
- [ ] No Docs Needed:


apps/agentstack-ui/src/app/(auth)/auth.ts
```diff
@@ -29,24 +27,23 @@ export function getAuthProviders(): AuthProvider[] {
 }
 
 function createOIDCProvider(config: ProviderConfig): OIDCConfig<unknown> {
-  const baseOptions = {
+  const options = {
     clientId: config.client_id,
     clientSecret: config.client_secret,
-    issuer: config.issuer,
+    issuer: config.external_issuer ?? config.issuer,
+    ...(config.external_issuer
```
pilartomas, line 34: Is the external issuer a backwards compatibility thing? Much more logic could be simplified if we just assumed Keycloak as the issuer.

jezekra1: No, this is a hack for localhost:
- nextjs server can access keycloak at http://keycloak:8336 (internal issuer url)
- browser can access keycloak at http://localhost:8336 (external issuer url)

In prod these two URLs should ideally by the same pointing to some https keycloak deployment, I've added warnings everywhere :D

PR #1889, opened 2026-01-23
docs: add external link validation manual tasks

### LINKED ISSUE #1926: Tooling to test for bad external links in docs
We are already using mintlify for the internal doc links.

We need to test external links to avoid document links that result in 404s.  And nice-to-have a way to find/fix permanent redirects.

This can be just a convenience tool for now (mise run).  There are a bunch of known flaky redirects etc that would need some working through before adding this to CI.

markstur:
## Summary

docs: add convenience tool for checking external links
Manual tool only for now. There are some flakey redirects and auth things that we probably
do not want in CI. We can just use this to do some doc cleanup for now.

## Linked Issues
Fixes #1926 

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/docs/tasks.toml
+++ b/docs/tasks.toml
@@ -59,3 +59,11 @@ dir = "{{config_root}}/docs"
 run = "pnpm mintlify broken-links"
 sources = ["**/*.mdx"]
 outputs = { auto = true }
+
+## manual check
+
+["docs:manual-check:external-links"]
+dir = "{{config_root}}/docs"
+sources = ["**/*.md", "**/*.mdx", "!node_modules/**", "lychee.toml"]
+outputs = { auto=true }
+run = "lychee --config lychee.toml '**/*.md' '**/*.mdx'"
--- a/docs/lychee.toml
+++ b/docs/lychee.toml
@@ -1,0 +1,42 @@
+# Lychee configuration for link checking
+# See: https://lychee.cli.rs/#/usage/config
+
+# Base URL or website root directory to check relative URLs.
+# We skip these (use mintlify), but need this to pass URL parsing.
+base_url = "https://localhost"
+
+# Uncomment to use verbose info to see redirects
+# verbose = "info"
+
+# Exclude patterns
+exclude = [
+  # Local development URLs
+  '^https?://localhost',
+  '^https?://127\.0\.0\.1',
+  # If we don't skip npmjs then we need to accept 403
+  '^https://www\.npmjs\.com',
+  # Example domains
+  '^https://your-public-url\.',
+  '^https://api\.agentstack\.example\.com',
+  '^https://api\.myapp\.com',
+  '^https://keycloak\.myapp\.com',
+]
+
+# GitHub API token to avoid github warning
+github_token = "secret"
+
+# Exclude paths
+exclude_path = ["node_modules"]
+
+# Accept status codes
+accept = [200, 206, 302, 307]
+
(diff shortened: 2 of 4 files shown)


docs/package.json
```diff
@@ -5,13 +5,16 @@
   "main": "index.js",
   "scripts": {
     "dev": "mintlify dev --port 3333",
-    "update-openapi": "wget localhost:8333/api/v1/openapi.json -O api-reference/openapi.json"
+    "update-openapi": "wget localhost:8333/api/v1/openapi.json -O api-reference/openapi.json",
+    "check:links": "mintlify broken-links",
+    "check:links:external": "find . -type f \\( -name '*.md' -o -name '*.mdx' \\) -not -path './node_modules/*' -print0 | xargs -0 pnpm exec markdown-link-check --config .markdown-link-check.json --quiet"
```
JanPokorny: `check:links` is not needed, and I'd prefer `check:links:external` to be inlined to `tasks.toml` since it's not really a npm task anyway

docs/tasks.toml
```diff
@@ -10,6 +10,13 @@ run = "pnpm dev"
 ["docs:check"]
 depends = ["docs:check:*"]
 
+["docs:manual:check:links:external"]
```
JanPokorny: I'd prefer `docs:manual-check:external-links` and move it to a new `## manual check` section to the bottom. It's just a bit more consistent that way.

docs/.markdown-link-check.json
```diff
@@ -0,0 +1,39 @@
...
+    {
+      "pattern": "^http://localhost"
+    },
+    {
+      "pattern": "^https://localhost"
+    },
+    {
+      "pattern": "^http://127.0.0.1"
+    }
+  ],
```
JanPokorny:
```suggestion
  "ignorePatterns": [
    {
      "pattern": "^(\\.?\\.?/|#)"
    }
  ]
```

docs/.markdown-link-check.json
```diff
@@ -0,0 +1,39 @@
...
+  "replacementPatterns": [],
+  "httpHeaders": [
+    {
+      "urls": ["https://github.com"],
+      "headers": {
+        "Accept-Encoding": "zstd, br, gzip, deflate"
+      }
```
JanPokorny: Just curious, why is this needed?

docs/.markdown-link-check.json
```diff
@@ -0,0 +1,39 @@
...
+    }
+  ],
+  "timeout": "20s",
+  "retryOn429": true,
+  "retryCount": 3,
+  "fallbackRetryDelay": "30s",
+  "aliveStatusCodes": [200, 206, 301, 302, 307, 308, 403]
```
JanPokorny: Why 403?

docs/tasks.toml
markstur: I was really going to ask for naming advice, but decided it was better not to explicitly ask... I do appreciate the review and suggestion.

docs/.markdown-link-check.json
markstur: Coding assistant recommended but I will remove.  I don't see how we would need this in our docs.

docs/.markdown-link-check.json
markstur: 403 is technically alive, but for our purposes I think I should remove it so we fail on forbidden links.

PR #1895, opened 2026-01-26
feat(server): add user feedback as phoenix annotation

pilartomas:
## Summary

Uploads user feedback also as Phoenix annotation, if configured. Server env vars:

```
TELEMETRY__PHOENIX_URL=...
TELEMETRY__PHOENIX_API_KEY=...
```

Helm chart has been extended to support external phoenix services:

```
phoenix:
  enabled: false

externalPhoenix:
  url: ""
  apiKey: ""
```

In either case (when phoenix is enabled or external phoenix is used), it sets the aforementioned server envs and appropriate collector pipelines.

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-server/src/agentstack_server/service_layer/services/user_feedback.py
+++ b/apps/agentstack-server/src/agentstack_server/service_layer/services/user_feedback.py
@@ -1,11 +1,11 @@
 # Copyright 2025 © BeeAI a Series of LF Projects, LLC
 # SPDX-License-Identifier: Apache-2.0
 
-import asyncio
 import logging
 from uuid import UUID
 
 import httpx
+from fastapi import BackgroundTasks
 from kink import inject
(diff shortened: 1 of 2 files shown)


apps/agentstack-server/src/agentstack_server/service_layer/services/user_feedback.py
```diff
@@ -51,6 +73,7 @@ async def create_user_feedback(
             )
             await uow.user_feedback.create(user_feedback=user_feedback)
             await uow.commit()
+            asyncio.create_task(self._try_send_to_phoenix(user_feedback=user_feedback))  # noqa: RUF006
             return user_feedback
```
pilartomas, line 78: @jezekra1 better idea? Is there some pool for background tasks?

Issue #1874, opened 2026-01-22
Canvas text selection fails in Firefox when selecting right-to-left


tomkis:
**Pre-requisities**
- [x] I am using the newest version of the platform

**Describe the bug**
Canvas range selection doesn't work in Firefox when selecting text from right to left. Left-to-right selection works correctly.

**To Reproduce**
1. Open canvas with markdown content in Firefox
2. Select text from left to right → works
3. Select text from right to left → fails

**Expected behavior**
Text selection should work regardless of selection direction.

**Set-up:**
- Browser: Firefox (latest)

**Additional context**
The selection implementation in `useTextSelection.ts` uses Range's `startContainer/endContainer` which are always normalized (start before end). Firefox may handle backward selections differently than Chrome. The code doesn't account for selection direction using `anchorNode/focusNode` properties.

kapetr:
I wasn’t able to reproduce the issue, but several improvements to text selection were made in #1933 , which might address it. Feel free to reopen if the problem persists.

PR #1933, opened 2026-01-28
fix(ui): handle keyboard text selection in canvas

kapetr:
## Summary
The PR fixes issue where text selection in canvas didn’t work with keyboard. 

## Linked Issues
Refs #1874

## Documentation
- [x] No Docs Needed: just a fix

### DIFF
--- a/apps/agentstack-ui/src/modules/canvas/markdown/hooks/useTextSelection.ts
+++ b/apps/agentstack-ui/src/modules/canvas/markdown/hooks/useTextSelection.ts
@@ -60,7 +60,7 @@ export function useTextSelection({ containerRef, onSelectionChange }: Props) {
       debouncedProcessSelection.cancel();
       document.removeEventListener('selectionchange', debouncedProcessSelection);
     };
-  }, [containerRef, processSelection]);
+  }, [processSelection]);
 }
 
 const SELECTION_DEBOUNCE_MS = 100;


JanPokorny:
I wonder if you could simplify this through https://developer.mozilla.org/en-US/docs/Web/API/Document/selectionchange_event ? Gemini is right that by this approach you would still miss non-arrow keyboard navigation.

apps/agentstack-ui/src/modules/canvas/markdown/hooks/useTextSelection.ts
```diff
@@ -53,10 +54,28 @@ export function useTextSelection({ containerRef, onSelectionChange }: Props) {
...
+    document.addEventListener('keyup', handleKeyUp);
 
     return () => {
       container.removeEventListener('mouseup', handleMouseUp);
+      document.removeEventListener('keyup', handleKeyUp);
     };
   }, [containerRef, onSelectionChange]);
 }
+
+const ARROW_KEYS = ['ArrowLeft', 'ArrowRight', 'ArrowUp', 'ArrowDown'];
```
PetrBulanek: Not any issue, but you can use `keycode-js` that we have as a dependency (see `apps/agentstack-ui/src/utils/form-utils.ts`). :-)

PetrBulanek (approved):

Issue #1870, opened 2026-01-22
Support plain A2A agents without AgentStack SDK


tomkis:
**Is your feature request related to a problem? Please describe.**
Developers with existing A2A agents cannot use them in AgentStack without modifying their agent or using the AgentStack SDK.

**Describe the solution you'd like**
Two approaches should work:

1. **Managed agent (Docker image without label)** - `agentstack add <docker-image>` where image has no `beeai.dev.agent.json` label. AgentStack discovers agent card by starting container and probing `/.well-known/agent-card.json`.

2. **Unmanaged agent (network URL without agent-detail extension)** - `agentstack add <url>` pointing to plain A2A agent. AgentStack fetches agent card and auto-injects `agent-detail` extension.

**Describe alternatives you've considered**
Require all agents to use AgentStack SDK or manually add the label/extension - but this creates unnecessary friction for existing A2A agents.

**Additional context**
- Plain A2A agents only need to implement standard A2A spec
- `agent-detail` extension is AgentStack-specific and should be injected automatically
- SDK becomes optional enhancement, not requirement

PR #1980, opened 2026-01-30
feat: adding unmanaged a2a agents

tomkis:
## Summary
- Allow to add unmanaged A2A agent that does not provide AS specific metadata
- Fixed some code review comments from previous PR hence the unrelated changes

## Linked Issues
Part of #1870 

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-server/tests/e2e/agents/test_agent_starts.py
+++ b/apps/agentstack-server/tests/e2e/agents/test_agent_starts.py
@@ -213,10 +213,3 @@ async def test_unmanaged_a2a_agent(unmanaged_a2a_server):
         provider.agent_card.capabilities.extensions[0].uri
         == "https://a2a-extensions.agentstack.beeai.dev/ui/agent-detail/v1"
     )
-
-    for _ in range(20):
-        provider = await Provider.get(provider.id)
-        if provider.state == "online":
-            break
-        await asyncio.sleep(1)
-    assert provider.state == "online"


apps/agentstack-server/src/agentstack_server/service_layer/services/provider_discovery.py
jezekra1, line 124: We should log the exception if deletion was unsuccessful.

This has the potential to create and leave pods running forever if it fails (or if it can be expoited?) and increase the cloud bill. But this is an operation that only trusted users should be able to do.

Ideally, the cleanup should retry the deletion, for example using an extra k8s label for the temporary deployments (might be overkill now, if k8s api is responding, there are not many reasons for deletion to fail).

docs/development/deploy-agents/a2a-agents.mdx
```diff
@@ -0,0 +1,64 @@
...
+---
+
+Got an existing A2A agent that is not wrapped using Agent Stack SDK? No problem, you can still connect it to Agent Stack by providing a Docker image or its running URL.
+
+## Add Docker A2A Agent
+
+When you want your agent to be managed via Agent Stack, provide a Docker image:
+
+```bash
+agentstack add <docker-image-url>
```
jezekra1, line 13: Do we have a test for this functionality?

Can we add a similar test as [test_imported_agent](https://github.com/i-am-bee/agentstack/blob/main/apps/agentstack-server/tests/e2e/agents/test_agent_starts.py#L27)?

tomkis: Proper tests will be solved in https://github.com/i-am-bee/agentstack/issues/1999

apps/agentstack-server/src/agentstack_server/service_layer/services/provider_discovery.py
tomkis: I've added just logging for now.

PR #1990, opened 2026-02-02
CLI - Improve `agentstack update` UX

jjeliga:
## Summary
When updating agents added using github url via `agentstack agent update ...`, the code now correctly recognizes the github origin and shows a list of tags to choose if available. Similar logic existed in the "add" command, where the functionality has been improved to allow tag selection right away when inputting a bare github url

## Linked Issues
Closes https://github.com/i-am-bee/agentstack/issues/1788

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-cli/src/agentstack_cli/utils.py
+++ b/apps/agentstack-cli/src/agentstack_cli/utils.py
@@ -311,7 +311,7 @@ def is_github_url(url: str) -> bool:
     return bool(re.match(github_url_verbose_pattern, url, re.VERBOSE))
 
 
-def get_github_token() -> str | None:
+def get_local_github_token() -> str | None:
     """Get GitHub token from standard local sources for authenticated API requests.
 
     Checks multiple sources in order of preference:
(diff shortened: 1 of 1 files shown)


apps/agentstack-cli/src/agentstack_cli/commands/agent.py
```diff
@@ -245,33 +245,44 @@ async def add_agent(
...
-        async with httpx.AsyncClient() as client:
-            response = await client.get(
-                f"https://api.github.com/repos/{owner}/{repo}/tags",
-                headers={"Accept": "application/vnd.github.v3+json"},
-            )
-            tags = [tag["name"] for tag in response.json()] if response.status_code == 200 else []
+        if version is None and path is None:
+            async with httpx.AsyncClient() as client:
+                response = await client.get(
+                    f"https://api.github.com/repos/{owner}/{repo}/tags",
```
jezekra1: The github api has strict rate limits for unauthenticated users (only 60 requests per hour!):
https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28

It's based on IP address so it will break for users under NAT who share the same IP (very common scenario).

Consider using unauthenticated approach using the internal `github.com` requests not `api.github.com`, similar to:

https://github.com/i-am-bee/agentstack/blob/main/apps/agentstack-server/src/agentstack_server/utils/github.py#L180

apps/agentstack-cli/src/agentstack_cli/commands/agent.py
```diff
@@ -362,7 +370,7 @@ async def update_agent(
             location = (
                 await inquirer.text(
                     message="Enter new agent location (public docker image or github url):",
-                    default=provider.source,
+                    default=provider.origin.lstrip("git+"),
```
JanPokorny, line 373: Beware, you likely wanted `removeprefix`. `lstrip` removes any characters from the set. In this case it might not make a difference but just pointing it out 😁

JanPokorny (approved):

PR #2032, opened 2026-02-06
Feat(sdk, cli): 1998 extend form extension for settings

### LINKED ISSUE #1998: [SDK] Extend Form extension to support settings rendering

Part of #1863. Settings and Form extensions share overlapping types (`SingleSelectField`, `OptionItem`, etc.) causing import clashes (#1653) and API redundancy.

1. Extend `FormExtensionSpec` to accept `settings` demand key (currently only `initial_form` supported)
2. Validate settings-rendered forms - only `SingleSelect`, `MultiSelect`, `Checkbox` allowed
3. Deprecate `SettingsExtension*` types with warning pointing to Form-based approach

Keep both extensions with aliased imports - doesn't solve duplication.

**Additional context**

Relates to: #1863, #1653

jjeliga:
## Summary
Moving the settings extension under form.
Extending `FormServiceExtensionSpec` to demand either form or settings or both.
Extending `FormServiceExtensionServer` to allow settings parsing.
Allowing only `CheckboxGroupField` and `SingleSelectField` to be used for settings to be aligned with the ui.
Deprecating the old settings, but keeping them for now to allow compatibility with older agents.
Modifying cli `agent run` cmd to be able to work with both the new and deprecated settings.
Including examples using both old and new method.
Updated docs.

## Linked Issues
Closes #1998 

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-server/tests/e2e/examples/agent-integration/forms/test_settings_form_rendering.py
+++ b/apps/agentstack-server/tests/e2e/examples/agent-integration/forms/test_settings_form_rendering.py
@@ -4,7 +4,7 @@
 import pytest
 from a2a.client.helpers import create_text_message_object
 from a2a.types import TaskState
-from agentstack_sdk.a2a.extensions import SettingsFormResponse, FormServiceExtensionMetadata, FormServiceExtensionSpec
+from agentstack_sdk.a2a.extensions import FormServiceExtensionMetadata, FormServiceExtensionSpec, SettingsFormResponse
 from agentstack_sdk.a2a.extensions.common.form import CheckboxGroupFieldValue, SingleSelectFieldValue
 
 from tests.e2e.examples.conftest import run_example
@@ -26,7 +26,8 @@ async def test_settings_form_rendering_example(subtests, get_final_task_from_str
                     "settings_form": SettingsFormResponse(
                         values={
                             "checkbox_settings": CheckboxGroupFieldValue(value={"thinking": True, "memory": False}),
-                            "response_style": SingleSelectFieldValue(value="humorous")}
+                            "response_style": SingleSelectFieldValue(value="humorous"),
+                        }
                     )
                 }
             ).model_dump(mode="json")


apps/agentstack-cli/src/agentstack_cli/commands/agent.py
```diff
@@ -50,29 +50,38 @@
...
     SettingsFieldValue,
     SettingsRender,
 )
 from agentstack_sdk.a2a.extensions.ui.settings import (
     CheckboxFieldValue as SettingsCheckboxFieldValue,
 )
+from agentstack_sdk.a2a.extensions.ui.settings import CheckboxGroupField as SettingsCheckboxGroupField
+from agentstack_sdk.a2a.extensions.ui.settings import (
+    CheckboxGroupFieldValue as SettingsCheckboxGroupFieldValue,
+)
```
jjeliga, line 86: I know these imports look weird, but that is the work of ruff. No-harm as it will be deleted soon

apps/agentstack-cli/src/agentstack_cli/commands/agent.py
```diff
@@ -1122,19 +1200,27 @@ async def run_agent(
...
             console.print(f"{user_greeting}\n")
-            settings_input = await _ask_settings_questions(settings_render) if settings_render else None
+            # Ask settings based on format (legacy or new)
+            settings_input: AgentRunSettings | SettingsFormResponse | None = None
+            if settings_render:
+                # TODO: remove once legacy settings extension is fully deprecated
+                if is_legacy_settings:
+                    settings_input = await _ask_settings_questions(settings_render)  # type: ignore
+                else:
+                    settings_input = await _ask_settings_form_questions(settings_render)  # type: ignore
```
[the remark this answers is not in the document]
jjeliga: this will be reduced to a on-liner during full settings removal

docs/development/agent-integration/agent-settings.mdx
```diff
@@ -3,75 +3,175 @@ title: "Configure Agent Settings"
...
-    """Agent that demonstrates settings configuration"""
+    """Demonstrate settings form extension"""
+
+    if not form:
+        yield "Form extension hasn't been activated, no settings are available"
+        return
 
-    if not settings:
-        yield "Settings extension hasn't been activated, no settings are available"
+    parsed_settings = form.parse_settings_form(model=SettingsModel)
```
jezekra1, line 127: In other extension we parse and store the models automatically, is there a reason why user might want to control or change this behavior?

I can imagine just a property that's already parsed
```python
if not form.settings:
    yield "No settings provided"
    return
```

similar to `llm_ext.data` for example.

Also I don't understand why we would want to gracefully handle parse errors -> that's an error of the client to provide correct data and the agent should fail loudly instead of quietly sending message "Parsed settings in unexpected format" and succeeding. Am I missing something?

cc @tomkis

jjeliga: ad "Parsed settings in unexpected format": This is just an example where it won't happen unless someone wrongly modifies the settings or model, it just shows user that it truly is set-up correctly or incorrectly if they play with it. I don't think anyone would do such a thing in production

jjeliga: ad automatic parsing: It is actually parsed, but this additional optional parsing retrieves info from depths of the form.data and can provided it in convenient format if user wishes

PR #2051, opened 2026-02-09
feat(ui): rework trajectory visualization

### LINKED ISSUE #1674: UI - Redesign trajectory visualization for better readability during agent execution
## Problem

The current trajectory display has major usability issues that make it hard to follow what the agent is doing:

1. Tiny scrolling window - Trajectories are squeezed into a small box that auto-scrolls constantly
2. Can't read content - The rapid auto-scrolling makes it impossible to read trajectory items as they appear
3. Poor visibility - Users can only see a fraction of the agent's reasoning at once

The trajectory section scrolls within a restricted ~240px height container, creating a "porthole" effect where content rapidly scrolls by.

## Desired Experience

Users should be able to see and read the full trajectory as the agent executes:

- Show all trajectory items without height restrictions or scrolling
- As new trajectory items arrive, the section should grow naturally downward
- Content should stay visible long enough to read without jumping around
- When users toggle trajectories open/closed, it should animate smoothly regardless of content length

## Design Requirements

- Remove the scrolling container constraint - let trajectories extend to their natural height
- Ensure smooth, natural expansion as new items arrive during execution
- Collapse trajectory once agent output begins to stream

## Figma

In progress
https://www.figma.com/design/oaJEgk4ZcH3mI1juPGITsx/Agent-Testbed-dev-ready--external-?node-id=1533-8728&p=f&t=c4QShobVWTStNXLI-0

kapetr:
## Summary
Implements the trajectory UI redesign with improved animations, dynamic header behavior, and better content management during agent streaming.

## Linked Issues
Closes #1674

## Documentation
- [x] No Docs Needed: UI only

### DIFF
--- a/apps/agentstack-ui/src/modules/trajectories/components/AnimatedCodeContent.tsx
+++ b/apps/agentstack-ui/src/modules/trajectories/components/AnimatedCodeContent.tsx
@@ -5,21 +5,17 @@
 
 import { CodeSnippet } from '#components/CodeSnippet/CodeSnippet.tsx';
 
+import type { UseAnimatedTextOptions } from '../hooks/useAnimatedText';
 import { useAnimatedText } from '../hooks/useAnimatedText';
 
-interface Props {
+interface Props extends UseAnimatedTextOptions {
   children: string;
-  totalDurationMs: number;
-  maxAnimatedChars?: number;
-  shouldAnimate?: boolean;
 }
 
-export function AnimatedCodeContent({ children, shouldAnimate, totalDurationMs, maxAnimatedChars }: Props) {
+export function AnimatedCodeContent({ children, ...textOptions }: Props) {
   const displayedText = useAnimatedText({
+    ...textOptions,
     text: children,
-    shouldAnimate,
-    totalDurationMs,
-    maxAnimatedChars,
   });
 
   return (
--- a/apps/agentstack-ui/src/modules/trajectories/components/MessageTrajectories.module.scss
+++ b/apps/agentstack-ui/src/modules/trajectories/components/MessageTrajectories.module.scss
@@ -3,25 +3,6 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-.root {
-  &:global(.cds--btn) {
-    @include reset-focus();
-    padding-inline: 0;
-    color: $expand-button-color;
-    &:hover,
-    &:active {
-      background-color: transparent;
-      color: $button-tertiary;
-    }
-  }
-
-  &.isOpen {
-    :global(.cds--btn__icon) {
-      transform: scaleY(-1);
-    }
-  }
-}
-
 .spinner {
   min-block-size: rem(26px);
   display: flex;
--- a/apps/agentstack-ui/src/modules/trajectories/components/TrajectoryList.tsx
+++ b/apps/agentstack-ui/src/modules/trajectories/components/TrajectoryList.tsx
@@ -5,7 +5,7 @@
 
 import clsx from 'clsx';
 import { AnimatePresence, motion } from 'framer-motion';
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { usePrevious } from '#hooks/usePrevious.ts';
 import type { UITrajectoryPart } from '#modules/messages/types.ts';
@@ -21,8 +21,6 @@ interface Props {
 }
 
 export function TrajectoryList({ trajectories, isOpen, isPending }: Props) {
-  const listRef = useRef<HTMLUListElement>(null);
-
   const [canClampContent, setCanClampContent] = useState(!isPending);
 
   const previouslyOpen = usePrevious(isOpen);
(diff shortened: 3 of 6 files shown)


apps/agentstack-ui/src/modules/trajectories/hooks/useAnimatedText.ts
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    if (!shouldAnimate) {
+      setDisplayedText(text);
+      return;
+    }
+
+    const charsToAnimate = Math.min(text.length, maxAnimatedChars);
+    const remainingText = text.slice(charsToAnimate);
+
+    const delayPerChar = Math.min(totalDurationMs / charsToAnimate, CHARS_DELAY_MAX_MS);
```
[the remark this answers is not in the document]
kapetr: The check is already included in `shouldAnimate`. But I agree that could be prone to errors on code change, we can double check it.

apps/agentstack-ui/src/modules/trajectories/utils.ts
```diff
@@ -14,3 +17,20 @@ export function hasViewableTrajectoryParts(trajectory: UITrajectoryPart) {
...
+      const paragraph = result.children[0];
+      const hasOnlyText = paragraph.children.every((child: PhrasingContent) => child.type === 'text');
+      return !hasOnlyText;
+    }
+
+    return true;
+  } catch {
+    return true;
+  }
+}
```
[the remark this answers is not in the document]
kapetr: This check is purely for eliminating attempts on markdown renders, when we know it's not necessary. And it should be defensive, so rendering non-Markdown as Markdown in some edge case is totally fine.
1. Not even true
2. For this case, it's safer to assume it could still be Markdown, and our MarkdownContent component might be able to handle it.

apps/agentstack-ui/src/modules/trajectories/components/AnimatedCodeContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { CodeSnippet } from '#components/CodeSnippet/CodeSnippet.tsx';
+
+import { useAnimatedText } from '../hooks/useAnimatedText';
+
+interface Props {
+  children: string;
+  totalDurationMs: number;
+  maxAnimatedChars?: number;
+  shouldAnimate?: boolean;
+}
```
PetrBulanek: This is kind of the same as in `AnimatedTextContent` component and also as `UseAnimatedTextOptions` interface. I think we could reuse it.

apps/agentstack-ui/src/modules/trajectories/components/MessageTrajectories.module.scss
```diff
@@ -0,0 +1,29 @@
...
+      color: $button-tertiary;
+    }
+  }
+
+  &.isOpen {
+    :global(.cds--btn__icon) {
+      transform: scaleY(-1);
+    }
+  }
+}
```
PetrBulanek: It seems those styles are not used.

apps/agentstack-ui/src/modules/trajectories/utils.ts
```diff
@@ -14,3 +17,20 @@ export function hasViewableTrajectoryParts(trajectory: UITrajectoryPart) {
...
+    // If there's a single paragraph with only text nodes, it's plain text (not markdown)
+    if (result.children.length === 1 && result.children[0].type === 'paragraph') {
+      const paragraph = result.children[0];
+      const hasOnlyText = paragraph.children.every((child: PhrasingContent) => child.type === 'text');
+      return !hasOnlyText;
+    }
+
+    return true;
+  } catch {
+    return true;
```
PetrBulanek, line 34: Shouldn't we rather return false, if the parsing fails?

apps/agentstack-ui/src/modules/trajectories/components/TrajectoryList.tsx
```diff
@@ -17,30 +17,22 @@ import classes from './TrajectoryList.module.scss';
...
   trajectories: UITrajectoryPart[];
   isOpen?: boolean;
-  autoScroll?: boolean;
+  isPending?: boolean;
 }
 
-export function TrajectoryList({ trajectories, isOpen, autoScroll }: Props) {
-  const { ref: autoScrollRef } = useAutoScroll<HTMLLIElement>([trajectories.length], { duration: AUTOSCROLL_DURATION });
+export function TrajectoryList({ trajectories, isOpen, isPending }: Props) {
   const listRef = useRef<HTMLUListElement>(null);
```
PetrBulanek: `listRef` is not used anymore and can me removed.

PetrBulanek (changes requested):

apps/agentstack-ui/src/modules/trajectories/utils.ts
kapetr: As per my reply to gemini - it's safer to assume it could still be Markdown, and our MarkdownContent component might be able to handle it. 

That is my take, but let me know if you still think it's wrong behaviour.

PetrBulanek: Ok, got it. 👍

PR #2066, opened 2026-02-10
chore(ui): migrate nextjs to v16

kapetr:
## Summary
Migrate Next.js to the new major version 16. Remove unnecessary `useCallback` and `useMemo` usages, because the new React compiler should handle basic optimizations. 

## Documentation
- [x] No Docs Needed: bump dependencies

### DIFF
--- a/apps/agentstack-ui/next.config.ts
+++ b/apps/agentstack-ui/next.config.ts
@@ -89,18 +89,6 @@ const nextConfig: NextConfig = withBundleAnalyzer({
   devIndicators: {
     position: 'bottom-right',
   },
-  turbopack: {
-    rules: {
-      '*.svg': {
-        loaders: ['@svgr/webpack'],
-        as: '*.js',
-      },
-    },
-    resolveAlias: {
-      // Carbon styles use tilde imports for IBM Plex fonts which is a webpack convention
-      '~@ibm/plex': '@ibm/plex',
-    },
-  },
 });
 
 export default nextConfig;
--- a/apps/agentstack-ui/src/components/SyntaxHighlighter/SyntaxHighlighter.tsx
+++ b/apps/agentstack-ui/src/components/SyntaxHighlighter/SyntaxHighlighter.tsx
@@ -8,6 +8,7 @@
 import { CodeSnippetSkeleton } from '@carbon/react';
 import clsx from 'clsx';
 import dynamic from 'next/dynamic';
+import type { SyntaxHighlighterProps } from 'react-syntax-highlighter';
 
 import { registerLanguagesAsync } from './languages';
 import classes from './SyntaxHighlighter.module.scss';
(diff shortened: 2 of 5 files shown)


apps/agentstack-ui/src/components/SyntaxHighlighter/SyntaxHighlighter.tsx
PetrBulanek, line 1: I think this is not needed. I got the same error, but it was just a deps mismatch. Removing and recreating lockfile solved it.

apps/agentstack-ui/next.config.ts
```diff
@@ -88,6 +89,18 @@ const nextConfig: NextConfig = withBundleAnalyzer({
   devIndicators: {
     position: 'bottom-right',
   },
+  turbopack: {
+    rules: {
+      '*.svg': {
+        loaders: ['@svgr/webpack'],
+        as: '*.js',
+      },
```
PetrBulanek: With the default webpack config we have set an option that disables removing viewBox. Don't we have to add it there as well?

apps/agentstack-ui/tasks.toml
```diff
@@ -85,7 +85,8 @@ run = "pnpm next start"
 depends = ["common:setup:pnpm", "agentstack-sdk-ts:build"]
 env.NODE_OPTIONS = "--no-experimental-global-navigator"
 dir = "{{config_root}}/apps/agentstack-ui"
-run = "pnpm next dev"
+# Using --webpack due to Turbopack not supporting tilde imports in @carbon/styles fonts
+run = "pnpm next dev --webpack"
```
PetrBulanek, line 89: Do we actually need to setup the Turbopack in config if use Webpack? :-)

apps/beeai-web/next.config.ts
```diff
@@ -89,6 +90,18 @@ const nextConfig: NextConfig = {
     cssChunking: false,
     optimizePackageImports: ['@carbon/react', '@carbon/icons-react'],
   },
+  turbopack: {
+    rules: {
+      '*.svg': {
+        loaders: ['@svgr/webpack'],
+        as: '*.js',
+      },
```
PetrBulanek: Same here.

apps/agentstack-ui/next.config.ts
kapetr: Unfortunately we won't be able to use Turbopack in the end - it does not support some of the required features. Removing.

apps/agentstack-ui/tasks.toml
kapetr: This was not supposed to be here :) But eventually we'll need after trying to get the Turbopack to work - it's not ready.

apps/beeai-web/next.config.ts
kapetr: Removed

apps/agentstack-ui/src/components/SyntaxHighlighter/SyntaxHighlighter.tsx
kapetr: This was needed, but I updated the solution without typecasting.

Issue #1551, opened 2025-11-13
UI - Support agent input prompts during execution


jenna-winkler:
## Problem

Agents can request input from users while they're running (like asking "What email should I use?" in the middle of a task), but the web UI doesn't support this yet. It works in the CLI but not in the browser.

## What it looks like in CLI

<img width="1020" height="644" alt="Image" src="https://github.com/user-attachments/assets/f3b37ac6-a276-4aa8-b354-3e1b4509b6ba" />

## Example use case

An agent is composing an email and needs to ask:
1. "What email address should I use?"
2. "What subject line do you want?"

The agent pauses, waits for your answer, then continues with your input.

## Code example

https://github.com/i-am-bee/agentstack/blob/jenna-winkler-patch-2/apps/agentstack-sdk-py/examples/input_required.py 

## Designs needed

- Show the agent's question in the chat
- Let users type their answer and submit it
- Agent continues running with the user's response
- Support multiple questions in one conversation

PR #2079, opened 2026-02-11
feat(ui): add text input required support

kapetr:
## Summary
Implements support for a generic input-required event in the UI.
When the event is received, a form with a text input is displayed, allowing the user to respond to the agent.

## Linked Issues
Refs #1551 

## Documentation
- [ ] No Docs Needed:

### DIFF
--- a/apps/agentstack-sdk-ts/src/client/core/handle-task-status-update.ts
+++ b/apps/agentstack-sdk-ts/src/client/core/handle-task-status-update.ts
@@ -11,6 +11,7 @@ import type { TaskStatusUpdateEvent } from '../a2a/protocol/types';
 import { extractUiExtensionData } from './extensions/extract';
 import type { TaskStatusUpdateResult } from './extensions/types';
 import { TaskStatusUpdateType } from './extensions/types';
+import { extractTextFromMessage } from './utils/extract-text-from-message';
 
 const secretsRequestExtensionExtractor = extractUiExtensionData(secretsRequestExtension);
 const oauthRequestExtensionExtractor = extractUiExtensionData(oauthRequestExtension);
@@ -52,11 +53,11 @@ export const handleTaskStatusUpdate = (event: TaskStatusUpdateEvent): TaskStatus
         request: approvalRequired,
       });
     } else {
-      const textPart = event.status.message?.parts?.find(({ kind }) => kind === 'text');
-      if (textPart && textPart.kind === 'text') {
+      const text = extractTextFromMessage(event.status.message);
+      if (text) {
         results.push({
           type: TaskStatusUpdateType.TextInputRequired,
-          text: textPart.text,
+          text,
         });
       }
     }
--- a/apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
+++ b/apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
@@ -3,10 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import type { TextField } from 'agentstack-sdk';
 import { useMemo } from 'react';
 
 import { useMessages } from '#modules/messages/contexts/Messages/index.ts';
-import type { UIAgentMessage } from '#modules/messages/types.ts';
+import type { UIAgentMessage, UITextInputPart } from '#modules/messages/types.ts';
 import { UIMessagePartKind } from '#modules/messages/types.ts';
 import { useAgentRun } from '#modules/runs/contexts/agent-run/index.ts';
 import { blurActiveElement } from '#utils/dom-utils.ts';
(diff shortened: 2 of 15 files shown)


apps/agentstack-sdk-ts/src/client/core/handle-task-status-update.ts
```diff
@@ -46,13 +46,19 @@ export const handleTaskStatusUpdate = (event: TaskStatusUpdateEvent): TaskStatus
...
-    }
-
-    if (approvalRequired) {
+    } else if (approvalRequired) {
       results.push({
         type: TaskStatusUpdateType.ApprovalRequired,
         request: approvalRequired,
       });
+    } else {
+      const textPart = event.status.message?.parts?.find(({ kind }) => kind === 'text');
```
PetrBulanek: What about using `extractTextFromMessage` helper to extract the text directly to simplify this?

PetrBulanek: Aha.. we are inside SDK, the helper is part of UI. Sorry. :-)

But we might actually add similar utils into the SDK as well, I guess.

apps/agentstack-ui/src/modules/runs/contexts/agent-run/agent-run-context.ts
```diff
@@ -34,6 +34,7 @@ interface AgentRunContextValue {
   startAuth: (url: string, taskId: TaskId) => void;
   submitSecrets: (taskId: TaskId, secrets: Record<string, string>) => Promise<void>;
   submitApproval: (taskId: TaskId, decision: ApprovalDecision) => Promise<void>;
```
PetrBulanek: Just nitpicking, but we could probably change the order of the params in `submitSecrets` and `submitApproval` to make it unified (first data, second taskId).

apps/agentstack-ui/src/modules/runs/contexts/agent-run/AgentRunProvider.tsx
PetrBulanek, line 1: I think we might switch here to use the `match(result)` pattern instead this large if/else to make it bit more readable.

Such as this.
```
match(result)
  .with({ type: TaskStatusUpdateType.OAuthRequired }, ({ url, taskId }) => {
    updateCurrentAgentMessage((message) => {
      message.status = UIMessageStatus.InputRequired;
      message.parts.push({ kind: UIMessagePartKind.OAuth, url, taskId });
    });
  })
```

apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { FormRenderer } from './FormRenderer';
+import classes from './MessageTextInput.module.scss';
+
+interface Props {
+  message: UIAgentMessage;
+}
+
+export function MessageTextInput({ message }: Props) {
+  const textInputPart = message.parts.find(
+    (part): part is Extract<typeof part, { kind: UIMessagePartKind.TextInput }> =>
```
PetrBulanek: I think just `part is UITextInputPart` should work as well.

apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+    return {
+      fields: [
+        {
+          id: 'text',
+          type: 'text' as const,
+          label: textInputPart.text,
+          required: true,
+          auto_resize: true,
+        },
```
PetrBulanek: I kind of gotten into using `satisfies`. You can also get rid of the `as const` casting.

```
{
  id: 'text',
  type: 'text',
  label: textInputPart.text,
  required: true,
  auto_resize: true,
} satisfies TextField
```

apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { submitTextInput } = useAgentRun();
+  const { isLastMessage } = useMessages();
+
+  const formDefinition = useMemo(() => {
+    if (!textInputPart) return null;
+
+    return {
+      fields: [
+        {
+          id: 'text',
```
PetrBulanek: We should probably make it a random id to avoid potential conflicts and I'd make it a variable so we can also reuse it when accessing the value. It is not explicitly apparent now.

apps/agentstack-sdk-ts/src/client/core/handle-task-status-update.ts
kapetr: Agree!

apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
kapetr: Much better!

apps/agentstack-ui/src/modules/form/components/MessageTextInput.tsx
kapetr: I'd also make it a variable and I agree with randomizing the id, but we should probably handle it at the field component level (TextField, ...), because the same issue applies to generated forms. What do you think?

PR #2112, opened 2026-02-17
fix(cli): terminating the auth session on logout and checking auth info changes on login

jjeliga:
## Summary
Using CLI
Fix1: when re-logging in to a known server, we check whether the auth_servers have changed and act accordingly
Fix2: when logging out of a server, the auth server session is now properly terminated (tested for local keycloak)

## Linked Issues
closes https://github.com/i-am-bee/agentstack/issues/1997

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-cli/src/agentstack_cli/commands/server.py
+++ b/apps/agentstack-cli/src/agentstack_cli/commands/server.py
@@ -124,12 +124,15 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] = 
             else:
                 # shortcut
                 auth_servers = []
-        except RuntimeError:
+        except Exception as e:
+            console.warning(
+                f"Failed to fetch server auth metadata, continuing without validation of current auth servers: {e!s}"
+            )
             # Continue with the current auth servers info, which might still be valid even if metadata fetching failed
             pass
 
         if not auth_servers:
-            # Re-check server capabilities in case auth was enabled after the initial login.
+            # Re-check in case auth was enabled (leads to new login) or disabled (just passes) after the latest login and auth info update.
             if oauth_metadata is None or not oauth_metadata.get("authorization_servers", []):
                 # Keep backward-compatible behavior when metadata cannot be fetched.
                 # This might lead to errors when the server actually requires auth but does not provide standard info; however, that is not expected


apps/agentstack-cli/src/agentstack_cli/commands/server.py
```diff
@@ -105,18 +105,43 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] =
...
         auth_servers = list(server_data.authorization_servers.keys())
+
+        oauth_metadata = None
+        try:
+            oauth_metadata = await config.auth_manager.fetch_oauth_protected_resource_metadata(server)
+            oauth_metadata_auth_servers = oauth_metadata.get("authorization_servers", [])
+            if oauth_metadata_auth_servers:
+                # something might have changed on the server side since we last logged in
+                # re-check that the auth servers we know about are still valid
+                auth_servers = [as_ for as_ in auth_servers if as_ in oauth_metadata_auth_servers]
```
JanPokorny, line 123: Nitpikcing time! IMO it would look better like `auth_servers = list(set(auth_servers) & set(oauth_metadata_auth_servers))` 😁 You could get rid of the wrapping `if` too. If you keep `auth_servers` as a `set`, then it could be just `auth_servers &= set(oauth_metadata_auth_servers)`.

apps/agentstack-cli/src/agentstack_cli/commands/server.py
```diff
@@ -105,18 +105,43 @@ async def server_login(server: typing.Annotated[str | None, typer.Argument()] =
...
+            else:
+                # shortcut
+                auth_servers = []
+        except RuntimeError:
```
JanPokorny: We might want to be a bit more specific here?

apps/agentstack-cli/src/agentstack_cli/commands/server.py
jjeliga: I wanted to use sets at first :D But I opted for this because `auth_server = auth_servers[0]` suggests that the order might matter

apps/agentstack-cli/src/agentstack_cli/commands/server.py
jjeliga: Added a console warning

PR #2141, opened 2026-02-24
feat(ui): adaptive animation for trajectory items

### LINKED ISSUE #2123: UI - Adaptive trajectory animation to prevent backlog and preserve live flow
Related to #1674 

## Context

During discussion in Slack, we identified a core issue with the current trajectory animation approach:

- When multiple trajectory steps arrive quickly, sequential character-by-character animation (1.5s per step) can create noticeable delays.
- In extreme cases (e.g. 10–50 steps emitted quickly), the UI may lag far behind the agent’s actual progress.
- Batching steps avoids delay but breaks the “live reasoning” feel and can appear visually abrupt.

We want to:
- Avoid long animation backlogs  
- Preserve sequential readability  
- Maintain a natural “live reasoning” experience  
- Avoid flicker or illegible instant rendering  

#### Option A – Adaptive Sequential Animation
Animate steps one by one, but dynamically speed up when the queue grows.

#### Option B – Animate Latest Only
Always animate only the newest step.  
If a new step arrives:
- Instantly complete the previous one  
- Start animating the new last step  

## Suggested Hybrid Approach

To balance UX and performance:

- Default animation duration: **1.5s per step**
- If `queue.length > 2`, progressively speed up: duration = 1.5 / queue.length
- If `queue.length > 6`, render the oldest steps instantly (performance optimization) and continue animating only the most recent ones.

## Expected Outcome

- Eliminates visible delay between agent execution and UI.
- Keeps sequential flow.
- Introduces a more natural, variable “streaming-like” animation speed when the agent emits steps quickly.
- Avoids large animation backlogs.
- Preserves readability without abrupt batch rendering.

kapetr:
## Summary
Improves trajectory animation by dynamically adjusting animation speed based on the number of pending trajectories, while preserving the sequential flow.

Instead of using a queue-length threshold, the solution introduces a time-based check for trajectories waiting in the queue. If a trajectory step has been waiting for more than 3 seconds since its creation, it skips the character-by-character animation and renders instantly.

Additionally, due to browser timing limitations (animations are effectively capped at ~60 FPS), I added an optimization:
If the calculated animation speed would require updates faster than the browser can realistically render, we animate multiple characters per frame instead of strictly one character at a time. This keeps the perceived speed consistent without overwhelming the rendering loop.

This brings two key benefits:
- **Stronger guardrail** — We never delay displaying a trajectory longer than 3 seconds, regardless of queue size. This prevents the UI from drifting too far behind the agent’s actual progress.
- **Performance optimization** — It avoids overwhelming the browser during heavy bursts. React and the rendering loop struggle with extremely small update intervals, so batching characters per frame ensures smooth animation even under load.

## Linked Issues
Closes #2123 

## Documentation
- [x] No Docs Needed:

### DIFF
--- a/apps/agentstack-ui/src/modules/trajectories/components/AnimatedTextContent.tsx
+++ b/apps/agentstack-ui/src/modules/trajectories/components/AnimatedTextContent.tsx
@@ -30,12 +30,12 @@ export function AnimatedTextContent({ children, className, linesClamp, ...textOp
 
   if (linesClamp && displayedText.length > 0) {
     return (
-      <LineClampText lines={linesClamp} useBlockElement={contentIsMarkdown} className={className}>
-        {contentIsMarkdown ? (
-          <MarkdownContent>{displayedText}</MarkdownContent>
-        ) : (
-          <span className={classes.content}>{displayedText}</span>
-        )}
+      <LineClampText
+        lines={linesClamp}
+        useBlockElement={contentIsMarkdown}
+        className={clsx(classes.content, className)}
+      >
+        {contentIsMarkdown ? <MarkdownContent>{displayedText}</MarkdownContent> : displayedText}
       </LineClampText>
     );
   }


apps/agentstack-ui/src/modules/trajectories/hooks/useAnimatedText.ts
```diff
@@ -18,35 +25,36 @@ interface UseAnimatedTextProps extends UseAnimatedTextOptions {
...
     }
 
+    const charsToAnimate = Math.min(text.length, maxAnimatedChars);
     const remainingText = text.slice(charsToAnimate);
     const delayPerChar = Math.min(totalDurationMs / charsToAnimate, CHARS_DELAY_MAX_MS);
 
     let intervalId: NodeJS.Timeout;
     const startAnimation = () => {
-      let currentIndex = 0;
       intervalId = setInterval(() => {
```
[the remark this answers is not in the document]
kapetr: This is by design, we don't want the animation to start over on parameteres change, but rather continue.

apps/agentstack-ui/src/modules/trajectories/components/TrajectoryList.tsx
```diff
@@ -22,16 +28,65 @@ interface Props {
...
+        ...trajectory,
+        status: AnimationStatus.Animating,
+        duration: durationMs,
+      });
+
+      animatingTrajectoryIdx = index;
+    }
+
+    return animatedTrajectories;
+  }, [animatedDoneTrajectoryIds, isPending, trajectories]);
```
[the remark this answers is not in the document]
kapetr: Gemini is right, but I don’t think it’s an issue in this case.

apps/agentstack-ui/src/modules/trajectories/components/AnimatedTextContent.tsx
```diff
@@ -29,11 +31,18 @@ export function AnimatedTextContent({ children, className, linesClamp, ...textOp
   if (linesClamp && displayedText.length > 0) {
     return (
       <LineClampText lines={linesClamp} useBlockElement={contentIsMarkdown} className={className}>
-        {contentIsMarkdown ? <MarkdownContent>{displayedText}</MarkdownContent> : displayedText}
+        {contentIsMarkdown ? (
+          <MarkdownContent>{displayedText}</MarkdownContent>
+        ) : (
+          <div className={classes.content}>{displayedText}</div>
```
PetrBulanek: If `contentIsMarkdown` is false, the `LineClampText` component renders span, so we should also use span instead of div here.

apps/agentstack-ui/src/modules/trajectories/components/TrajectoryItem.tsx
```diff
@@ -34,22 +36,24 @@ export function TrajectoryItem({ trajectory, isPending, canClampContent }: Props
     return null;
```
PetrBulanek: Trajectory without content is not rendered, so it never calls `onAnimationEnd`. This might be an issue, because that trajectory is not marked as completed and it could block the queue.

PetrBulanek (changes requested):

apps/agentstack-ui/src/modules/trajectories/components/TrajectoryItem.tsx
PetrBulanek: Simple fix could be not to mark the trajectories that are not rendered (don't have content) as Animating, but mark them as Completed, or skip them altogether.

PetrBulanek: Or we might actually want to render at least the trajectories' title if present.

apps/agentstack-ui/src/modules/trajectories/components/AnimatedTextContent.tsx
kapetr: I might be missing something, but I see `div` rendered in both cases.

apps/agentstack-ui/src/modules/trajectories/components/TrajectoryItem.tsx
kapetr: Good catch!

apps/agentstack-ui/src/modules/trajectories/components/AnimatedTextContent.tsx
PetrBulanek: The `LineClampText` component renders span or div based on the `useBlockElement` boolean. (`const Component = useBlockElement ? 'div' : 'span';`) That means if the `contentIsMarkdown` is false, the `LineClampText` internally renders `span`, but the children is `div`. Hope it makes sense now. :-)

kapetr: Ah, you're right. Fixing now.

PR #2296, opened 2026-03-05
Add commit-msg git hook to rewrite AI co-author trailers

tomkis:
## Summary
- Add `scripts/commit-msg.sh` that replaces `Co-authored-by` trailers from Claude/Cursor with a single `Assisted-By` trailer
- Install the `commit-msg` hook alongside the existing `pre-commit` hook in `common:setup:git-hooks` mise task
- Switch from `printf` to heredocs to avoid shebang escaping issues in zsh

## Test plan
- [x] Run `mise run setup` and verify `.git/hooks/commit-msg` is created
- [x] Make a commit with a Claude co-author trailer and verify it gets rewritten

🤖 Generated with [Claude Code](https://claude.com/claude-code)


JanPokorny (changes requested):
Can you do this directly with Mise tasks? Consistent with `git-hooks:pre-commit` task we already have for `check`. You can create `git-hooks:commit-msg` and inline this script as an implementation.

